PR #7, opened 2018-07-20
Use pruning with dep

justinsb:
Reduces the size of our vendor directory and lets us see what is
really in use.

```release-note
NONE
```


justinsb:
Done in #9

PR #6, opened 2018-07-20
Add scripts to that we can turn on presubmit checks

roberthbailey:
**What this PR does / why we need it**: Add scripts to that we can turn on presubmit checks

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**: This will be simplest to review one commit at a time. Once https://github.com/kubernetes/test-infra/pull/8751 merges prow will auto-squash these commits. 

**Release note**:

```release-note
NONE
```

PR #8, opened 2018-07-20
Fix paths in Makefile

justinsb:
Note that the image and push targets are still broken

```release-note
NONE
```


justinsb:
Done in #6

PR #25, opened 2018-07-25
Introduce bazel / gazelle

justinsb:
* We add a WORKSPACE & root BUILD.bazel file
* Add symlinks to gitignore
* Add targets to the makefile so that we run gazelle after dep (and we erase all vendored BUILD / BUILD.bazel files and regenerate them)
* Run gazelle to generate bazel

```release-note
NONE
```


Makefile
```diff
@@ -16,9 +16,24 @@
...
+
+.PHONY: delete-vendored-bazel
+# go libraries often ship BUILD and BUILD.bazel files, but they often don't work.
+# We delete them and regenerate them
+delete-vendored-bazel:
+	find vendor -name "BUILD" -delete
+	find vendor -name "BUILD.bazel" -delete
+
+.PHONY: gazelle
+gazelle:
```
roberthbailey: Should this target depend on delete-vendored-bazel?

justinsb: So the root problem here is that I learned that you can't have comments in the middle of a makefile target :-(  This was my attempt at doing the right thing, but it came up ugly

What I've done before is this:

```
depend:
  dep ensure -v
  # go libraries often ship BUILD and BUILD.bazel files, but they often don't work.
  find vendor -name "BUILD" -delete
  find vendor -name "BUILD.bazel" -delete
  bazel run //:gazelle
```

But I recently learned that the comment is actually passed to the shell, and relies on the shell ignoring it.

OTOH, I think all shells ignore `#` ?

Thoughts?

roberthbailey: According to https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html GNU make (which is what is on my mac at least) uses `/bin/sh` regardless of your environment and sh uses `#` for comments so doing comments in the target should be fine.

apelisse: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_03

I'd be very concerned if a shell wasn't using that part of the posix spec.

justinsb: It's mostly that emacs makes it look like I have done something horrible.  But I'll go with the inline comment :-)

justinsb:
Now with simpler Makefile.  Just don't open it in emacs!

PR #27, opened 2018-07-25
Update cluster-api dependency to latest master

justinsb:
Note that I'm fixing the SHA in gopkg.toml, so it is explicit.

Working towards automatic updating via a PR bot

```release-note
NONE
```


roberthbailey:
> Note that I'm fixing the SHA in gopkg.toml, so it is explicit.

Does this mean that running `dep ensure -update sigs.k8s.io/cluster-api` will no longer do anything (because it won't move forward on the master branch)?

justinsb:
> Does this mean that running dep ensure -update sigs.k8s.io/cluster-api will no longer do anything (because it won't move forward on the master branch)?

Yes, that's exactly the case.

However, my understanding is that `dep ensure -update sigs.k8s.io/cluster-api` is subtle.  If there's a breaking change in the dependency on master that dep can detect (e.g removal of a package), the dependency won't be updated.  My understanding is that `dep ensure -update` can actually rewind a dependency, if I update some other dependency that then forces selection of a prior version of a dependency.

I may be wrong here, but for this reason I've shied away from allowing dep to perform version selection.

PR #32, opened 2018-07-26
Fix typo: boostrap -> bootstrap

justinsb:
Note that one of these is probably a breaking change, because it is on
a label.  But this is the time to take the breakage.

```release-note
Fix typo in bootstrap label
```


justinsb:
I noticed this because I made the same typo!

PR #41, opened 2018-10-09
Update getting started instructions

ntfrnzn:
Fixes a few issues with the getting started process

1. `clusterctl` runs locally and needs authorization to call to GCP.  README.md adds instructions for using the machine-controller serviceaccount
2. machine-controller creates an event in the default namespace when the machine is created.  The rbac template is changed to add full access to the event kind.
3.  `addons.yaml` is missing a version specification, and kubectl fails to apply (with an very unclear error).  `/v1` is added to the template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Participation in the Kubernetes community is governed by the [Kubernetes Code of
 
 ### Cluster Creation
 
-1. Create the `cluster.yaml`, `machines.yaml`, `provider-components.yaml`, and `addons.yaml` files:
+1. Create the `cluster.yaml`, `machines.yaml`, `provider-components.yaml`, and `addons.yaml` files, and create GCP serviceaccounts:
 
    ```bash
    cd cmd/clusterctl/examples/google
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -47,9 +47,11 @@ Participation in the Kubernetes community is governed by the [Kubernetes Code of
    echo "---" >> cmd/clusterctl/examples/google/out/provider-components.yaml
    kustomize build vendor/sigs.k8s.io/cluster-api/config/default/ >> cmd/clusterctl/examples/google/out/provider-components.yaml
    ```
-1. Create a cluster:
+2. Create a cluster, ensuring that the locally-running clusterctl process has access to GCP apis:
```
roberthbailey: nit: Markdown will automatically create the correct numbers. It's convenient to just put `1.` on each line if you later want to move steps around or insert steps, because then a human doesn't need to re-number each line.

README.md
```diff
@@ -47,9 +47,11 @@ Participation in the Kubernetes community is governed by the [Kubernetes Code of
...
 
    ```bash
+   export GOOGLE_APPLICATION_CREDENTIALS=cmd/clusterctl/examples/google/out/machine-controller-serviceaccount.json
```
roberthbailey: This is a neat trick. I wonder if it is worth explaining why it's useful to use this service account in clusterctl.

ntfrnzn: I added a short statement to hint that it's better to use the same credentials that your child processes will use rather than the convenient gcloud setup of the development environment (which might fail and be hard to debug).

Issue #51, opened 2018-11-02
 [FR] authentication with GCP


radkat:
Currently the authentication is done via cloud service account. Allow authentication similar to that in https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/gce/gce.go#L305

PR #52, opened 2018-11-02
Adding authorization via GCP tokens.

radkat:
Also add a flag for machine setup config.

token_source.go was copy-pasted from https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/gce/token_source.go -- let me know if you see a better way of reusing the code than that. 

For GKE integration we need to be able to authenticate via HMS (token-based) like we do in the core Kubernetes. 

Fixes 
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/51

**Release note**:

```release-note
Allow alternative authentication via cloud config.
```


roberthbailey:
Can you split this into multiple commits? It's common practice to put the automatic updates (e.g. vendored dependencies in this case, code gen in others) into a separate commit so that it's clear what was human-changed vs. machine-changed.

Also, it looks like you need to sign the CLA.

radkat:
Abandoning this in favor of https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/53 (split into two commits per the reviewer request).

PR #53, opened 2018-11-06
Adding authorization via GCP tokens.

radkat:
Also add a flag for machine setup config.

token_source.go was copy-pasted from https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/gce/token_source.go -- let me know if you see a better way of reusing the code than that.

For GKE integration we need to be able to authenticate via HMS (token-based) like we do in the core Kubernetes.

Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/51

**Release note**:

```release-note
Allow alternative authentication via cloud config.
```


pkg/cloud/google/machineactuator.go
```diff
@@ -796,19 +800,51 @@ func getOrNewComputeServiceForMachine(params MachineActuatorParams) (GCEClientCo
...
-		return nil, err
+
+	var client *http.Client
+	var err error
+	// If specified in the GCE config, use the alternative authentication.
+	if params.CloudConfigPath != "" {
+		glog.Info("Trying to get open the GCE config")
+		client, err = clientWithAltTokenSource(params.CloudConfigPath)
 	}
+	if params.CloudConfigPath == "" || err != nil {
```
roberthbailey: Would it be better to crash if an invalid config file is provided instead of falling back to the default client? If a file is specified and ends up not being used (due to a parsing error) that seems like a confusing situation to debug.

pkg/cloud/google/machineactuator.go
```diff
@@ -796,19 +800,51 @@ func getOrNewComputeServiceForMachine(params MachineActuatorParams) (GCEClientCo
...
+		glog.Info("Using the default GCP client")
+		if err != nil {
```
roberthbailey: The nesting of this check seems weird, since the above log line will print before this warning.

pkg/cloud/google/machineactuator.go
```diff
@@ -796,19 +800,51 @@ func getOrNewComputeServiceForMachine(params MachineActuatorParams) (GCEClientCo
...
+func clientWithAltTokenSource(gceConfigPath string) (*http.Client, error) {
+	glog.Info("Trying to get the alt token")
+	gceConfig := struct {
+		Global struct {
+			ProjectID string `gcfg:"project-id"`
+			TokenURL  string `gcfg:"token-url"`
+			TokenBody string `gcfg:"token-body"`
+		}
+	}{}
+	if err := warnings.FatalOnly(gcfg.ReadFileInto(&gceConfig, gceConfigPath)); err != nil {
```
roberthbailey: Why not `gcfg.FatalOnly` to avoid the extra import? (That call just proxies to `warnings.FatalOnly` anyway)

I ask because that's how it's done [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/gce/gce.go#L277).

pkg/cloud/google/machineactuator.go
```diff
@@ -796,19 +800,51 @@ func getOrNewComputeServiceForMachine(params MachineActuatorParams) (GCEClientCo
...
+		return nil, err
+	}
+	tokenSource := clients.NewAltTokenSource(gceConfig.Global.TokenURL, gceConfig.Global.TokenBody)
+	client := oauth2.NewClient(context.Background(), tokenSource)
```
roberthbailey: Would `oauth2.NoContext` be better than a background context? Or `context.TODO`?

In case it makes a difference, this week we decided that it would be a good idea to plumb context into the actuator interface, so it might makes sense soon to refactor this such that the http client gets created each time the actuator needs to call out to gcp so that we maintain the calling context down the call stack.

cmd/manager/main.go
```diff
@@ -33,6 +33,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 )
 
+var (
+	cloudConfig        = flag.String("cloud_config", "", "path to the GCE config")
+	machineSetupConfig = flag.String("machine_setup_config", "/etc/machinesetup/machine_setup_configs.yaml", "path to the machine setup config")
```
justinsb: +1 to flags!

I did a consistency check against flag naming in k/k; we seem to use `-` as the separator e.g. `cloud-config`.  (And we do use the name `cloud-config` for the equivalent file in k-c-m, so that's good)

Nit picking, but it's a pain to change later (although I think some of the flag libraries can normalize `_` => `-`?)

radkat: Sure thing! Changed to "-"

pkg/cloud/google/machineactuator.go
radkat: Makes sense. Changed.

pkg/cloud/google/machineactuator.go
radkat: Disappeared because of the above comment.

pkg/cloud/google/machineactuator.go
radkat: I copied over from here: https://github.com/kubernetes/cloud-provider-gcp/blob/1aa2dddbbc427899e0dd5f313829552c0892e247/cmd/gcp-controller-manager/app/options.go#L112

Switching to gcfg.FatalOnly

radkat: Should I remove the vendored stuff now?

pkg/cloud/google/machineactuator.go
radkat: I'd rather submit a change :-)

So if it's okay with you, I'll leave this as is so that "context.TODO" can be an indication for later refactoring for making sure that context is used and used appropriately.

roberthbailey:
Just push commits (unless you want to squash). The bot will squash them.

pkg/cloud/google/machineactuator.go
roberthbailey: context.TODO sounds great. I didn't mean to imply we should change it now, but a TODO makes us remember that we should change it in the future.

pkg/cloud/google/machineactuator.go
roberthbailey: If you re-run `dep` does it remove any vendored code taking out this import?

roberthbailey: You can also run `dep status` to ensure that your dependencies are all properly accounted for.

Issue #50, opened 2018-10-30
Cannot create cluster as HEAD


oxddr:
I cannot create a new cluster using HEAD of this repo. The first issue I hit is https://github.com/kubernetes-sigs/cluster-api/issues/533. I've managed to get bootstrap cluster to work by combination of updating minikube to the newest version and passing `--minikube "kubernetes-version=v1.12.0"` to `clusterctl`. However I still hit the same problem in target cluster, which is running K8s 1.9.4. A dummy changing version anywhere do 1.12.1 did not help (I suspect the startup script should be update somehow). 

Any tips on how to proceed would be welcome.

PR #54, opened 2018-11-09
Add support for K8s v1.12, drop support for v1.9

oxddr:
Some changes to startup-script were required due to changed in kubeadm.

Tested locally in my project.

Fixes: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/50


cmd/clusterctl/examples/google/provider-components.yaml.template
```diff
@@ -253,7 +253,7 @@ data:
             advertiseAddress: ${PUBLICIP}
             bindPort: ${PORT}
           networking:
-            serviceSubnet: ${SERVICE_CIDR}
+            serviceSubnet: ${SERVICE_CIDR}q
```
roberthbailey: This seems like a typo

roberthbailey:
I'm not sure we need to keep 1.9 in for any reason; what about replacing 1.9 with 1.12 instead of doing both (especially since you said 1.9 didn't work for you and there's no real reason to support such an old version)?

cmd/clusterctl/examples/google/provider-components.yaml.template
oxddr: Done

oxddr:
I've removed the support for v1.9. Diffs are much easier to understand now.

cmd/clusterctl/examples/google/machine_setup_configs.yaml
```diff
@@ -1,11 +1,10 @@
-items:
```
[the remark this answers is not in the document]
oxddr: No, thanks for catching that. I apologize for all those typos. I still feel not very comfortable with this new tooling.

PR #58, opened 2018-11-23
Update to latest cluster API

justinsb:
And make the required changes:
* rename provider config -> spec
* add context parameter to actuator

```release-note

```


roberthbailey:
I merged your other PRs and now this needs a rebase....

Gopkg.toml
```diff
@@ -16,6 +16,9 @@ required = [
 [prune]
   go-tests = true
 
+[[constraint]]
+  name="sigs.k8s.io/cluster-api"
+  revision="f8522af25c3a1cbcb304cf1034ec8d40d7153a4c"
```
roberthbailey: Is this necessary to update? Can you just run dep ensure update for this package to update the gopkg.lock instead of inserting a constraint here?

pkg/cloud/google/machineactuator.go
```diff
@@ -140,47 +140,47 @@ func NewMachineActuator(params MachineActuatorParams) (*GCEClient, error) {
 			privateKeyPath: privateKeyPath,
 			user:           user,
 		},
-		client:                   params.Client,
+		client: params.Client,
```
roberthbailey: It seems like this needs a gofmt.

pkg/cloud/google/machineactuator.go
```diff
@@ -236,19 +236,19 @@ func (gce *GCEClient) ProvisionClusterDependencies(cluster *clusterv1.Cluster) e
 	return gce.serviceAccountService.CreateMasterNodeServiceAccount(cluster)
 }
 
-func (gce *GCEClient) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
+func (gce *GCEClient) Create(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
```
roberthbailey: Is ctx unused (for now)? If so, maybe make it `_ context.Context` to make that more obvious to the reader.

justinsb:
Looks like we need to build with go 1.10 or later, for https://github.com/kubernetes-sigs/yaml/blob/master/yaml_go110.go

roberthbailey:
Now that the test-infra PR has gone in, let's see if the build presubmit passes.

justinsb:
It looks like it is testing.  I'm going to rebase and resubmit just to force a retest though :-)

config/crds/gceproviderconfig_v1alpha1_gceclusterproviderspec.yaml
```diff
@@ -4,12 +4,12 @@ metadata:
   creationTimestamp: null
   labels:
     controller-tools.k8s.io: "1.0"
-  name: gceclusterproviderconfigs.gceproviderconfig.k8s.io
+  name: gceclusterproviderspecs.gceproviderconfig.k8s.io
 spec:
   group: gceproviderconfig.k8s.io
```
justinsb: I'm wondering if we should try to have something to do with cluster api in there, and try to be less specific about config.

e.g. cluster-api is `cluster.k8s.io` - should we be `gce.cluster.k8s.io` ?

roberthbailey: I'd be happy to discuss that (the name was just taken verbotem from what we used pre-CRDs, but not that it's more visible to users, it's worth revisiting). Can you open a separate issue though, since bikeshedding on the resource name here isn't going to be very useful.

pkg/cloud/google/machineactuator.go
justinsb: Ooops - missed this comment before.  I just implemented it.

(Although I think we should be passing the context around to e.g. Instances.Get(), now that we have it ... but that's a separate PR)

config/crds/gceproviderconfig_v1alpha1_gceclusterproviderspec.yaml
justinsb: Opened https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/75

justinsb:
Assuming this passes CI, anything blocking this now @roberthbailey ?

PR #61, opened 2018-11-28
Make providerSpec optional, default project

justinsb:
In the absence of defaulting, we can tolerate missing fields.  For the
project, we can just use the "current GCE project" as the default
(where such a thing exists); because that isn't deterministic we might
not actually use for API defaulting anyway.

```release-note
Default providerSpec project value to current GCE project
```


justinsb:
Builds on #58 so I'll need to get that one fixed up first!

roberthbailey:
Needs a rebase.

PR #68, opened 2018-12-03
Provide bootstrap CA cert and config for nodes.

### LINKED ISSUE #51: [FR] authentication with GCP

radkat:
I'd like also to add a sample file for the bootstrap kubeconfig -- what's the best place to put it?

It's essentially this:

```yaml
apiVersion: v1
kind: Config
clusters:
- name: local
  cluster:
    server: http://localhost:8080
contexts:
- context:
    cluster: local
  name: localhost-8080
current-context: localhost-8080
```

For GKE integration, we need to be able to bootstrap a node by having its CA cert and kubeconfig. 

Fixes #51 

**Release note**:

```release-note
Provide bootstrap CA cert and config for nodes.
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -90,11 +90,11 @@ func initStaticDeps(mgr manager.Manager) {
 
 	bootstrapKubeconfig, err := ioutil.ReadFile(*bootstrapKubeconfig)
 	if err != nil {
-		glog.Exitf("Couldn't open bootstrap kubeconfig file %v", err)
+		glog.Warningf("Couldn't open bootstrap kubeconfig file %v", err)
 	}
 	bootstrapCACert, err := ioutil.ReadFile(*bootstrapCACert)
 	if err != nil {
-		glog.Exitf("Couldn't open bootstrap CA cert file %v", err)
+		glog.Warningf("Couldn't open bootstrap CA cert file %v", err)
 	}
 
 	google.MachineActuator, err = google.NewMachineActuator(google.MachineActuatorParams{
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"flag"
+	"io/ioutil"
 	"log"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


radkat:
cc'ing @maisem

cmd/manager/main.go
```diff
@@ -85,7 +88,18 @@ func initStaticDeps(mgr manager.Manager) {
 		glog.Fatalf("Could not create config watch: %v", err)
 	}
 
+	bootstrapKubeconfig, err := ioutil.ReadFile(*bootstrapKubeconfig)
```
justinsb: This effectively makes these flags required - is that deliberate?

radkat: Nope! Thanks for catching.

radkat:
Ping?

cmd/manager/main.go
justinsb: Can we skip trying to load this if `*bootstrapKubeconfig == ""` ?  (because otherwise we'll spew a warning)

cmd/manager/main.go
```diff
@@ -85,7 +88,18 @@ func initStaticDeps(mgr manager.Manager) {
...
+	if err != nil {
+		glog.Warningf("Couldn't open bootstrap kubeconfig file %v", err)
```
justinsb: Nit: generally we start error messages with a lower case letter (`couldn't open..`) and we end with `: %v`.  (...`kubeconfig file: %v`).  It gives something that looks a bit like a nested exception.

cmd/manager/main.go
```diff
@@ -85,7 +88,18 @@ func initStaticDeps(mgr manager.Manager) {
...
+	}
+	bootstrapCACert, err := ioutil.ReadFile(*bootstrapCACert)
```
justinsb: Same here

pkg/cloud/google/metadata.go
```diff
@@ -40,19 +44,21 @@ type metadataParams struct {
...
 	params := metadataParams{
-		Token:          token,
-		Cluster:        cluster,
-		Machine:        machine,
-		Project:        project,
-		Metadata:       metadata,
-		PodCIDR:        getSubnet(cluster.Spec.ClusterNetwork.Pods),
-		ServiceCIDR:    getSubnet(cluster.Spec.ClusterNetwork.Services),
-		MasterEndpoint: getEndpoint(cluster.Status.APIEndpoints[0]),
+		BootstrapKubeconfig: base64.StdEncoding.EncodeToString(bootstrapKubeconfig),
```
justinsb: TIL that nil -> `""`!  👍

pkg/cloud/google/metadata.go
```diff
@@ -125,6 +131,8 @@ const nodeEnvironmentVars = `
 #!/bin/bash
 KUBELET_VERSION={{ .Machine.Spec.Versions.Kubelet }}
 TOKEN={{ .Token }}
+CA_CERT={{ .BootstrapCACert }}
+KUBECONFIG={{ .BootstrapKubeconfig }}
```
justinsb: I'm assuming you're using a different script here.

justinsb:
Sorry for delay @radkat - what do you think about avoiding loading the file if it isn't specified (or at least, warning about it!)

PR #73, opened 2018-12-17
Add an option for creating Machines using InstanceTemplate.

oxddr:

Add an option to create Machine objects using given InstanceTemplate. 

This is my first bigger change to Kubernetes-related project, so please be extra careful - I might have implement things completely wrong. Also feel free to be extra picky - so I can learn even more from a review. 

I am not sure what tests to add, so please point it to me. It was hard for me to induct this from current code.

### DIFF
--- a/pkg/apis/gceproviderconfig/v1alpha1/gcemachineproviderconfig_types.go
+++ b/pkg/apis/gceproviderconfig/v1alpha1/gcemachineproviderconfig_types.go
@@ -32,11 +32,13 @@ type GCEMachineProviderSpec struct {
 	Roles []MachineRole `json:"roles,omitempty"`
 
 	Zone        string `json:"zone"`
-	MachineType string `json:"machineType"`
+	MachineType string `json:"machineType,omitempty"`
 
 	// The name of the OS to be installed on the machine.
 	OS    string `json:"os,omitempty"`
 	Disks []Disk `json:"disks,omitempty"`
+
+	InstanceTemplate string `json:"instanceTemplate,omitempty"`
 }
 
 // The MachineRole indicates the purpose of the Machine, and will determine
--- a/pkg/cloud/google/machineactuator.go
+++ b/pkg/cloud/google/machineactuator.go
@@ -294,7 +294,7 @@ func (gce *GCEClient) createFromInstanceTemplate(cluster *clusterv1.Cluster, clu
 	project := clusterConfig.Project
 	zone := machineConfig.Zone
 
-	return gce.computeService.InstancesInsertFromTemplate(project, zone, name, fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/global/instanceTemplates/%s", project, machineConfig.InstanceTemplate))
+	return gce.computeService.InstancesInsertFromTemplate(project, zone, name, machineConfig.InstanceTemplate)
 }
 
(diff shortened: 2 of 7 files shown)


pkg/apis/gceproviderconfig/v1alpha1/gcemachineproviderconfig_types.go
```diff
@@ -32,11 +32,13 @@ type GCEMachineProviderConfig struct {
 	Roles []MachineRole `json:"roles,omitempty"`
 
 	Zone        string `json:"zone"`
-	MachineType string `json:"machineType"`
+	MachineType string `json:"machineType,omitempty"`
```
maisem: Can we also add project here?

pkg/cloud/google/machineactuator.go
```diff
@@ -255,93 +255,109 @@ func (gce *GCEClient) Create(cluster *clusterv1.Cluster, machine *clusterv1.Mach
...
+	if gce.client != nil {
+		return gce.updateAnnotations(cluster, machine)
+	}
+	return nil
+}
+
+func (gce *GCEClient) createFromInstanceTemplate(cluster *clusterv1.Cluster, clusterConfig *gceconfigv1.GCEClusterProviderConfig, machine *clusterv1.Machine, machineConfig *gceconfigv1.GCEMachineProviderConfig) (*compute.Operation, error) {
+	name := machine.ObjectMeta.Name
+	project := clusterConfig.Project
+	zone := machineConfig.Zone
```
maisem: inline?

pkg/cloud/google/machineactuator.go
```diff
@@ -255,93 +255,109 @@ func (gce *GCEClient) Create(cluster *clusterv1.Cluster, machine *clusterv1.Mach
 		return gce.handleMachineError(machine, verr, createEventAction)
 	}
```
maisem: Can you also move the machineSetupConfigGetter checks to create? If an instanceTemplate is provided we don't need to rely on machineSetupConfigs.

pkg/cloud/google/machineactuator.go
```diff
@@ -255,93 +255,109 @@ func (gce *GCEClient) Create(cluster *clusterv1.Cluster, machine *clusterv1.Mach
...
```
maisem: If we add a project field to machine provider spec we can remove the reliance on the cluster object as well if the machine was created via an instanceTemplate.

pkg/apis/gceproviderconfig/v1alpha1/gcemachineproviderconfig_types.go
oxddr: Project is part of cluster config. I don't think we need to duplicate this here. Or do you have another use case in mind?

pkg/cloud/google/machineactuator.go
oxddr: I recall `project` was part of machine config once, but it was removed for some reasons (I don't know them, code/config duplication maybe?). In all other places (e.g. in machine actuator), reconciliation of Cluster API relies on `Cluster` anyway.

oxddr:
Friendly ping :)

oxddr:
/assign @justinsb  

Could you please review this PR?

detiber:
@oxddr are you still interested in this change?

oxddr:
Not personally, this could be useful change though. I fine with both closing it or having rebased and possible merged.

detiber:
@oxddr ok, I'm going to close and we can potentially revisit after the v1alpha2 changes land.
/close

PR #77, opened 2018-12-22
Changes after running make on a clean checkout.

roberthbailey:
**What this PR does / why we need it**: fixes formatting on one file; adds a few vendored diffs. 

Fixes #

**Release note**:

```release-note
NONE
```

/assign @justinsb


justinsb:
The gofmt changes are go 1.10 vs 1.11, but at this point we probably should be using go 1.11.

Not sure about the vendor changes TBH, but it's pretty low consequence if they in fact aren't needed (just another PR).

/lgtm

PR #107, opened 2019-01-26
Refactor initStaticDeps to return an error instead of panic

justinsb:
panic is generally a code-smell, IMO.

```release-note
NONE
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"log"
 
+	"github.com/pkg/errors"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"k8s.io/klog"
 	"sigs.k8s.io/cluster-api-provider-gcp/pkg/apis"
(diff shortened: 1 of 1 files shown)


cmd/manager/main.go
```diff
@@ -65,7 +66,9 @@ func main() {
 	}
 
 	log.Printf("Initializing Dependencies.")
-	initStaticDeps(mgr)
+	if err := initStaticDeps(mgr); err != nil {
+		log.Fatal(err)
```
roberthbailey: should this be klog?

BenTheElder: all logging in main appears to be `log` rather than `klog` currently... 🤔 

@justinsb perhaps we should change this?

BenTheElder:
> How does moving the fatal call one level up the stack prevent panic? Is it that klog panics and log doesn't?

IMHO, it's just that the only `exit()`-equivalent calls should occur in `main()`, that way the rest is re-usable, testable, etc, and only the actual application binary ever calls exit / panic / fatal / ...

detiber:
/cancel-lifecycle rotten
/hold cancel

This looks fine, any additional changes can be addressed through followup (re klog vs log)

PR #108, opened 2019-02-12
Update cluster-api dependency

vincepri:

This PR updates the cluster-api upstream dependency and the rest of the vendor folder, it also removes glog and fixes some breaking changes. The machine objects now reference the cluster object by using `cluster.k8s.io/cluster-name` label.

Fixes #

**Release note**:

```release-note

```


vincepri:
Waiting for https://github.com/kubernetes/test-infra/pull/11290

cmd/clusterctl/main.go
```diff
@@ -17,27 +17,13 @@ limitations under the License.
...
-	flags := flag.NewFlagSet("klog", flag.ExitOnError)
-	klog.InitFlags(flags)
-	flags.Set("alsologtostderr", "true")
-	flags.Set("v", "4")
-	flag.Parse()
-}
-
 func main() {
-	initLogs()
 	var err error
```
justinsb: We don't need klog.InitFlags(nil) here?

pkg/cloud/google/machineactuator.go
```diff
@@ -28,21 +28,19 @@ import (
 	"strings"
 	"time"
 
-	"github.com/golang/glog"
+	"github.com/ghodss/yaml"
```
justinsb: FYI I think we have `sigs.k8s.io/yaml` that is our preferred fork...

vincepri: +1

cmd/clusterctl/main.go
vincepri: The flags are already initialized upstream as part of `cmd.Execute`.

vincepri:
@justinsb Rebased, updated with feedback and pushed :)

vincepri:
@justinsb @roberthbailey ptal.

Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY cmd/    cmd/
 COPY vendor/ vendor/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager sigs.k8s.io/cluster-api-provider-gcp/cmd/manager
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -ldflags '-extldflags "-static"' -o manager sigs.k8s.io/cluster-api-provider-gcp/cmd/manager
```
justinsb: It is weird that we do this, but ... I agree.  The `cc` distroless image does include libc and thus avoids the need for static linking, but this feels better.  I think there's some limitations around static linking on osx, but we're in docker anyway...

justinsb:
Thanks @vincepri 

/approve
/lgtm

PR #111, opened 2019-03-04
Fixed disk and version problem specially for node creation from machine set

masudur-rahman:
…ne set

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/google/machineactuator.go
+++ b/pkg/cloud/google/machineactuator.go
@@ -33,12 +33,14 @@ import (
 	"golang.org/x/oauth2/google"
 	compute "google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
+
 	gcfg "gopkg.in/gcfg.v1"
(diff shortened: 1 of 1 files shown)


justinsb:
Hi @masudur-rahman - these look like good changes.  Are you able to sign the CLA?

I'd also suggest splitting them into separate changes; we'd likely want to discuss them separately.  For example, logging the machine error earlier seems like an obvious improvement.  Defaulting the disks seems helpful, but it suggests that we're not treating machines.yaml as correct - maybe we should treat that as an error.  Stripping the characters from the version seems good, but will this make it harder to use CI builds (for example), where we might reference them by URL? 

Splitting them up lets us have these discussions at their own pace, and stops one discussion holding up the other (good) changes!

masudur-rahman:
Hi @justinsb, I created separate PR for the changes as you asked. Can you please check if anything needs to be changed ?

PR #114, opened 2019-03-19
Attempt to patch up docker version and make consistent between ControlPlane and Nodes

detiber:
Attempt to patch up docker version and make consistent between ControlPlane and Nodes

**Release note**:
```release-note
NONE
```

### DIFF
--- a/cmd/clusterctl/examples/google/machine_setup_configs.yaml
+++ b/cmd/clusterctl/examples/google/machine_setup_configs.yaml
@@ -50,9 +50,11 @@ items:
 
       install_configure_docker
 
-      curl -sf https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
-      touch /etc/apt/sources.list.d/kubernetes.list
-      sh -c 'echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" > /etc/apt/sources.list.d/kubernetes.list'
+      curl -fs https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
+      cat <<EOF > /etc/apt/sources.list.d/kubernetes.list
+      deb http://apt.kubernetes.io/ kubernetes-xenial main
+      EOF
+      apt-get update
 
       curl -fsSL https://dl.k8s.io/release/${VERSION}/bin/linux/${ARCH}/kubeadm > /usr/bin/kubeadm.dl
       chmod a+rx /usr/bin/kubeadm.dl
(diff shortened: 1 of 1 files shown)


detiber:
Completely untested as of right now.

cmd/clusterctl/examples/google/machine_setup_configs.yaml
```diff
@@ -28,30 +28,32 @@ items:
...
-      touch /etc/apt/sources.list.d/kubernetes.list
-      sh -c 'echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" > /etc/apt/sources.list.d/kubernetes.list'
       apt-get update -y
-      apt-get install -y \
-        socat \
-        ebtables \
-        apt-transport-https \
-        cloud-utils \
-        prips
+      apt-get install -y socat ebtables apt-get install apt-transport-https ca-certificates curl software-properties-common cloud-utils prips
```
justinsb: Missing newline I think?

cmd/clusterctl/examples/google/machine_setup_configs.yaml
```diff
@@ -28,30 +28,32 @@ items:
...
-        apt-get install -y docker.io
-        echo 'DOCKER_OPTS="--iptables=false --ip-masq=false"' > /etc/default/docker
-        systemctl daemon-reload
-        systemctl enable docker
-        systemctl start docker
+          # prevent docker from auto-starting
+          echo "exit 101" > /usr/sbin/policy-rc.d
+          chmod +x /usr/sbin/policy-rc.d
+          trap "rm /usr/sbin/policy-rc.d" RETURN
+          apt-get install -y docker-ce=18.06.2~ce~3-0~ubuntu
```
justinsb: I do wonder if we should just use the tar.gz packaging - the docker packages seem to cause as many problems as they solve.  We support it in kops, but only do it for OSes where docker doesn't package.  But when we go to containerd, I think we're going to try switching.

detiber: I was mainly trying to patch things up enough to get them working again without too much disruption :)

cmd/clusterctl/examples/google/machine_setup_configs.yaml
detiber: Apparently `apt-get install` doesn't know how to install `apt-get` and `install`, fixed up in the latest version :joy:

detiber:
@justinsb this should be good to go now. This doesn't completely fix up the `clusterctl create` workflow, but it does get further now.

There appears to be an issue during pivoting where the gcp provider controller manager fails to come up, but I don't currently have time to debug further.

justinsb:
Going to merge this, as it's moving us in the right direction!

/approve
/lgtm

PR #116, opened 2019-03-20
Update README.md - Change cluster create command to use bootstrap option instead of minikube.

meirlaker:
Change cluster create syntax to use bootstrap instead of minikube option.

The kubernetes version required by ClusterAPI was previously passed using the --minikube option to the cluster create command. But, the --minikube option was removed some time ago from the cluster create command, in favor of using --bootstrap options to pass the kubernetes version.

Using the old syntax now fails with no explanation to the user as to how to solve the issue.

This PR updates the README so that users will conform to the correct command syntax.


justinsb:
Thanks!

/approve
/lgtm

PR #118, opened 2019-03-21
Add default disk

masudur-rahman:


detiber:
@masudur-rahman it looks like the tests need to be updated for this change.

masudur-rahman:
> @masudur-rahman it looks like the tests need to be updated for this change.

Hi @detiber, I have updated the test. Can you please check if it's okay ?

justinsb:
Thanks @masudur-rahman - this does seem reasonable.  There is a broader discussion of whether we should default values - is the MachineDeployment expected to be something users interact with directly.  However, this is a good change, and doesn't really force our hand in either direction - we can't have machines without a boot disk.

/approve
/lgtm

PR #119, opened 2019-03-21
Add NodeRole if nodeRole not provided

masudur-rahman:


pkg/cloud/google/machineactuator.go
```diff
@@ -252,6 +252,18 @@ func (gce *GCEClient) Create(_ context.Context, cluster *clusterv1.Cluster, mach
...
+	if !isMaster(machineConfig.Roles) {
+		isNodeSet := false
+		for _, r := range machineConfig.Roles {
+			if r == gceconfigv1.NodeRole {
+				isNodeSet = true
+				break
+			}
+		}
+		if !isNodeSet {
+			machineConfig.Roles = append(machineConfig.Roles, gceconfigv1.NodeRole)
```
justinsb: Changing this in place does mean it will be persisted back to k8s if we do an update.  I don't know if we should do that - I can see the argument for both sides.

justinsb:
It's not clear what the future of NodeRole is, but I don't think we should add a role of Node whenever something is not a master - it means we could never have another role.  (kops has bastions, openshift has ingress nodes etc).

I don't mind in the code assuming that anything that is not a master is a node, for now, because that's easier to change, but I don't think we should be creating the assumption that `!master` => `node` by updating the roles collection.

alejandrox1:
Do you have any additional comments @masudur-rahman ?

@vincepri should this PR be kept open? doesn't seem like a patch that may be appropriate in the long term.

PR #120, opened 2019-03-21
Strip `v` from `kubelet` and `controlplane` version

masudur-rahman:
In some cases `kubelet` and `controlplane` versions are copied from `Kubernetes` version. `Kubernetes` version contains an extra `v` from `kubelet & controlplane` version. That's why I striped `v` from `kubelet` and `controlplane` version.

### DIFF
--- a/pkg/cloud/google/machineactuator.go
+++ b/pkg/cloud/google/machineactuator.go
@@ -646,6 +646,8 @@ func (gce *GCEClient) machineproviderconfig(providerSpec clusterv1.ProviderSpec)
 */
 
 func (gce *GCEClient) updateMasterInplace(cluster *clusterv1.Cluster, oldMachine *clusterv1.Machine, newMachine *clusterv1.Machine) error {
+	oldMachine.Spec.Versions.ControlPlane = stripVersion(oldMachine.Spec.Versions.ControlPlane)
+	newMachine.Spec.Versions.ControlPlane = stripVersion(newMachine.Spec.Versions.ControlPlane)
 	if oldMachine.Spec.Versions.ControlPlane != newMachine.Spec.Versions.ControlPlane {
 		cmd := fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


detiber:
I don't know enough about this workflow to comment on this change.
/assign @justinsb

justinsb:
This seems sensible to me, but I propose two things:

1) We avoid writing it back into the API object, but rather call stripVersion when we want to read the field
2) We discuss this in the cluster-api meeting on Wednesday.  We want this to be consistent across providers.  We could make this one of the first pieces of shared code in this area.  Alternatively we could validate a strict format e.g. "no v"; a reason for that would be to make it easier for other code to read the version from the Cluster/Machine object (without having to maintain the normalization logic)

pkg/cloud/google/machineactuator.go
```diff
@@ -689,7 +693,14 @@ func (gce *GCEClient) updateMasterInplace(cluster *clusterv1.Cluster, oldMachine
 	return nil
 }
 
+func stripVersion(version string) (newVersion string) {
+	cutSet := "vV"
+	newVersion = strings.Trim(version, cutSet)
+	return
```
vincepri: This could be a one liner

```suggestion
	return strings.Trim(version, "vV")
```

masudur-rahman:
Hi @justinsb, can you please recheck if the changes are okay ?

pkg/cloud/google/machineactuator.go
```diff
@@ -689,6 +692,10 @@ func (gce *GCEClient) updateMasterInplace(cluster *clusterv1.Cluster, oldMachine
 	return nil
 }
 
+func stripVersion(version string) string {
+	return strings.Trim(version, "vV")
```
justinsb: Probably should be `TrimLeft`, but in practice it's not going to end with v anyway

justinsb:
Thanks for the changes @masudur-rahman .  As I see it now, the primary case is to make sure that we don't specify a `v` in front of the version we pass to kubeadm or apt-get, and it looks narrowly scoped to that.

/approve
/lgtm

PR #123, opened 2019-03-26
feat: support additional metadata for instances

### LINKED ISSUE #122: Support additional metadata for instances
As I've been getting familiar with the GCP cluster-api provider, I've noticed that there is really only a "startup-script" that can get passed in as a metadata item for the instances that get created. I'd love to be able to squirrel away other metadata as desired. I've got a PR that I'll be chasing this issue with that provides this functionality.

rsmitty:
Interested to see what folks think about this approach. This PR would add the ability to supply additional metadata to instances that get created by the provider. As it currently stands, there's only the "startup-script" metadata field that can be edited. I'd like to be able to squirrel other data away in metadata during instance creation time. This PR allows metadata to be specified as a list with name/value fields, like:

```bash
  metadata:
  - name: "startup-script"
    value: |
      #!/bin/bash
      echo "this is metadata"
  - name: "additional-data"
    value: |
      this: "just a test"
```
It allows us to keep the "startup-script" in it's current form, while also allowing for additions.

Fixes #122

### DIFF
--- a/cmd/clusterctl/examples/google/machine_setup_configs.yaml
+++ b/cmd/clusterctl/examples/google/machine_setup_configs.yaml
@@ -8,7 +8,8 @@ items:
       controlPlane: 1.12.0
   image: projects/ubuntu-os-cloud/global/images/family/ubuntu-1604-lts
   metadata:
-    startupScript: |
+  - name: "startup-script"
+    value: |
       set -e
       set -x
       (
@@ -155,7 +156,8 @@ items:
       kubelet: 1.12.0
   image: projects/ubuntu-os-cloud/global/images/family/ubuntu-1604-lts
   metadata:
-    startupScript: |
+  - name: "startup-script"
+    value: |
       set -e
       set -x
       (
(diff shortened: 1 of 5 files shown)


rsmitty:
To maybe add some additional color to what I'm aiming for here. Not sure if this will confuse or help. I'm working on creating a [talos](https://github.com/autonomy/talos) cluster-api provider. But instead of booting instances myself for each cloud environment, I'm hoping to just wrap the cluster-api for that given provider. We currently provision the talos nodes with userdata, but it looks for a metadata path of `user-data`, not `startup-script`. Plus it expects yaml, so it wouldn't know what to do with the generated exports anyways. So I'm kind of hoping to make the `startup-script` a no-op essentially and inject the things I need to the `user-data` path.

justinsb:
Thanks @rsmitty - this looks good.  I can see a debate about whether we should allow templating etc, and if you look at my stream of recent PRs you can also see that I want to allow this metadata to be specified in the MachineDeployment, which ends up making `machine_setup_config.yaml` optional.

But this is a big step forwards for machine_setup_config.yaml, and unlocks your use-case and I am sure others!

Thanks for the PR :-)

/approve
/lgtm

PR #125, opened 2019-03-26
feat: support externalIPs for machines

### LINKED ISSUE #124: Support External IPs
I'm interested in having some defined external IPs that I can attach to my master instances, so that I can know how to reach them ahead of time. I'll be following this issue with a PR that provides code to do this so we can get started talking about how folks feel about it.

rsmitty:
This PR would add the ability for pre-generated external IPs to be specified in the machine manifest for masters (or really any instance type). One of the things I'm interested in is generating some static-ish machine manifests for each master that provides this externalIP field. An example that's similar to what I've been using:

```yaml
apiVersion: "cluster.k8s.io/v1alpha1"
kind: Machine
metadata:
  generateName: gce-master-
  labels:
    cluster.k8s.io/cluster-name: test1-7s5v2
    set: master
spec:
  providerSpec:
    value:
      apiVersion: "gceproviderconfig/v1alpha1"
      kind: "GCEMachineProviderSpec"
      roles:
      - Master
      zone: "us-central1-f"
      machineType: "n1-standard-1"
      os: "ubuntu-1604-lts"
      disks:
      - initializeParams:
          diskSizeGb: 30
          diskType: "pd-standard"
      externalIP: "111.222.333.444"
  versions:
    kubelet: 1.12.0
    controlPlane: 1.12.0
``` 
Similar to #123, I'm hoping to have this ability so that I can wrap the GCP provider for use with talos, where we will pre-generate certs for the master IPs that then get pushed into the nodes as userdata. Also want to note that adding these external IPs would be totally optional and creating a machine manifest without it still works just fine.

Fixes #124

### DIFF
--- a/config/crds/gceproviderconfig_v1alpha1_gcemachineproviderspec.yaml
+++ b/config/crds/gceproviderconfig_v1alpha1_gcemachineproviderspec.yaml
@@ -37,6 +37,8 @@ spec:
             - initializeParams
             type: object
           type: array
+        externalIP:
+          type: string
         kind:
           description: 'Kind is a string value representing the REST resource this
             object represents. Servers may infer this from the endpoint the client
--- a/pkg/apis/gceproviderconfig/v1alpha1/gcemachineproviderconfig_types.go
+++ b/pkg/apis/gceproviderconfig/v1alpha1/gcemachineproviderconfig_types.go
@@ -35,8 +35,9 @@ type GCEMachineProviderSpec struct {
 	MachineType string `json:"machineType"`
 
 	// The name of the OS to be installed on the machine.
-	OS    string `json:"os,omitempty"`
-	Disks []Disk `json:"disks,omitempty"`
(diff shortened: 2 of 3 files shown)


justinsb:
Thanks @rsmitty  - this looks really good.  If you look at https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/130, you can see there I am adding some more networking options.  I'm wondering if we should put externalIP into the `networkInterfaces` structure?

I can also see the argument for just keeping-it-simple as a top-level field though - what do you think?

detiber:
@rsmitty Do you plan on rebasing this and addressing the feedback from @justinsb?

rsmitty:
Apologies, just saw this. I'll probably just close this PR out for now. We wound up not using the GCP provider under the hood in Talos. Once the infra provisioners are solidified, maybe we'll revisit.

PR #132, opened 2019-05-19
machineactuator: Get cluster from metadata if not set

justinsb:
This lets us avoid requiring the cluster in the machineactuator,
particularly as we seem to only need it for the project.

```release-note
NONE
```


alejandrox1:
@justinsb any updates on this one? Is there anything we can do to help?

PR #143, opened 2019-09-09
✨Support v1alpha2

vincepri:
This PR updates the repository to be in line with other Cluster API providers and support v1alpha2 API types.

The commits have been separated to provide a better way to review the large changes.

NB: Before merging this in, we should cut a `release-0.1` branch.

PR #145, opened 2019-09-11
Follow-up fixes from #143

vincepri:

- Removes the Cluster.Spec.DefaultZone
- Makes sure to set the NetworkID as part of the instance's Network Interface definition
- Updates CAPI to latest master
- Uses the record package to record events

Fixes #


cloud/services/compute/instances.go
```diff
@@ -68,11 +68,13 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*compute.Instance,
...
 	input := &compute.Instance{
-		Name:              scope.Name(),
-		Zone:              scope.Zone(),
-		MachineType:       fmt.Sprintf("zones/%s/machineTypes/%s", scope.Zone(), scope.GCPMachine.Spec.InstanceType),
-		CanIpForward:      true,
-		NetworkInterfaces: []*compute.NetworkInterface{{}},
+		Name:         scope.Name(),
+		Zone:         scope.Zone(),
+		MachineType:  fmt.Sprintf("zones/%s/machineTypes/%s", scope.Zone(), scope.GCPMachine.Spec.InstanceType),
+		CanIpForward: true,
```
justinsb: BTW if we're using Calico I don't think we need CanIpForward...

justinsb:
Thanks for the follow-up @vincepri !

/approve
/lgtm

main.go
```diff
@@ -133,18 +134,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Initialize event recorder.
+	record.InitFromRecorder(mgr.GetEventRecorderFor("aws-controller"))
```
tahsinrahman: ```gcp-controller```

vincepri: whoops! Are you cool with fixing it as part of the other PR you have open?

PR #155, opened 2019-09-13
✨ Add the ability to customize Service Account and Scopes

### LINKED ISSUE #150: Add the ability to customize Service Account and Scopes
/kind feature
/help
/good-first-issue
/milestone v0.2

The current implementation has hardcoded the service account and its scope https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/420b2cea43922672bdaae9b69eb66045ff33b41b/cloud/services/compute/instances.go#L103-L112. 

We need to add a way for users to specify their own service account email and which scopes to assign to the machine. A new optional struct needs to be added to the Machine.Spec object, in case the user doesn't specify anything, we can default to the current code.

In addition, we'd need to document the minimum amount of scope necessary for Cluster API to work.

tahsinrahman:

Fixes #150


cloud/services/compute/instances.go
```diff
@@ -102,15 +102,18 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*compute.Instance,
 				},
 			},
 		},
-		ServiceAccounts: []*compute.ServiceAccount{
+	}
+	if len(scope.GCPMachine.Spec.ServiceAccounts) > 0 {
+		input.ServiceAccounts = scope.GCPMachine.Spec.ServiceAccounts
+	} else {
```
vincepri: I'd remove the need for the else by keeping the defaults in the struct (like it was before), and overwriting it if there is one that's specified.

vincepri:
@tahsinrahman I think we might need to copy the struct here to accomodate the deep copy?

api/v1alpha2/types.go
```diff
@@ -195,3 +195,29 @@ var (
...
+	// used to include empty fields in Patch requests.
+	ForceSendFields []string `json:"-"`
+
+	// NullFields is a list of field names (e.g. "Email") to include in API
+	// requests with the JSON null value. By default, fields with empty
+	// values are omitted from API requests. However, any field with an
+	// empty value appearing in NullFields will be sent to the server as
+	// null. It is an error if a field in this list has a non-empty value.
+	// This may be used to include null fields in Patch requests.
+	NullFields []string `json:"-"`
```
vincepri: We can probably remove the NullFields and ForceSendFields

api/v1alpha2/gcpmachine_types.go
```diff
@@ -84,7 +83,7 @@ type GCPMachineSpec struct {
 	// ServiceAccounts specify the service account email and which scopes to assign to the machine.
 	// Defaults to: email: "default", scope: []{compute.CloudPlatformScope}
 	// +optional
-	ServiceAccounts []*compute.ServiceAccount `json:"serviceAccounts,omitempty"`
+	ServiceAccounts []ServiceAccount `json:"serviceAccounts,omitempty"`
```
vincepri: There can be only a single ServiceAccount associated to a Machine, we should switch this to a pointer, wdyt?

PR #157, opened 2019-09-13
🐛 fix: remove leading v from kube version

### LINKED ISSUE #156: Error when creating machines b/c of k8s version
/kind bug

The examples that get generated don't work out of the box once you create the controlplane. This seems to be failing on the semver parsing of the kubernetes version, specifically because of the leading `v`. Here's the error I encountered:

```
E0913 13:50:43.730099       1 controller.go:218] controller-runtime/controller "msg"="Reconciler
 error" "error"="failed to create GCPMachine instance: error parsing Spec.Version on Machine 
\"test1-controlplane-0\" in namespace \"default\", expected valid SemVer string: Invalid character(s) 
found in major number \"v1\""  "controller"="gcpmachine" "request"=
{"Namespace":"default","Name":"test1-controlplane-0"}
```

Control plane gets created.

We don't make use of this value in the talos provider, so I can't really say whether the `v` is required in the kubeadm bootstrapper code, but removing it seems to cause the machine creation to pass the semver check. Will follow this issue with a PR of the small change.

rsmitty:
This PR fixes #156 by removing the v from KUBERNETES_VERSION and
allowing it to pass the semver check during machine creation.

Fixes #156

### DIFF
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -26,7 +26,7 @@ command -v "${ENVSUBST}" >/dev/null 2>&1 || echo -v "Cannot find ${ENVSUBST} in 
 
 # Cluster.
 export CLUSTER_NAME="${CLUSTER_NAME:-test1}"
-export KUBERNETES_VERSION="${KUBERNETES_VERSION:-v1.15.3}"
+export KUBERNETES_VERSION="${KUBERNETES_VERSION:-1.15.3}"
 
 # Machine settings.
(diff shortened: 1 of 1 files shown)


vincepri:
@rsmitty We should check and strip the `v` from the version from the code that parses the version instead. Every other example in Cluster API uses the `vX.Y.Z`

rsmitty:
Okay, that seems fair. I'll PR that shortly.

PR #158, opened 2019-09-13
🐛 fix: fix reading of semver for kube version 

### LINKED ISSUE #156: Error when creating machines b/c of k8s version

rsmitty:

This PR updates how we parse the kubernetes version in the
instances.go and allows it to pass the semver check during machine creation.

Fixes #156

### DIFF
--- a/cloud/services/compute/instances.go
+++ b/cloud/services/compute/instances.go
@@ -61,7 +61,7 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*compute.Instance, 
 			scope.Name(), scope.Namespace())
 	}
 
-	version, err := semver.Make(*scope.Machine.Spec.Version)
+	version, err := semver.ParseTolerant(*scope.Machine.Spec.Version)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error parsing Spec.Version on Machine %q in namespace %q, expected valid SemVer string",
 			scope.Name(), scope.Namespace())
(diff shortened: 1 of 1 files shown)


rsmitty:
@vincepri This should be closer to what you were looking for. Still allows for input of `v1.x.y` in the yaml files.

cloud/services/compute/instances.go
```diff
@@ -61,7 +62,11 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*compute.Instance,
 			scope.Name(), scope.Namespace())
 	}
 
-	version, err := semver.Make(*scope.Machine.Spec.Version)
```
jdef: Instead of manual parsing, maybe use `ParseTolerant` (instead of `Make`).

rsmitty:
@jdef @vincepri okay, updated to use parse tolerant. feels much cleaner, thanks for the pointer to that function! also added in the `v` for the default source image.

PR #172, opened 2019-09-20
🐛 Reconcile instance groups before updating backends

vincepri:

Fixes #


cloud/services/compute/loadbalancers.go
```diff
@@ -85,10 +85,6 @@ func (s *Service) ReconcileLoadbalancers() error {
 
 	s.scope.Network().APIServerBackendService = pointer.StringPtr(backendService.SelfLink)
 
-	if err := s.UpdateBackendServices(); err != nil {
```
vincepri: This has been removed because we call it from the machine actuator, there isn't any need to call it again here, which might lead into errors / races.

PR #177, opened 2019-09-27
Harden ci scripts and get e2e working

dims:
- Add a router and cloud nat to ensure that the VM(s) can access the
    internet to pull debs and docker images
- use a python based `envsubst` command for generating example
    configuration and avoid installing yet another binary
 - correct resource type in boskos is `"gce-project"`
 - Ability to dump logs (loop through all the instances and try getting
    as much information to save in log files)
 - Add timeouts for things that may get stuck
 - install ansible if it is missing (or prompt user if not running as
    root)
 - create a new user id for running the packer based build for image when
    running in the CI environment. packer/ansible fails otherwise.
 - Add loops to wait for things to start up before running tests
 - Add more logs to be sure we checkin/checkout projects from boskos
 - Also install kustomize

### DIFF
--- a/hack/ci/e2e-conformance.sh
+++ b/hack/ci/e2e-conformance.sh
@@ -25,47 +25,143 @@ CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
 
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 
-# our exit handler (trap)
-cleanup() {
+# dump logs from kind and all the nodes
+dump-logs() {
   # always attempt to dump logs
   kind "export" logs --name="clusterapi" "${ARTIFACTS}/logs" || true
+
+  # iterate through any nodes we brought up and collect logs
+  gcloud compute instances list --project "${GCP_PROJECT}" --format='value(name, zone)' \
+  | grep "${CLUSTER_NAME}" | while read node_name node_zone; do
+    echo "collecting logs from ${node_name}"
+    dir="${ARTIFACTS}/logs/${node_name}"
+    mkdir -p ${dir}
+
+    gcloud compute instances get-serial-port-output --project "${GCP_PROJECT}" \
+      --zone "${node_zone}" --port 1 "${node_name}" > "${dir}/serial-1.log" || true
+
+    ssh-to-node "${node_name}" "${node_zone}" "sudo chmod -R a+r /var/log" || true
+    gcloud compute scp --recurse --project "${GCP_PROJECT}" --zone "${node_zone}" \
+      "${node_name}:/var/log/cloud-init.log" "${node_name}:/var/log/cloud-init-output.log" "${dir}" || true
+
+    ssh-to-node "${node_name}" "${node_zone}" "sudo journalctl --output=short-precise" > "${dir}/systemd.log" || true
+  done
+
+  timeout 300 gcloud logging read --order=asc \
+    --freshness="3h" \
+    --format='table(timestamp,jsonPayload.resource.name,jsonPayload.event_subtype)' \
+    --project "${GCP_PROJECT}" > "${ARTIFACTS}/logs/activity.log" || true
+}
+
+# our exit handler (trap)
+cleanup() {
+  # dump all the logs
+  dump-logs
+
   # KIND_IS_UP is true once we: kind create
   if [[ "${KIND_IS_UP:-}" = true ]]; then
-    kubectl \
+    timeout 600 kubectl \
       --kubeconfig=$(kind get kubeconfig-path --name="clusterapi") \
       delete cluster test1 || true
-    kubectl \
+     timeout 600 kubectl \
       --kubeconfig=$(kind get kubeconfig-path --name="clusterapi") \
       wait --for=delete cluster/test1 || true
     make kind-reset || true
   fi
   # clean up e2e.test symlink
   (cd "$(go env GOPATH)/src/k8s.io/kubernetes" && rm -f _output/bin/e2e.test) || true
+
+  # Force a cleanup of cluster api created resources using gcloud commands
+  gcloud compute forwarding-rules delete --project $GCP_PROJECT --global $CLUSTER_NAME-apiserver --quiet || true
+  gcloud compute target-tcp-proxies delete --project $GCP_PROJECT $CLUSTER_NAME-apiserver --quiet || true
+  gcloud compute backend-services delete --project $GCP_PROJECT --global $CLUSTER_NAME-apiserver --quiet || true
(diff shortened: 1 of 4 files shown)


hack/ci/e2e-conformance.sh
```diff
@@ -25,47 +25,143 @@ CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
...
+          exit 1
+        else
+          # we need pip to install ansible
+          curl -L https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+          python get-pip.py --user
+          rm -f get-pip.py
+
+          # install ansible needed by packer
+          version="2.8.5"
+          python -m pip install "ansible==${version}"
```
detiber: Should this be installed with --user as well?

hack/ci/e2e-conformance.sh
```diff
@@ -25,47 +25,143 @@ CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
...
+    --project "${GCP_PROJECT}" --zone "${zone}" "${node}" --command "${cmd}"
+}
+
 init_image() {
   image=$(gcloud compute images list --project $GCP_PROJECT \
     --no-standard-images --filter="family:capi-ubuntu-1804-k8s" --format="table[no-heading](name)")
   if [[ -z "$image" ]]; then
+      if ! command -v ansible &> /dev/null; then
+        if [[ $EUID -ne 0 ]]; then
+          echo "Please install ansible and try again."
```
detiber: Should we update the image running this in prow to include ansible as well?

hack/ci/e2e-conformance.sh
```diff
@@ -25,47 +25,143 @@ CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
...
+      if [[ $EUID -ne 0 ]]; then
+        (cd "$(go env GOPATH)/src/sigs.k8s.io/image-builder/images/capi" && \
+          GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS \
+          make build-gce-default)
+      else
+        # assume we are running in the CI environment as root
+        # Add a user for ansible to work properly
+        groupadd -r packer && useradd -m -s /bin/bash -r -g packer packer
+        # use the packer user to run the build
+        su - packer -c "bash -c 'cd /go/src/sigs.k8s.io/image-builder/images/capi && GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS make build-gce-default'"
```
detiber: I'm wondering if we need to add support to the kubeadm provider to override binaries used, then we wouldn't need to re-bake an image just to run the tests.

hack/ci/e2e-conformance.sh
```diff
@@ -25,47 +25,143 @@ CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
...
+  dump-logs
+
   # KIND_IS_UP is true once we: kind create
   if [[ "${KIND_IS_UP:-}" = true ]]; then
-    kubectl \
+    timeout 60 kubectl \
       --kubeconfig=$(kind get kubeconfig-path --name="clusterapi") \
       delete cluster test1 || true
-    kubectl \
+     timeout 60 kubectl \
```
detiber: 60 seconds might be a bit too short of a timeout here, maybe give it at least 10-15 minutes here?

hack/ci/e2e-conformance.sh
```diff
@@ -25,47 +25,143 @@ CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
...
+       | bash) || true
+  (gcloud compute firewall-rules list --project $GCP_PROJECT | grep $CLUSTER_NAME \
+       | awk '{print "gcloud compute firewall-rules delete --project '$GCP_PROJECT' --quiet " $1 "\n"}' \
+       | bash) || true
+
+  # cleanup the networks
+  gcloud compute routers nats delete "${CLUSTER_NAME}-mynat" --router-region="${GCP_REGION}" \
+    --router="${CLUSTER_NAME}-myrouter" --quiet || true
+  gcloud compute routers delete "${CLUSTER_NAME}-myrouter" --project="${GCP_PROJECT}" \
+    --region="${GCP_REGION}" --quiet || true
```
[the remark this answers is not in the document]
dims: Ack, i can do that as a follow up

hack/ci/e2e-conformance.sh
dims: we do this only if we are running as root, so i dropped the --user. see the line above where bailout with ` echo "Please install ansible and try again."`

hack/ci/e2e-conformance.sh
dims: quite possibly, unless we figure out a better way to build/share the images.

hack/ci/e2e-conformance.sh
dims: sounds like a good idea!

hack/ci/e2e-conformance.sh
dims: Ack, i'll bump it up to 10 and we can tweak it

dims:
/assign @vincepri @detiber 

Jason, Vince,
I'd like to merge this and iterate. Specifically i want to clone this into a periodic job and see how it fares over the weekend. (say run every 3 hours).

thanks,
Dims

hack/ci/e2e-conformance.sh
detiber: Actually, if we add support for overriding binaries in CABPK, then I think we can avoid ansible all together.

dims: Added an issue https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/258

hack/ci/e2e-conformance.sh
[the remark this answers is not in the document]
dims: Created an issue and assigned it to myself https://github.com/kubernetes/test-infra/issues/14498

scripts/ci-e2e.sh
```diff
@@ -22,38 +22,35 @@
...
+
+if [ ! "${checkout_account_status}" = "0" ]; then
+  echo "error getting account from boskos" 1>&2
+  exit "${checkout_account_status}"
 fi
 
 (cd "${REPO_ROOT}" && hack/ci/e2e-conformance.sh)
+
+echo "cleaning up - checking in boskos account $BOSKOS_RESOURCE_NAME on host $BOSKOS_HOST"
+hack/checkin_account.py
```
detiber: Putting this back into a signal handler would probably be good, otherwise this will probably not run if the conformance test fails.

dims: Done! thanks @detiber

PR #180, opened 2019-09-27
:book: Add prerequisites file

vincepri:

Fixes #

/assign @dims

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -1,0 +1,54 @@
+# Prerequisites
+
+## Requirements
+
+- Linux or MacOS (Windows isn't supported at the moment).
+- A [Google Cloud](https://console.cloud.google.com) account.
+- [Packer](https://www.packer.io/intro/getting-started/install.html) and [Ansible](https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html) to build images
+- Make to use `Makefile` targets
+
+### Setup a Network and Cloud NAT
+
+Google Cloud accounts come with a `default` network which can be found under
+[VPC Networks](https://console.cloud.google.com/networking/networks).
+If you prefer to create a new Network, follow [these instructions](https://cloud.google.com/vpc/docs/using-vpc#create-auto-network).
+
+#### Cloud NAT
+This infrastructure provider sets up Kubernetes clusters using a
+[Global Load Balancer](https://cloud.google.com/load-balancing/) with a public ip address.
+
+Kubernetes nodes, to communicate with the control plane, pull container images from registried (e.g. gcr.io or dockerhub) need to have NAT access or a public ip.
+By default, the provider creates Machines without a public IP.
+
+To make sure your cluster can communicate with the outside world, and the load balancer, you can create a [Cloud NAT](https://cloud.google.com/nat/docs/overview) in the region you'd like your Kubernetes cluster to live in by following [these instructions](https://cloud.google.com/nat/docs/using-nat#create_nat).
+
+### Create a Service Account
+
+To create and manager clusters, this infrastructure providers uses a service account to authenticate with GCP's APIs.
+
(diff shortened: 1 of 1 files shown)


docs/prerequisites.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Create a Service Account
+
+To create and manager clusters, this infrastructure providers uses a service account to authenticate with GCP's APIs.
+
+From your cloud console, follow [these instructions](https://cloud.google.com/iam/docs/creating-managing-service-accounts#creating) to create a new service account with `Editor` permissions. Afterwards, generate a JSON Key and store it somewhere safe.
+
+### Building images
+
+```bash
```
dims: - we need to add a caveat here to NOT run the make as `root`
- add a note that this uploads the image as well.

Also they can check if they already have an image using:
`gcloud compute images list --project $GCP_PROJECT --no-standard-images --filter="family:capi-ubuntu-1804-k8s"`

docs/prerequisites.md
```diff
@@ -0,0 +1,48 @@
...
+
+Google Cloud accounts come with a `default` network which can be found under
+[VPC Networks](https://console.cloud.google.com/networking/networks).
+If you prefer to create a new Network, follow [these instructions](https://cloud.google.com/vpc/docs/using-vpc#create-auto-network).
+
+#### Cloud NAT
+This infrastructure provider sets up Kubernetes clusters using a
+[Global Load Balancer](https://cloud.google.com/load-balancing/) with a public ip address.
+
+Kubernetes nodes, to communicate with the control plane, need to have NAT access or a public ip. By default, the provider creates Machines without a public IP.
```
dims: Also needed to fetch container images from gcr.io and/or dockerhub as well.

docs/prerequisites.md
```diff
@@ -0,0 +1,48 @@
+# Prerequisites
+
+## Requirements
+
+- Linux or MacOS (Windows isn't supported at the moment).
+- A [Google Cloud](https://console.cloud.google.com) account.
+- [Packer](https://packer.io) to build images
```
dims: - Do you want to add `kustomize` (needed for the manifests?)
- also `ansible` is needed at least for now? (used from packer)

docs/prerequisites.md
```diff
@@ -0,0 +1,48 @@
...
+- Make to use `Makefile` targets
```
dims: Make should be lower case `M`?

docs/prerequisites.md
```diff
@@ -0,0 +1,48 @@
...
+# Clone the image builder repository if you haven't already.
+git clone https://sigs.k8s.io/image-builder.git image-builder
+
+# Change directory to images/capi within the image builder repository
+cd image-builder/images/capi
+
+# Run the Make target to generate GCE images.
+make build-gce-default
+```
+
```
dims: Do you want to show how to build the example manifests?

vincepri: The examples folder shouldn't be needed here. I plan to embed/link this document from the Quick Start after we release a first alpha version

docs/prerequisites.md
vincepri: Yeah, it's capitalized because it's the start of a sentence 😬

docs/prerequisites.md
vincepri: +1 to ansible, kustomize should be optional at this point. Unless folks want to build the examples

docs/prerequisites.md
vincepri: Is the filter a prefix here?

vincepri: Because the actual family have the k8s version as suffix https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/gce/packer.json#L43

dims:
/lgtm

we can always iterate

PR #188, opened 2019-09-30
Build/use gcp cluster-api-provider image explicitly

dims:
So far we have been using the `gcr.io/k8s-staging-cluster-api-gcp/cluster-api-gcp-controller:latest` container image by default. We should build and push an image from the source code into the project specific repo (that we get from boskos) and use that to start the gcp controller in the kind cluster. We should do the same irrespective of whether this is a periodic job or a pre-submit job to ensure that we are testing the exact changes by building the image from the code.

Change-Id: I2d2cc8dd2c6e3ad8808fbf4bc9130e115f00c050

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,8 @@ GOLANGCI_LINT := $(TOOLS_BIN_DIR)/golangci-lint
 MOCKGEN := $(TOOLS_BIN_DIR)/mockgen
 
 # Define Docker related variables. Releases should modify and double check these vars.
-REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+GCP_PROJECT ?= $(shell gcloud config get-value project)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -46,7 +46,8 @@ GOLANGCI_LINT := $(TOOLS_BIN_DIR)/golangci-lint
 MOCKGEN := $(TOOLS_BIN_DIR)/mockgen
 
 # Define Docker related variables. Releases should modify and double check these vars.
-REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+GCP_PROJECT ?= $(shell gcloud config get-value project)
+REGISTRY ?= gcr.io/$(GCP_PROJECT)
```
detiber: Instead of building/pushing to a remote registry here, why not take advantage of `kind load docker-image` here instead?

PR #192, opened 2019-10-01
Create a new network for us to use, still allow default to work

dims:
Change-Id: Ia1438ba496707680309ff645962b4295439dc977

Fixes #


detiber:
@dims are there perhaps reasons why some testing might prefer to be run against legacy networks? Should we instead test for legacy network and attempt to checkout another GCE account instead?

dims:
@detiber have to ask folks in sig-testing on why we have a mix of these. Since the gcp-janitor seems to remove all networks and firewalls, i was trying to see if these are remnants of a bad gcp janitor run as well. (and see if we could clean these and reuse them)

vincepri:
Instead of modifying the default network, what if we create a brand new one and delete it at the end?

dims:
@vincepri good point!

PR #206, opened 2019-10-04
Add option to cleanup

dims:
Change-Id: I1c536c080d0d1e5df3ec370e5c2ca43fb4461181

Fixes #

### DIFF
--- a/hack/ci/e2e-conformance.sh
+++ b/hack/ci/e2e-conformance.sh
@@ -26,6 +26,7 @@ NETWORK_NAME=${NETWORK_NAME:-"${CLUSTER_NAME}-mynetwork"}
 
 TIMESTAMP=$(date +"%Y-%m-%dT%H:%M:%SZ")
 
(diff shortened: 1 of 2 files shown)


hack/ci/e2e-conformance.sh
```diff
@@ -361,14 +365,18 @@ EOF
     return 2
   fi
 
+  if [[ ${1:-} == "--clean" ]]; then
```
detiber: Should this use the ENV var to align with the doc PR? 

Otherwise, might need to do some shifting of args to support both this and the --verbose flag.

dims: you mean support `hack/ci/e2e-conformance.sh --clean --verbose` ? can we do that later, merge and iterate

detiber: Sure, but should the docs be updated for the flag rather than the env var?

PR #215, opened 2019-10-14
✨ Allow the ability to configure frontend and backend port for LB

rsmitty:

Interested to see how everyone feels about this.

This PR will add the ability to pass in ports you wish to use for the front/backends of the LB that gets created by the GCP provider. This was useful for me because I wanted to change the port k8s is listening on from 6443 to 443 all the way through. After some discussion on the cluster-api slack, I moved the frontend port to  respect what might be set in `Cluster.Spec.ClusterNetwork.APIServerPort`. This matches what is supported in the AWS provider.

### DIFF
--- a/api/v1alpha2/types.go
+++ b/api/v1alpha2/types.go
@@ -96,6 +96,10 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
+
+	// Allow for configuration of load balancer backend (useful for changing apiserver port)
+	// +optional
+	LoadBalancerBackendPort *int32 `json:"loadBalancerBackendPort,omitempty"`
 }
 
 // APIEndpoint represents a reachable Kubernetes API endpoint.
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -130,6 +130,23 @@ func (s *ClusterScope) Region() string {
 	return s.GCPCluster.Spec.Region
 }
 
+// LoadBalancerFrontendPort returns the loadbalancer frontend if specified
+// in the cluster resource's network configuration.
+func (s *ClusterScope) LoadBalancerFrontendPort() int64 {
+	if s.Cluster.Spec.ClusterNetwork.APIServerPort != nil {
+		return int64(*s.Cluster.Spec.ClusterNetwork.APIServerPort)
+	}
+	return 443
+}
+
+// LoadBalancerBackendPort returns the loadbalancer backend if specified.
+func (s *ClusterScope) LoadBalancerBackendPort() int64 {
+	if s.GCPCluster.Spec.Network.LoadBalancerBackendPort != nil {
+		return int64(*s.GCPCluster.Spec.Network.LoadBalancerBackendPort)
+	}
+	return 6443
+}
+
 // ControlPlaneConfigMapName returns the name of the ConfigMap used to
 // coordinate the bootstrapping of control plane nodes.
 func (s *ClusterScope) ControlPlaneConfigMapName() string {
--- a/cloud/services/compute/firewalls.go
+++ b/cloud/services/compute/firewalls.go
@@ -81,7 +81,7 @@ func (s *Service) getFirewallSpecs() []*compute.Firewall {
 				{
 					IPProtocol: "TCP",
 					Ports: []string{
-						strconv.Itoa(APIServerLoadBalancerBackendPort),
+						strconv.FormatInt(s.scope.LoadBalancerBackendPort(), 10),
 					},
 				},
 			},
--- a/cloud/services/compute/instancegroup.go
+++ b/cloud/services/compute/instancegroup.go
@@ -77,7 +77,7 @@ func (s *Service) GetOrCreateInstanceGroup(zone, name string) (*compute.Instance
 			NamedPorts: []*compute.NamedPort{
 				{
 					Name: "apiserver",
-					Port: 6443,
+					Port: s.scope.LoadBalancerBackendPort(),
 				},
 			},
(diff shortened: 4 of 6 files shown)


cloud/scope/cluster.go
```diff
@@ -130,6 +130,23 @@ func (s *ClusterScope) Region() string {
 	return s.GCPCluster.Spec.Region
 }
 
+// LoadBalancerFrontendPort returns the loadbalancer frontend if specified
+// in the cluster resource's network configuration.
+func (s *ClusterScope) LoadBalancerFrontendPort() int64 {
+	if s.Cluster.Spec.ClusterNetwork.APIServerPort != nil {
+		return int64(*s.Cluster.Spec.ClusterNetwork.APIServerPort)
+	}
+	return 0
```
vincepri: Add the default here?

cloud/scope/cluster.go
```diff
@@ -130,6 +130,23 @@ func (s *ClusterScope) Region() string {
...
+}
+
+// LoadBalancerBackendPort returns the loadbalancer backend if specified.
+func (s *ClusterScope) LoadBalancerBackendPort() int64 {
+	if s.GCPCluster.Spec.Network.LoadBalancerBackendPort != nil {
+		return *s.GCPCluster.Spec.Network.LoadBalancerBackendPort
+	}
+	return 0
```
vincepri: Add the default here?

cloud/services/compute/firewalls.go
```diff
@@ -73,6 +73,11 @@ func (s *Service) DeleteFirewalls() error {
 }
 
 func (s *Service) getFirewallSpecs() []*compute.Firewall {
+	backendPort := int64(APIServerLoadBalancerBackendPort)
+	if s.scope.LoadBalancerBackendPort() != 0 {
+		backendPort = s.scope.LoadBalancerBackendPort()
+	}
```
vincepri: It seems like the default logic is duplicated here as well, let's make it happen in one place

cloud/services/compute/instancegroup.go
```diff
@@ -71,13 +71,18 @@ func (s *Service) DeleteInstanceGroups() error {
...
+		}
+
 		spec := &compute.InstanceGroup{
 			Name:    name,
 			Network: s.scope.NetworkSelfLink(),
 			NamedPorts: []*compute.NamedPort{
 				{
 					Name: "apiserver",
-					Port: 6443,
+					Port: backendPort,
```
vincepri: It'd be great if we can call `s.scope.LoadBalancerBackendPort()` here directly

cloud/services/compute/loadbalancers.go
```diff
@@ -304,12 +309,16 @@ func (s *Service) getAPIServerIPAddressSpec() *compute.Address {
...
+		frontendPortRange = fmt.Sprintf("%d-%d", s.scope.LoadBalancerFrontendPort(), s.scope.LoadBalancerFrontendPort())
+	}
 	return &compute.ForwardingRule{
 		Name:                fmt.Sprintf("%s-%s", s.scope.Name(), infrav1.APIServerRoleTagValue),
 		IPAddress:           *s.scope.Network().APIServerAddress,
 		IPProtocol:          APIServerLoadBalancerProtocol,
 		LoadBalancingScheme: APIServerLoadBalancerScheme,
-		PortRange:           APIServerLoadBalancerFrontendPortRange,
+		PortRange:           frontendPortRange,
 		Target:              *s.scope.Network().APIServerTargetProxy,
```
vincepri: Same comments from above

rsmitty:
All great feedback, thx! I've moved to just returning the defaults directly in the cluster scope functions and amended my commit. Feels cleaner for sure

api/v1alpha2/types.go
```diff
@@ -96,6 +96,10 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
+
+	// Allow for configuration of load balancer backend (useful for changing apiserver port)
+	// +optional
+	LoadBalancerBackendPort *int64 `json:"lbBackendPort,omitempty"`
```
vincepri:
```suggestion
	LoadBalancerBackendPort *int64 `json:"loadBalancerBackendPort,omitempty"`
```

rsmitty: Resolved in an amended commit

api/v1alpha2/types.go
```diff
@@ -96,6 +96,10 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
+
+	// Allow for configuration of load balancer backend (useful for changing apiserver port)
+	// +optional
+	LoadBalancerBackendPort *int64 `json:"loadBalancerBackendPort,omitempty"`
```
detiber: Any specific reason for `int64` here instead of `int32`?

rsmitty: Yeah, all of the GCP compute services seem to require int64 for ports. So I went ahead and just used int64 directly here instead of converting it later on. We convert the Cluster.Spec.ClusterNetwork.APIServerPort field in the LoadBalancerFrontendPort function as well. I could change it to int32 though, I suppose if it makes more sense to have it match Cluster.Spec.ClusterNetwork.APIServerPort

vincepri:
@rsmitty Would you be able to change the type to int32 for consistency?

I'd like to do another soft release today, and include this change :)

rsmitty:
@vincepri Amended so that we accept int32 pointers as input and we convert and output to int64 with the LoadBalancerFrontendPort and LoadBalancerBackendPort function calls.

PR #228, opened 2019-10-17
🐛 whoops - local change sneaked in

dims:

Fixes #

### DIFF
--- a/config/default/manager_image_patch.yaml
+++ b/config/default/manager_image_patch.yaml
@@ -9,5 +9,6 @@ spec:
       containers:
       # Change the value of image field below to your controller image URL
       - image: gcr.io/vincepri-dev/cluster-api-gcp-controller-amd64:dev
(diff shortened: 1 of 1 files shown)


config/default/manager_image_patch.yaml
```diff
@@ -8,6 +8,7 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
+      - image: gcr.io/vincepri-dev/cluster-api-gcp-controller-amd64:dev
```
vincepri: :master?

detiber:
```suggestion
```

vincepri: This shouldn't be removed entirely, we should use the staging bucket as it was before

detiber: The image is listed twice

PR #260, opened 2019-12-03
🏃 add tests for gcp_machine_controller 

nader-ziada:
- improve test coverage
- create a mock gcp compute service to be able to mock out the calls to gcp in the test
- change the controller to allow for passing of the mock service


controllers/gcpmachine_controller_test.go
```diff
@@ -17,12 +17,27 @@ limitations under the License.
 package controllers
 
 import (
+	"bytes"
+	"context"
+	"flag"
+
+	capierrors "sigs.k8s.io/cluster-api/errors"
```
detiber: nit, this should be grouped with the imports below

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +64,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+			machineScope.SetErrorReason(capierrors.UpdateMachineError)
+			machineScope.SetErrorMessage(errors.New("GCE error"))
+			Expect(k8sClient.Update(ctx, machineScope.GCPMachine)).Should(Succeed())
+
+			buf := new(bytes.Buffer)
+			klog.SetOutput(buf)
+
+			_, err := reconciler.reconcile(ctx, machineScope, clusterScope)
+			Expect(err).To(BeNil())
+			Expect(buf.String()).To(ContainSubstring("Error state detected, skipping reconciliation"))
```
detiber: Checking for a specific log message seems a bit brittle. Do we need a better way to bubble this information up?

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +64,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+		It("should return if infrastructure is not ready", func() {
+			clusterStatus := clusterv1.ClusterStatus{InfrastructureReady: false}
+			machineScope.Cluster.Status = clusterStatus
+
+			buf := new(bytes.Buffer)
+			klog.SetOutput(buf)
+
+			_, err := reconciler.reconcile(ctx, machineScope, clusterScope)
+			Expect(err).To(BeNil())
+			Expect(buf.String()).To(ContainSubstring("Cluster infrastructure is not ready yet"))
```
detiber: Checking for a specific log message seems a bit brittle. Do we need a better way to bubble this information up?

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +64,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+					Data: nil,
+				}}
+			machineScope.Machine.Spec = spec
+
+			buf := new(bytes.Buffer)
+			klog.SetOutput(buf)
+
+			_, err := reconciler.reconcile(ctx, machineScope, clusterScope)
+			Expect(err).To(BeNil())
+			Expect(buf.String()).To(ContainSubstring("Bootstrap data is not yet available"))
```
detiber: Checking for a specific log message seems a bit brittle. Do we need a better way to bubble this information up?

controllers/gcpmachine_controller_test.go
nader-ziada: will change

controllers/gcpmachine_controller_test.go
nader-ziada: The controller currently sets the machineScope.info with this message and intentionally sets the error to nil, would you prefer returning a specific error?

detiber: I believe there was an issue with returning an error and having a way to tell controller-runtime not to requeue the request. According to https://github.com/kubernetes-sigs/controller-runtime/issues/617 it looks like this is still an outstanding issue.

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+		})
+
+		AfterEach(func() {
+			Expect(k8sClient.Delete(ctx, machineScope.GCPMachine)).Should(Succeed())
+		})
+
+		It("should return early if gcp machine is in error state", func() {
+			machineScope.SetErrorReason(capierrors.UpdateMachineError)
+			machineScope.SetErrorMessage(errors.New("GCE error"))
+			Expect(k8sClient.Update(ctx, machineScope.GCPMachine)).Should(Succeed())
```
wfernandes: Do we need this Update here?

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+						Name:       gcpMachine.Name,
+					},
+					Bootstrap: clusterv1.Bootstrap{
+						Data: &bootstrapData,
+					},
+					Version: &version,
+				},
+			}
+
+			Expect(k8sClient.Create(ctx, gcpMachine)).Should(Succeed())
```
wfernandes: Is there a need to use `k8sClient`? Can we just set up a `fakeClient = fake.NewFakeClient()` and use that everywhere?

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+					Namespace: "default",
+				},
+				Spec: infrav1.GCPClusterSpec{
+					Project: "gcp-project",
+				},
+			}
+
+			// Create the cluster scope
+			clusterScope, err = scope.NewClusterScope(scope.ClusterScopeParams{
+				Client:     fake.NewFakeClient(),
```
wfernandes: Use the above `fakeClient`?

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+				Logger:     log.Log,
+				Cluster:    cluster,
+				GCPCluster: gcpCluster,
+			})
+			Expect(err).To(BeNil())
+
+			// Create the machine scope
+			machineScope, err = scope.NewMachineScope(scope.MachineScopeParams{
+				Logger:     log.Log,
+				Client:     fake.NewFakeClient(),
```
wfernandes: Use the above `fakeClient`?

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+				Machine:    machine,
+				GCPCluster: gcpCluster,
+				GCPMachine: gcpMachine,
+			})
+			Expect(err).To(BeNil())
+
+		})
+
+		AfterEach(func() {
+			Expect(k8sClient.Delete(ctx, machineScope.GCPMachine)).Should(Succeed())
```
wfernandes: Use the above `fakeClient`?

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+			machineScope, err = scope.NewMachineScope(scope.MachineScopeParams{
+				Logger:     log.Log,
+				Client:     fake.NewFakeClient(),
+				Cluster:    cluster,
+				Machine:    machine,
+				GCPCluster: gcpCluster,
+				GCPMachine: gcpMachine,
+			})
+			Expect(err).To(BeNil())
+
```
wfernandes: nit: extra unnecessary whitespace

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+			clusterStatus := clusterv1.ClusterStatus{InfrastructureReady: false}
+			machineScope.Cluster.Status = clusterStatus
+
+			buf := new(bytes.Buffer)
+			klog.SetOutput(buf)
+
+			_, err := reconciler.reconcile(ctx, machineScope, clusterScope)
+			Expect(err).To(BeNil())
+			Expect(buf.String()).To(ContainSubstring("Cluster infrastructure is not ready yet"))
+
```
wfernandes: nit: extra unnecessary whitespace

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+				}}
+			machineScope.Machine.Spec = spec
+
+			buf := new(bytes.Buffer)
+			klog.SetOutput(buf)
+
+			_, err := reconciler.reconcile(ctx, machineScope, clusterScope)
+			Expect(err).To(BeNil())
+			Expect(buf.String()).To(ContainSubstring("Bootstrap data is not yet available"))
+
```
wfernandes: nit: extra unnecessary whitespace

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+			_, err := reconciler.reconcile(ctx, machineScope, clusterScope)
+			Expect(err).To(BeNil())
+
+			By("Checking providerId")
+			Expect(machineScope.GetProviderID()).To(Equal("gce://gcp-project/us-west4-a/gce-instance"))
+			By("Checking instanceStatus")
+			Expect(*machineScope.GetInstanceStatus()).To(Equal(infrav1.InstanceStatus("RUNNING")))
+			By("Checking a finalizer is added")
+			Expect(machineScope.GCPMachine.Finalizers).To(ContainElement(infrav1.MachineFinalizer))
+
```
wfernandes: nit: extra unnecessary whitespace

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +63,171 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+				Spec: clusterv1.MachineSpec{
+					InfrastructureRef: corev1.ObjectReference{
+						APIVersion: "infrastructure.cluster.x-k8s.io/v1alpha3",
+						Kind:       "GCPMachine",
+						Name:       gcpMachine.Name,
+					},
+					Bootstrap: clusterv1.Bootstrap{
+						Data: &bootstrapData,
+					},
+					Version: &version,
```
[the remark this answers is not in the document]
nader-ziada: This is a required field that will give a validation error if not present

controllers/gcpmachine_controller_test.go
nader-ziada: k8sclient is setup in the `suite_test` and is the recommended way in kubebuilder docs. I was already there as part of the existing test

controllers/gcpmachine_controller_test.go
nader-ziada: yes, good catch, no need for another fake client, will use k8sClient

controllers/gcpmachine_controller_test.go
nader-ziada: will switch to k8sClient

controllers/gcpmachine_controller_test.go
nader-ziada: ok, will remove the extra lines

controllers/gcpmachine_controller_test.go
nader-ziada: k8sClient is the client setup in suite_test, removed the other fake client

controllers/gcpmachine_controller_test.go
nader-ziada: I guess it can be removed

controllers/gcpmachine_controller_test.go
nader-ziada: there is a series of checks that happen before this call is made, so I don't think just checking for `InstanceIfExists` not being called would be enough replacement that the right check failed

controllers/gcpmachine_controller_test.go
nader-ziada: there is a series of checks that happen before this call is made, so I don't think just checking for InstanceIfExists not being called would be enough replacement that the right check failed

controllers/gcpmachine_controller_test.go
nader-ziada: ok

controllers/gcpmachine_controller_test.go
nader-ziada: there is a series of checks that happen before this call is made, so I don't think just checking for InstanceIfExists not being called would be enough replacement that the right check failed

controllers/gcpmachine_controller_test.go
nader-ziada: ok

controllers/gcpmachine_controller_test.go
nader-ziada: ok

nader-ziada:
@detiber @wfernandes since you looked at this code, can one of you give this PR an `ok-to-test` if you think it makes sense. Thanks

nader-ziada:
I've had a chat with @wfernandes and explained why checking that a func has not been called yet in the test might not guarantee that the required validation happened since all these validation happen in sequence in the same `reconcile` method. 

is there anything else I should be changing here? 

Thanks

controllers/gcpmachine_controller_test.go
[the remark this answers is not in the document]
wfernandes: Hmm...what validation is this? When I take it out and run `make test` everything passes. And currently the property is optional so I'm confused as to why it is required in this context. (https://github.com/kubernetes-sigs/cluster-api/blob/900a2ff1f6688773ff48dc56c459484f22c21c6b/api/v1alpha3/machine_types.go#L60-L61).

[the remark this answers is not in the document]
nader-ziada: can't find this check anymore, will remove it

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +62,163 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+				GCPMachine: gcpMachine,
+			})
+			Expect(err).To(BeNil())
+		})
+
+		AfterEach(func() {
+			Expect(k8sClient.Delete(ctx, machineScope.GCPMachine)).Should(Succeed())
+		})
+
+		It("should return early if gcp machine is in error state", func() {
```
wfernandes: Can we change the test title to better reflect what is being tested? Currently it doesn't seem like we are testing early return. Is there a way to assert that reconciliation hasn't happened? Maybe verifying if ProviderID isn't set. since that is a mandatory step in reconciliation. WDYT?

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +62,163 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+
+			buf := new(bytes.Buffer)
+			klog.SetOutput(buf)
+
+			_, err := reconciler.reconcile(ctx, machineScope, clusterScope)
+			Expect(err).To(BeNil())
+			Expect(buf.String()).To(ContainSubstring("Error state detected, skipping reconciliation"))
+		})
+
+		It("should return if infrastructure is not ready", func() {
```
wfernandes: Same comment as above.

controllers/gcpmachine_controller_test.go
```diff
@@ -48,4 +62,163 @@ var _ = Describe("GCPMachineReconciler", func() {
...
+
+			buf := new(bytes.Buffer)
+			klog.SetOutput(buf)
+
+			_, err := reconciler.reconcile(ctx, machineScope, clusterScope)
+			Expect(err).To(BeNil())
+			Expect(buf.String()).To(ContainSubstring("Cluster infrastructure is not ready yet"))
+		})
+
+		It("should return if bootstrap data is not available", func() {
```
wfernandes: Same comment as above.

controllers/gcpmachine_controller_test.go
nader-ziada: I'm not sure I understand your point? returning early is just another way of saying reconciliation didn't happen and since there is an error message it also means reconciliation didn't happen

nader-ziada:
@vincepri @wfernandes any comments of this PR? I'm not of any outstanding changes I should still make. Thanks

nader-ziada:
I don't think the failure in the conformance tests are related to the changes in this PR, will investigate further anyways

PR #264, opened 2019-12-06
 ⚠️ Rename cert-manager API group from certmanager.k8s.io to cert-manager.io

nader-ziada:
 - Rename cert-manager API group from certmanager.k8s.io to cert-manager.io to be compatible with cert-manager


nader-ziada:
@vincepri Are there any changes required in this PR?

PR #269, opened 2019-12-16
:running: Remove clusterctl dependency from create-cluster

vincepri:


joonas:
Overall it looks good, but I wonder if the `create-cluster` target has become large enough to warrant being turned in to a separate bash script?

vincepri:
Probably, but I'd leave it for another time. It's only used for development or e2e tests right now

vincepri:
Conformance is failing because of the data secret name, which will be fixed in another PR

PR #272, opened 2019-12-17
🏃 Let GCB build release images

ncdc:
test-infra has been updated to use GCB to build release version images,
so we no longer need to do it here in the repo.

Requires https://github.com/kubernetes/test-infra/pull/15624

Fixes #


ncdc:
Depends on the fix to #273, whatever that is

PR #284, opened 2020-02-11
🐛 fix-failing-test

fabriziopandini:
Fixes the problem introduced by kubernetes/kubernetes#87585 (no more .deb generated on CI)

Fixes kubernetes/kubernetes#87960

/assign @ncdc @detiber

### DIFF
--- a/examples/machinedeployment/kustomizeversions.yaml
+++ b/examples/machinedeployment/kustomizeversions.yaml
@@ -26,7 +26,7 @@ spec:
           fi
           gsutil version
 
-          # This test installs debian packages that are a result of the CI and release builds.
+          # This test installs binaries that are a result of the CI and release builds.
           # It runs '... --version' commands to verify that the binaries are correctly installed
           # and finally uninstalls the packages.
           # For the release packages it tests all versions in the support skew.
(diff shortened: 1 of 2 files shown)


examples/machinedeployment/kustomizeversions.yaml
```diff
@@ -40,18 +40,20 @@ spec:
...
-            echo "* downloading package: $CI_URL/$CI_PACKAGE.$PACKAGE_EXT"
-            gsutil cp "$CI_URL/$CI_PACKAGE.$PACKAGE_EXT" "$CI_DIR/$CI_PACKAGE.$PACKAGE_EXT"
-            ${SUDO} dpkg -i "$CI_DIR/$CI_PACKAGE.$PACKAGE_EXT" || echo "* ignoring expected 'dpkg -i' result"
+            echo "* downloading binary: $CI_URL/$CI_PACKAGE"
+            gsutil cp "$CI_URL/$CI_PACKAGE" "$CI_DIR/$CI_PACKAGE"
+            chmod +x "$CI_DIR/$CI_PACKAGE"
+            mv "$CI_DIR/$CI_PACKAGE" "/usr/bin/$CI_PACKAGE"
             done
 
+            systemctl restart kubelet
```
neolit123: the kubelet has some dependencies:
https://github.com/kubernetes/release/blob/master/cmd/kubepkg/templates/latest/deb/kubelet/debian/control#L13

IMO not having these can cause issues:
iptables, kubernetes-cni, socat, mount.

and it's better to `apt-get install` fixed versions of them.

detiber: So, currently the base image we are using has v1.16 packages already installed for `kubelet`, `kubeadm`, etc. Though, we should likely update that to v1.17.

The only time this would be a problem is if there is a new dependency or a new minimum requirement is added for the latest development release.

PR #285, opened 2020-02-12
✨health: Add liveness and readiness probes

cpanato:
Adding liveness and readiness probes similar we added here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1487 and https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/374

Fixes #

```release-note
- liveness and readiness probes added
```

/cc @vincepri

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/cluster-api/util/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -59,6 +60,7 @@ func main() {
 		gcpClusterConcurrency int
 		gcpMachineConcurrency int
 		syncPeriod            time.Duration
+		healthAddr            string
```
jdef: I think you need to pass this into the controller-manager Options via HealthProbeBindAddress

cpanato: good catch, forgot that, thanks!

PR #291, opened 2020-04-03
:running: [ci-conformance] Fix kustomize template to remove ami reference

detiber:
Conformance tests against k8s master are currently broken, this fixes errors related to the kustomize templating.


detiber:
/assign @ncdc 
Fixes the current cause of failure for the v1alpha3/k8s master periodic job

PR #292, opened 2020-04-06
:running: [ci-conformance] Fix kustomize template to better handle envsubst

detiber:

Updates the script used during conformance testing against upstream k8s to remove bash variables other than the ones we replace with the call to envsubst, since envsubst is replacing them with empty strings currently.

### DIFF
--- a/controllers/gcpcluster_controller.go
+++ b/controllers/gcpcluster_controller.go
@@ -50,7 +50,7 @@ func (r *GCPClusterReconciler) SetupWithManager(mgr ctrl.Manager, options contro
 		WithOptions(options).
 		For(&infrav1.GCPCluster{}).
 		Watches(
-			&source.Kind{Type: &infrav1.GCPCluster{}},
+			&source.Kind{Type: &infrav1.GCPMachine{}},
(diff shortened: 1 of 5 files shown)


detiber:
/assign @gab-satchi 
@gab-satchi ptal at this change as well, it is very similar to the changes made to CAPA, with the one exception that we need to build an image as part of CI anyway, so rather than overriding the binaries at cloud-init bootstrapping time, we are now building the k8s ci artifacts into the image we are building instead.

controllers/gcpcluster_controller.go
```diff
@@ -50,7 +50,7 @@ func (r *GCPClusterReconciler) SetupWithManager(mgr ctrl.Manager, options contro
 		WithOptions(options).
 		For(&infrav1.GCPCluster{}).
 		Watches(
-			&source.Kind{Type: &infrav1.GCPCluster{}},
+			&source.Kind{Type: &infrav1.GCPMachine{}},
```
gab-satchi, line 53: I guess this handler was just failing until now :p

detiber: Correct, I only noticed when I switched to testing and troubleshooting locally.

PR #295, opened 2020-04-23
:running: [ci-conformance] Fix tests when not using CI artifacts

detiber:
Fixes conformance test jobs where CI artifacts are not being tested, such as: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-cluster-api-provider-gcp-make-conformance-v1alpha3/1253298101955858434

### DIFF
--- a/hack/ci/e2e-conformance.sh
+++ b/hack/ci/e2e-conformance.sh
@@ -222,6 +222,8 @@ init_image() {
       rm packer.zip && \
       ln -s "$PWD"/packer /usr/local/bin/packer
   fi
+
+  if [[ -n ${CI_VERSION:-} ]]; then
     cat << EOF > "$(go env GOPATH)/src/sigs.k8s.io/image-builder/images/capi/override.json"
 {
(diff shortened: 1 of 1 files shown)


detiber:
conformance job failure is the same failure we are seeing w/ the periodic job, still need to debug that issue but it's separate from the issue that this PR is resolving.

hack/ci/e2e-conformance.sh
```diff
@@ -222,6 +222,8 @@ init_image() {
       rm packer.zip && \
       ln -s "$PWD"/packer /usr/local/bin/packer
   fi
+
+  if [[ -n ${CI_VERSION:-} ]]; then
```
gab-satchi, line 226: do we need the `:-`? My bash-fu isn't great. I thought it's typically used for defaulting

detiber: Yes, without it we would error if the variable is unset (because of the set command on line 19 that enables `-o nounset`)

PR #299, opened 2020-05-27
Register only with first component of local_hostname

justinsb:
This avoids the 63 character limit on the kubernetes.io/hostname label.

https://github.com/kubernetes/kubernetes/issues/90686

### DIFF
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -39,7 +39,7 @@ spec:
   kubeadmConfigSpec:
     initConfiguration:
       nodeRegistration:
-        name: '{{ ds.meta_data.local_hostname }}'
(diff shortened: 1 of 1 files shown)


justinsb:
Going to let this run in e2e and see if it does what I think it does...

justinsb:
Thanks for the nudge @hasheddan  - this does do what I expected (nodes just have the first component), and it does pass e2e, so removing WIP!

e2e took 2 hours though - is that a regression?

hasheddan:
@justinsb Nice! I think ~2 hours looks like the norm here: https://testgrid.k8s.io/sig-release-master-informing#capg-conformance-v1alpha3-k8s-master&graph-metrics=test-duration-minutes

detiber:
2 hours sounds about right.

Issue #267, opened 2019-12-10
auto create cloudnat / bastian nodes if needed


jzhoucliqr:
/kind feature

Do you guys think it makes sense to have option to auto create cloudnat / bastian nodes if user choose not to have external IP for the nodes?

vincepri:
We've left it to users to configure Cloud NAT, given that it's fairly expensive today. It should be better from January 1st, when GCP said it'll cost 0$ if unused (IIRC, need to double check)

PR #307, opened 2020-07-20
✨ Auto create Cloud NAT if network is created by CAPG

snehala27:
In case where the network name is provided in GCPCluster configuration(for customisation) and the the network does not exist, in that case network and subnets are auto created by CAPG. In such a case router and Cloud NAT should also be auto created for the entire functionality to work. 
This change creates router and CloudNAT only if network is created by CAPG.

Fixes the issue partially #267

### DIFF
--- a/cloud/services/compute/network.go
+++ b/cloud/services/compute/network.go
@@ -87,6 +87,22 @@ func (s *Service) DeleteNetwork() error {
 		return nil
 	}
 
+	// Delete Router.
+	router, err := s.routers.Get(s.scope.Project(), s.scope.Region(), getRouterName(s.scope.NetworkName())).Do()
+	if err == nil {
+		op, err := s.routers.Delete(s.scope.Project(), s.scope.Region(), router.Name).Do()
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete router")
+		}
+		if err := wait.ForComputeOperation(s.scope.Compute, s.scope.Project(), op); err != nil {
+			return errors.Wrapf(err, "failed to wait for delete router")
+		}
+	} else {
+		if !gcperrors.IsNotFound(err) {
+			return errors.Wrapf(err, "failed to get router to delete")
+		}
+	}
+
 	// Delete Network.
 	op, err := s.networks.Delete(s.scope.Project(), network.Name).Do()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cloud/services/compute/network.go
```diff
@@ -31,7 +32,9 @@ func (s *Service) ReconcileNetwork() error {
 	// Create Network
 	spec := s.getNetworkSpec()
 	network, err := s.networks.Get(s.scope.Project(), spec.Name).Do()
+	autoCreateCloudNat := false
 	if gcperrors.IsNotFound(err) {
+		autoCreateCloudNat = true
```
jzhoucliqr, line 37: This is a behavior change. If it is a concern, one option is we introduce another flag in networkspec to indicate whether we should auto create cloudNAT, with that flag default to false for now.

cloud/services/compute/network.go
```diff
@@ -89,3 +98,60 @@ func (s *Service) DeleteNetwork() error {
 	s.scope.GCPCluster.Spec.Network.Name = nil
 	return nil
 }
+
+func (s *Service) createCloudNat(network *compute.Network) error {
```
jzhoucliqr, line 102: Is delete need to be handled ?

cloud/services/compute/network.go
detiber: That is true, however there has not been an official release for v0.3.0 yet, so behavioral changes may be fine. I will defer the decision to the community, though since I'm not sure how folks are currently using cluster-api-provider-gcp today and how the change might affect them.

cloud/services/compute/network.go
snehala27: ++ack

cloud/services/compute/network.go
```diff
@@ -78,6 +87,22 @@ func (s *Service) DeleteNetwork() error {
 		return nil
 	}
 
+	// Delete Router.
+	router, err := s.routers.Get(s.scope.Project(), s.scope.Region(), getRouterName(s.scope.NetworkName())).Do()
```
jzhoucliqr, line 91: shall we only delete the router when it is auto created by us?

snehala27: This flow is executed only when the network has **owned** tag. https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/f321e5d79c59b5122af808cd3d62d9e4b31f4749/cloud/services/compute/network.go#L86

jzhoucliqr:
@detiber @rsmitty Do you get a chance to take a look? Thanks

PR #309, opened 2020-07-29
Update e2e-conformance.sh to use image-builder deps-gce target for installing dependencies

detiber:

Update e2e-conformance.sh to use image-builder deps-gce target for installing dependencies

Depends on: https://github.com/kubernetes/test-infra/pull/18531 and an update to the kubekins-e2e image to include the new python3 packages before this will work correctly.

Fixes #

### DIFF
--- a/hack/ci/e2e-conformance.sh
+++ b/hack/ci/e2e-conformance.sh
@@ -196,33 +196,6 @@ init_image() {
     fi
   fi
 
-  if ! command -v ansible &> /dev/null; then
-    if [[ $EUID -ne 0 ]]; then
-      echo "Please install ansible and try again."
-      exit 1
-    else
-      # we need pip to install ansible
-      curl -L https://bootstrap.pypa.io/get-pip.py -o get-pip.py
-      python get-pip.py --user
-      rm -f get-pip.py
-
-      # install ansible needed by packer
-      version="2.8.5"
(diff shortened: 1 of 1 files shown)


detiber:
It looks like there is also an issue in image-builder with how the python2 fallbacks were being handled, created https://github.com/kubernetes-sigs/image-builder/pull/312 to remove those in favor of explicit use of python3/pip3

gab-satchi:
let's /hold until python3 bump is sorted out

ncdc:
Got past the ansible error. Now this:

```
hack/ensure-packer.sh
mkdir: cannot create directory '.local': Permission denied
```

Which comes from the following (from the image-builder repo):

```
# Change directories to the parent directory of the one in which this
# script is located.
cd "$(dirname "${BASH_SOURCE[0]}")/.."

source hack/utils.sh

if command -v packer >/dev/null 2>&1; then exit 0; fi

mkdir -p .local/bin && cd .local/bin
```

In this job, the image-builder repo is in `/home/prow/go/src/sigs.k8s.io/image-builder/images/capi`, which is one of the `extra_refs` from the Prow job. Maybe it's read-only?

ncdc:
Continuing this discussion in https://github.com/kubernetes-sigs/image-builder/issues/342

Issue #303, opened 2020-07-09
Route controller fails to create route for nodes in multi zone setup


snehala27:
/kind bug

**What steps did you take and what happened:**
Created a cluster using CAPG by setting the FailureDomain for worker nodes in MachineDeployment.
If the FailureDomain(Zones) are different for the control plane node and the worker node in the same region, route controller fails to create route for the node with error: `instance not found`


**What did you expect to happen:**
Routes should be created if different zone is selected for worker node in same region

**Anything else you would like to add:**
However if the control plane node and the worker nodes are created in same zone, route gets created and everything works fine.
Also the node status is set to **Ready** even when network is unavailable as a result of which new pods get deployed on the node with no network access

**Environment:**

- Cluster-api version: v1/alpha3
- Minikube/KIND version:
- Kubernetes version: (use `kubectl version`): 1.16.2
- OS (e.g. from `/etc/os-release`): ubuntu

kube-controller logs:

`I0709 14:21:22.711655       1 route_controller.go:194] Creating route for node test-workerpool-us-central1-a-c95e-dvpr5 192.168.2.0/24 with hint f3d3afc7-4dd0-4ccc-ba04-1c40d83eabfb, throttled 12.984µs`

`E0709 14:21:22.917150       1 route_controller.go:218] Could not create route f3d3afc7-4dd0-4ccc-ba04-1c40d83eabfb 192.168.2.0/24 for node test-workerpool-us-central1-a-c95e-dvpr5: instance not found`

`I0709 14:21:22.917353       1 event.go:255] Event(v1.ObjectReference{Kind:"Node", Namespace:"", Name:"test-workerpool-us-central1-a-c95e-dvpr5", UID:"test-workerpool-us-central1-a-c95e-dvpr5", APIVersion:"", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'FailedToCreateRoute' Could not create route f3d3afc7-4dd0-4ccc-ba04-1c40d83eabfb 192.168.2.0/24 for node test-workerpool-us-central1-a-c95e-dvpr5 after 205.46811ms: instance not found
`

Node Events:
                                                                                                                                                                                                                                  `Could not create route 9f3fe925-fbc2-4be9-ac9c-2b4706758f56 192.168.4.0/24 for node test-workerpool-us-central1-a-c95e-cm9h4 after 189.758088ms: instance not found`

PR #310, opened 2020-07-30
✨ Support FailureDomains override(optional) in GCPClusterSpec

snehala27:
Added a field FailureDomains in GCPClusterSpec which if provided, would override the default behaviour where the GCPClusterStatus.FailureDomains are defaulted to all the zones in the specified region. This change would set only the provided zones to true.

This PR would control the cluster components creation in the specified zones and thus unblock #303 to some extent.

### DIFF
--- a/controllers/gcpcluster_controller.go
+++ b/controllers/gcpcluster_controller.go
@@ -190,17 +190,21 @@ func (r *GCPClusterReconciler) reconcile(clusterScope *scope.ClusterScope) (ctrl
 	if err != nil {
 		return ctrl.Result{}, errors.Wrapf(err, "failed to get available zones for GCPCluster %s/%s", gcpCluster.Namespace, gcpCluster.Name)
 	}
+
+	// FailureDomains list should be empty by default.
 	gcpCluster.Status.FailureDomains = make(clusterv1.FailureDomains, len(zones))
+
+	// Iterate through all zones
 	for _, zone := range zones {
+		// If we have failuredomains in spec, see if this zone is in valid zone
+		// Add to the status _only_ if it's mentioned in the gcpCluster spec
 		if len(gcpCluster.Spec.FailureDomains) > 0 {
-			found := false
 			for _, fd := range gcpCluster.Spec.FailureDomains {
 				if fd == zone {
-					found = true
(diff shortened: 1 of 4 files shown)


controllers/gcpcluster_controller.go
```diff
@@ -192,8 +192,20 @@ func (r *GCPClusterReconciler) reconcile(clusterScope *scope.ClusterScope) (ctrl
 	}
 	gcpCluster.Status.FailureDomains = make(clusterv1.FailureDomains, len(zones))
 	for _, zone := range zones {
-		gcpCluster.Status.FailureDomains[zone] = clusterv1.FailureDomainSpec{
-			ControlPlane: true,
+		if len(gcpCluster.Spec.FailureDomains) > 0 {
+			found := false
+			for _, fd := range gcpCluster.Spec.FailureDomains {
```
rsmitty, line 197: I'm not quite sure this will have the desired effect as written. This would still allow for all failure domains in a given zone to get added, just the ones specifed in the spec wouldn't be eligible for controlplane scheduling. 

Wouldn't we rather _only_ add the failuredomains supplied in the spec to the list in the status field?

rsmitty: I'm thinking something more like:
```bash
gcpCluster.Status.FailureDomains = make(clusterv1.FailureDomains, len(zones))

// Iterate through all zones
for _, zone := range zones {

	// If we have failuredomains in spec, see if this zone is in that list
	// Add to the status _only_ if it's mentioned in the gcpCluster spec
	if len(gcpCluster.Spec.FailureDomains) > 0 {
		for _, fd := range gcpCluster.Spec.FailureDomains {
			if fd == zone {
				gcpCluster.Status.FailureDomains[zone] = clusterv1.FailureDomainSpec{
					ControlPlane: true,
				}
			}
		}
	// Default to adding all zones in failuredomain status
	} else {
		gcpCluster.Status.FailureDomains[zone] = clusterv1.FailureDomainSpec{
			ControlPlane: true,
		}
	}
}
```

snehala27: ++ack. Will update the branch.

PR #312, opened 2020-08-11
✨[api/gcpmachine] Support root volume type param.

dilyevsky:

Support confirurable root volume type as (optional) parameter within
GCPMachineSpec.

Supported types:
* `pd-standard` - Standard persistent disk (default)
* `pd-ssd` - SSD persistent disk

The above are only disk types currently supported by GCP as root volumes.


dilyevsky:
@detiber @justinsb friendly ping

api/v1alpha3/gcpmachine_types.go
```diff
@@ -77,6 +84,14 @@ type GCPMachineSpec struct {
 	// +optional
 	RootDeviceSize int64 `json:"rootDeviceSize,omitempty"`
 
+	// RootDeviceType is the type of the root volume.
+	// Supported types of root volumes:
+	// 1. "pd-standard" - Standrad (HDD) persistent disk
```
rsmitty, line 89: small typo here: "Standrad". looks like it also got copy/pasted in a few other places below.

rsmitty:
May be that this needs a rebase to pull in https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/309

api/v1alpha3/gcpmachine_types.go
dilyevsky: Thanks, fixed!

dilyevsky (commented):
Hey @rsmitty thanks for review, rebased and fixed typo. PTAL

config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmachines.yaml
```diff
@@ -101,6 +101,11 @@ spec:
                   Defaults to 30.
                 format: int64
                 type: integer
+              rootDeviceType:
+                description: 'RootDeviceType is the type of the root volume. Supported
+                  types of root volumes: 1. "pd-standard" - Standrad (HDD) persistent
```
rsmitty, line 106: Still a typo here and in config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmachinetemplates.yaml

dilyevsky: Doh, forgot to rerun `make generate`. Fixed (looks like it also picked up unrelated doc string change that was made previously.

PR #325, opened 2020-11-05
🏃 Update worker nodes for conformance test

cpanato:
Update worker nodes for conformance test, maybe makes sense when running conformance, i saw in other places that uses at least 5 nodes

also update the variables to match the k8s version we are using

/assing @detiber @rsmitty 

if that makes no sense feel free to close it

Fixes #


cpanato:
trying to see why the PR the conformance runs fine and in the periodic it fails

PR #186, opened 2019-09-29
✨ Support additionalMetadata on GCEInstance

justinsb:
This allows for specification of additional metadata, which is often used
as a tagging mechanism on GCE.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #342, opened 2020-12-20
Support additionalMetadata on an instance

justinsb:
Because of the limitations of labels & tags, metadata is often used to
store tracking information.

/kind feature
/kind api-change
```release-note
Support additionalMetadata on GCPMachine and GCPMachineTemplate, for
custom metadata fields.
```


justinsb:
This is a revamp of #186 that I lost track of.

@vincepri previously asked whether we should do `map[string]string` instead; I don't have a strong opinion, but my preference is for this mode because it better reflects the underlying API (which in theory supports null values) and also because it can support additional modes / fields (e.g. if we wanted to get a value from secret or configmap).

cpanato (approved):
/lgtm

thanks @justinsb

api/v1alpha3/gcpmachine_types.go
```diff
@@ -64,6 +64,11 @@ type GCPMachineSpec struct {
 	// +optional
 	AdditionalLabels Labels `json:"additionalLabels,omitempty"`
 
+	// AdditionalMetadata is an optional set of metadata to add to an instance, in addition to the ones added by default by the
+	// GCP provider.
+	// +optional
+	AdditionalMetadata []MetadataItem `json:"additionalMetadata,omitempty"`
```
detiber, line 70: I think this should probably also have some markers set to indicate that additionalMetadata should be treated as type `x-kubernetes-list-type` of `map`:
```suggestion
	// AdditionalMetadata is an optional set of metadata to add to an instance, in addition to the ones added by default by the
	// GCP provider.
	// +listType=map
	// +listMapKey=key
	// +optional
	AdditionalMetadata []MetadataItem `json:"additionalMetadata,omitempty"`
```

cpanato: @justinsb can you take a look at Jason's comment when you have a chance?

justinsb: Done - great point @detiber  - and thanks for the ping @cpanato !

cpanato (approved):
/lgtm

PR #343, opened 2020-12-21
WIP: Label GCP instances with GCPMachine UID

justinsb:
This allows for easy identification of the associated machine, even
across clusters, and can help with idempotent creation.

/kind feature

```release-note
Created GCP instances will now have a label k8s-io-owner-uid to
indicate the controlling GCPMachine object.
```

### DIFF
--- a/api/v1alpha3/labels.go
+++ b/api/v1alpha3/labels.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+
+	"k8s.io/apimachinery/pkg/types"
 )
 
 // Labels defines a map of tags.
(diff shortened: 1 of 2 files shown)


justinsb:
WIP because I'm not sure whether we have a label like this elsewhere we could leverage.  The intent is that we can use this to associate GCP resources with their controlling KRM object for all the GCP resources we create, I'm just starting with GCP instance.

Another WIP reason; it looks like this type is part of the apimachinery (e.g. deepcopy is being generated), I'm not sure if that's deliberate ... if so we probably have to put this field elsewhere (maybe in the Build method itself)

justinsb:
I'm actually also wondering if metadata is more appropriate, because of the warning on https://cloud.google.com/compute/docs/labeling-resources  "We do not recommend creating large numbers of unique labels, such as for timestamps or individual values for every API call. "

api/v1alpha3/labels.go
```diff
@@ -108,6 +110,9 @@ const (
 
 	// APIServerRoleTagValue describes the value for the apiserver role.
 	APIServerRoleTagValue = "apiserver"
+
+	// NameKubernetesObjectUID is the tag name for the UID of the "creating" kubernetes object.
+	NameKubernetesObjectUID = "k8s-io-owner-uid"
```
detiber, line 115: @justinsb I'd warn against using UID here, since the UID would change if the k8s resource is moved to a different Cluster API management cluster or if it is restored from a backup.

cpanato:
@justinsb @detiber this is something we still want? otherwise can we close?

PR #344, opened 2020-12-28
[gcpmachine]: Support additional drives.

### LINKED ISSUE #316: Allow for local-ssd
[Also submitted in openshift/cluster-api-provider-gcp](https://github.com/openshift/cluster-api-provider-gcp/issues/124)
/kind feature

Given that in GCP local-ssds are very performant and with a 3 year commitment less expensive than pd-standard (I do not believe you can get commitment discounts on other storage) I would like to use them for my /var (and use a small pd-standard boot disk).  The fact that they are ephemeral and I can not shutdown machines where they are present is not a concern for kubernetes nodes.

Unfortunately, the cluster-api does not support them. When I create a machine config with a disk of type local-ssd I get the following error message (in openshift which shares the same limitations with a different implementation):
```
error launching instance: googleapi: Error 400: Invalid value for field 'resource.disks[1].type': 'PERSISTENT'. Cannot create local SSD as persistent disk., invalid
```
Here is the spec/providerSpec/disk[1] definition I used:
```
            - autoDelete: true
              image: blank
              sizeGb: 375
              type: local-ssd
```
When I look at the allowed [configuration values](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/11b95277f3121dc377412c589b3d6316ca6bef54/cloud/services/compute/instances.go#L80) I don't see any way to specify the type as required by the [google api](https://cloud.google.com/compute/docs/disks/local-ssd#api).

Additionally, I could not find a way to keep the `sourceImage` and 'sizeGB' from being specified or set the `interface`.

I am hesitant to work on a PR because I don't use Kubernetes directly but Openshift instead.  Even though this limitation exists in both implementations those implementations appear to be very different.  The relationship between the two is not clear.  It also isn't clear if I submit code to one project I would be able to submit a code based on the same work product to the other (based on licensing.  Because of this I don't have a good way to dogfood my change.

dilyevsky:
/kind feature
/kind api-change

Adds `additionalDisks` object list parameter to `GCPMachineSpec` to allow auto-provisioning of additional disks upon GCPMachine creation. Following disks are supported:

* `pd-standard` (default)
*  `pd-ssd`
*  `local-ssd` (new) - this disk type can only be added as "additional disk" as it can not be a boot drive. 

Fixes #316

I have tested this with my test cluster with the following spec:
```
...
      additionalDisks:
      - deviceType: "local-ssd"
      - deviceType: "pd-ssd"
        size: 128
      - deviceType: "pd-standard"
---
```

which had the following disks created on the node:

Screen Shot 2020-12-27 at 10 41 16 PM

## Release notes 

```release-notes
Adds additionalDisks object list parameter to GCPMachineSpec to allow auto-provisioning of additional disks upon GCPMachine creation
```


dilyevsky:
@detiber @cpanato whenever you're back from the break, this is ready for review, thanks!

cpanato:
@dilyevsky i will try to review this week

api/v1alpha3/gcpmachine_types.go
```diff
@@ -92,6 +109,9 @@ type GCPMachineSpec struct {
 	// +optional
 	RootDeviceType *DiskType `json:"rootDeviceType,omitempty"`
 
+	// AdditionalDisks are optional non-boot attached disks.
```
detiber, line 112:
```suggestion
	// AdditionalDisks are optional non-boot attached disks.
	// +optional
```

api/v1alpha3/gcpmachine_types.go
```diff
@@ -33,8 +33,25 @@ type DiskType string
...
+	// 1. "pd-standard" - Standard (HDD) persistent disk
+	// 2. "pd-ssd" - SSD persistent disk
+	// 3. "local-ssd" - Local SSD disk (https://cloud.google.com/compute/docs/disks/local-ssd).
+	// Default is "pd-standard".
+	// +optional
+	DeviceType *DiskType `json:"deviceType,omitempty"`
+	// Size is the size of the disk in GBs.
+	// Defaults to 30GB. For "local-ssd" size is always 375GB.
+	// +optional
+	Size int64 `json:"size,omitempty"`
```
detiber, line 52: If this was a pointer, it would be easier to set a default using `k8s.io/utils/pointer` with `pointer.Int64PtrDerefOr()`.

cloud/services/compute/instances.go
```diff
@@ -86,7 +90,7 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*compute.Instance,
 				AutoDelete: true,
 				Boot:       true,
 				InitializeParams: &compute.AttachedDiskInitializeParams{
-					DiskSizeGb:  30,
+					DiskSizeGb:  defaultDiskSizeGB,
 					DiskType:    diskTypeURL(scope.Zone(), infrav1.PdStandardDiskType),
```
detiber, line 94:
```suggestion
					DiskType:    diskTypeURL(scope.Zone(), pointer.StringPtrDerefOr(scope.GCPMachine.Spec.RootDeviceType, infrav1.PdStandardDiskType)),
```

cloud/services/compute/instances.go
```diff
@@ -146,6 +150,32 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*compute.Instance,
 	if scope.GCPMachine.Spec.RootDeviceType != nil {
 		input.Disks[0].InitializeParams.DiskType = diskTypeURL(scope.Zone(), *scope.GCPMachine.Spec.RootDeviceType)
 	}
```
detiber, line 152:
```suggestion
```

cloud/services/compute/instances.go
```diff
@@ -146,6 +150,32 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*compute.Instance,
...
+			// For local SSDs set interface to NVME (instead of default SCSI) which is faster.
+			// Most OS images would work with both NVME and SCSI disks but some may work
+			// considerably faster with NVME.
+			// https://cloud.google.com/compute/docs/disks/local-ssd#choose_an_interface
+			ad.Interface = "NVME"
+		} else if d.Size > 0 {
+			ad.InitializeParams.DiskSizeGb = d.Size
+		}
+
+		input.Disks = append(input.Disks, ad)
```
detiber, line 177:
```suggestion
		ad := &compute.AttachedDisk{
			AutoDelete: true,
			InitializeParams: &compute.AttachedDiskInitializeParams{
				DiskSizeGb: d.Size,
				DiskType:   diskTypeURL(scope.Zone(), pointer.StringPtrDerefOr(d.DeviceType, infrav1.PdStandardDiskType)),
			},
		}
		
		if ad.InitializeParams.DiskSizeGb == 0 {
			ad.InitializeParams.DiskSizeGb = defaultDiskSizeGB
		}
		
		if ad.InitializeParams.DiskType == infrav1.LocalSsdDiskType{
			ad.Type = "SCRATCH" // Default is PERSISTENT.
			
			// Override the Disk size
			ad.InitializeParams.DiskSizeGb = 375

			// For local SSDs set interface to NVME (instead of default SCSI) which is faster.
			// Most OS images would work with both NVME and SCSI disks but some may work
			// considerably faster with NVME.
			// https://cloud.google.com/compute/docs/disks/local-ssd#choose_an_interface
			ad.Interface = "NVME"
		}

		input.Disks = append(input.Disks, ad)
```

cloud/services/compute/instances.go
dilyevsky: Ah cool, thanks!

api/v1alpha3/gcpmachine_types.go
dilyevsky: Ah good point, thanks. Should probably change that for root disk size in the future too...

dilyevsky: Unfortunately for DiskType type alias you had to do a ton of casting so I wrote custom helper for this.

dilyevsky (commented):
thanks @detiber PTAL

cpanato:
@dilyevsky maybe you can rebase and squash the commits

dilyevsky:
@cpanato squashed (was nothing to rebase)

cpanato (approved):
/lgtm
/assign @detiber

cpanato:
/release-note "Adds additionalDisks object list parameter to GCPMachineSpec to allow auto-provisioning of additional disks upon GCPMachine creation"

PR #349, opened 2021-02-18
Make AdditionalMetadata Key field required

hasheddan:
**What type of PR is this?**

/kind api-change

Using x-kubernetes-list-type with v1 CRDs requires that one specify that
the key field is either required or has a default value. In the case of
AdditionalMetadata, the key field should be required.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Partial revert of #342 

See https://github.com/kubernetes/kubernetes/issues/91395 for more info.

/assign @detiber @justinsb @cpanato 

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/v1alpha3/gcpmachine_types.go
+++ b/api/v1alpha3/gcpmachine_types.go
@@ -112,7 +112,7 @@ type GCPMachineSpec struct {
 // MetadataItem defines a single piece of metadata associated with an instance.
 type MetadataItem struct {
 	// Key is the identifier for the metadata entry.
-	Key string `json:"key,omitempty"`
+	Key string `json:"key"`
 	// Value is the value of the metadata entry.
 	Value *string `json:"value,omitempty"`
 }
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmachines.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmachines.yaml
@@ -82,6 +82,8 @@ spec:
                     value:
                       description: Value is the value of the metadata entry.
                       type: string
+                  required:
+                  - key
                   type: object
                 type: array
                 x-kubernetes-list-map-keys:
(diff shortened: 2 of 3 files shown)


hasheddan:
I may be missing something here as the original PR must have passed CI, but I was unable to `kubectl apply` CRDs locally in clusters 1.18+.

cpanato:
the tests are running on 1.19.x
`clusterctl config cluster capg-e2e-9ku7rt --infrastructure (default) --kubernetes-version v1.19.3 --control-plane-machine-count 3 --worker-machine-count 2 --flavor (default)`

hasheddan:
@cpanato I believe you are correct in that the cluster created by CAPG is v1.19.x. However, the bootstrap cluster is being created by `kind`, and I can tell from the [e2e logs](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-gcp/349/pull-cluster-api-provider-gcp-e2e-test/1362536243916705792#1:build-log.txt%3A1078) that `kind` `v0.7.1-0.20200303021537-981bd80d3802` is being used (corresponds to CAPI `v0.3.12`, which is determining it because we consume the `bootstrap` package. In `kind` `v0.7.x` the default k8s node is `v1.17.0` (see [here](https://github.com/kubernetes-sigs/kind/releases/tag/v0.7.0)), which I believe is why the issue observed in the PR is not surfacing (it would occur on installation of these CRDs into a 1.18+ cluster). I am happy to bump the version of CAPI / kind here too if desired.

cpanato:
ahh got it @hasheddan, I think worth upgrading the mgmt cluster version :) 
will take a look at that

detiber:
@hasheddan I'm trying to follow the linked issues and they seem to be related more to requiring the key to be a required field (which it already is). I'm wondering if we would be good if we just removed `omitempty` from the definition?

detiber:
I created a PR to upgrade kind to a newer version here: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/350

hasheddan:
@detiber I think that would work. I had mistakenly thought it was saying the whole slice must be required, which didn't make a lot of sense 😅  let me update and test it out real quick 👍

hasheddan:
@detiber updated and tested against a variety of v1.18+ k8s versions successfully

PR #350, opened 2021-02-19
update kind

detiber:
**What type of PR is this?**

/kind feature

Upgrades kind to a newer version to ensure we are doing validation around kubernetes list types

```release-note
NONE
```


detiber:
Closing, this doesn't affect the version of kind used for the e2e tests, that version is embedded in the e2e test suite.

There also doesn't appear to be a way to override the version of k8s used for the bootstrap cluster when using the e2e test suite currently...

/close

PR #355, opened 2021-03-31
Grab config.toml for containerd

dims:
Capture more information about containerd when we run the conformance CI job (basically capture the /etc/contained/config.toml) 

/kind feature

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


detiber:
@dims are you thinking that the changes to image-builder that modified the systemd cgroup driver may be the reason for the failed tests?

dims:
@detiber yes, please see fix in https://github.com/kubernetes-sigs/image-builder/pull/573

PR #399, opened 2021-07-17
use dl.k8s.io instead of hardcoded GCS URIs

spiffxp:
Use dl.k8s.io instead of hardcoded GCS URIs

**What type of PR is this?**

/kind cleanup

The only time a kubernetes GCS bucket name should be showing up in a
hardcoded URI is if `gsutil` is being used (e.g. gsutil cp gs://foo/bar .)

Otherwise, for tools like `curl` or `wget`, dl.k8s.io is much nicer for us
as a project, since we can transparently change where that redirects to
without having to change code everywhere

These changes will mean no changes will be necessary to accommodate a
`gs://kubernetes-release` -> `gs://k8s-release` migration equivalent of
the CI migration we're going through right now

These changes also address the `gs://kubernetes-release-dev` references
currently used by this repo (ref:
https://github.com/kubernetes/k8s.io/issues/2318)

I'm not sure that a release note is required for this, but I put one in
just in case

Similar / related PRs:
- https://github.com/kubernetes-sigs/image-builder/pull/656
- https://github.com/kubernetes-sigs/cluster-api/pull/4958

**TODOs**:

- [x] squashed commits
- [x] ~includes documentation~ (n/a)
- [x] ~adds unit tests~ (n/a)

**Release note**:

```release-note
kubernetes artifacts are pulled from https://dl.k8s.io instead of
hardcoded GCS bucket URIs (kubernetes-release, kubernetes-release-dev)

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ $(GINKGO): ## Build ginkgo.
 $(KUBECTL): ## Build kubectl
 	mkdir -p $(TOOLS_BIN_DIR)
 	rm -f "$(KUBECTL)*"
-	curl --retry $(CURL_RETRIES) -fsL https://storage.googleapis.com/kubernetes-release/release/$(KUBECTL_VER)/bin/$(GOOS)/$(GOARCH)/kubectl -o $(KUBECTL)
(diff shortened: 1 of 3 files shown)


cpanato (approved):
thanks Aaron
/approve
/lgtm

PR #421, opened 2021-09-23
External provider

sayantani11:
**What type of PR is this?**

This PR adds tools and template for external provider
Fixes #

We would also have to add a file in the docs section related to the same

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


CONTRIBUTING.md
```diff
@@ -1,71 +0,0 @@
-# Contributing Guidelines
```
cpanato, line 1: why removing the doc?

templates/create_cloud_conf.sh
```diff
@@ -0,0 +1,111 @@
+# Copyright 2020 The Kubernetes Authors.
```
cpanato, line 1: usually, the scripts stay in the `script` directory

templates/create_cloud_conf.sh
```diff
@@ -0,0 +1,111 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Provides tools for external cloud provider
+
+CAPO_SCRIPT=create_cloud_conf.sh
```
cpanato, line 17: why we define CAPO?

also this is not clear to me why we are adding this. can you give an explanation? maybe update the PR summary to add the reason we are adding this.

and add a readme to explain how to use that and what is the intent as well 

thanks so much!

cpanato (changes requested):
a few questions to clarify why we need this and how to use

sayantani11:
@cpanato So we here we are trying to deploy a cluster using [external cloud provider](https://github.com/kubernetes/cloud-provider-gcp). And we will be using [this](https://docs.projectcalico.org/v3.19/manifests/calico.yaml) to deploy our CNI Solution. The `create_cloud_.sh` will basically be used to create the cloud secrets using this configurations and we can move that to the `scripts` folder as well. I do plan to add a README in the docs folder, but I am stuck as I am having problems in setting up the cluster locally. 
As of now we can move forward with the YAML file only. We can add the `.sh` file after I add the external-cloud-provider.md file to the docs. 
And concerning the `CONTRIBUTING.md` file, it was in a separate branch but somehow I messed up and it came here as well, so I had to remove it as this PR was not concerning this, seems like the file is getting deleted from the repo and I don't know how can I fix this🤦‍♀️

sayantani11:
@cpanato If somehow you can ignore the `CONTRIBUTING.md` file from merging. My local branch is kind of corrupted and I am unable to remove it even after trying. It will be helpful of you can do it.

PR #463, opened 2021-11-09
Created ci-entrypoint.sh

Darshnadas:
**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/scripts/ci-entrypoint.sh
+++ b/scripts/ci-entrypoint.sh
@@ -1,0 +1,120 @@
+#!/bin/bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+###############################################################################
+
+# To run locally, set GCP_CLIENT_ID, GCP_CLIENT_SECRET, GCP_SUBSCRIPTION_ID, GCP_TENANT_ID
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${REPO_ROOT}" || exit 1
+
+# shellcheck source=hack/ensure-go.sh
+source "${REPO_ROOT}/hack/ensure-go.sh"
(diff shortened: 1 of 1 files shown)


cpanato:
@Darshnadas thanks for the pr, but can you explain why we need that and how to use it? i saw some variables that you define in the new script is might be defined in specific scripts like the e2e or conformance.

Darshnadas:
@cpanato I have not explored the project so deeply so I took help from the reference link of azure project and tried doing it. I won't be able to explain but if I get more elaborate overview of the project or the thing I might be able to work more properly. Also there were many files which were present in azure project but were not present in gcp so because of that also I got confused in a lot of areas. As this is my first PR in this project little bit more explanation and of course more exploring from my side to the project will help me getting involved more and I can work on this PR for accurately.

scripts/ci-entrypoint.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
```
cpanato, line 1: i think a ci entrypoint will the good to have to make a standard entry point for e2e/conformance tests.

on those scripts we have some other things that are in common like setup the gcp network and build the node image if needed. maybe we can add those here and remove the duplication from those scripts. wdyt?
then those scripts should inkove this in the start to setup all the requirements.

scripts/ci-entrypoint.sh
```diff
@@ -0,0 +1,120 @@
...
+    export GCP_REGION="<GCP_REGION>"
+    export GCP_PROJECT="<GCP_PROJECT>"
+    export KUBERNETES_VERSION=1.20.9
+    export GCP_CONTROL_PLANE_MACHINE_TYPE=n1-standard-2
+    export GCP_NODE_MACHINE_TYPE=n1-standard-2
+    export GCP_NETWORK_NAME=<GCP_NETWORK_NAME or default>
+    export CLUSTER_NAME="<CLUSTER_NAME>"
+}
+
+wait_for_nodes() {
```
cpanato, line 71: this is not needed

Darshnadas:
Hi @cpanato Thank you so much reviewing this, and sorry for the late reply, will update the changes soon

scripts/ci-entrypoint.sh
Darshnadas: Hi, @cpanato will it be okay if you can explain me a bit more in a detailed way? this is my first time writing a script, so what things should be added if you could explain me more clearly, it will be helpful

scripts/ci-entrypoint.sh
Darshnadas: @cpanato from which line number shall I check that these are not needed?

scripts/ci-entrypoint.sh
Darshnadas: It will be easier for me if I get an example of what things to use for the following

PR #597, opened 2022-05-09
Root Disk Image lookup should use a filter on image name rather than family

### LINKED ISSUE #293: Root Disk Image lookup should use a filter on image name rather than family when Version is specified but Spec.Image and Spec.ImageFamily are not
/kind bug

When only GCPMachine.Spec.Version is defined (without also defining Spec.Image or Spec.ImageFamily), the root disk lookup defaults to using image family, which groups all images by kubernetes major and minor versions (ignoring patch version), instead we should fall back to using filtering on the name to also include the patch version as well as sorting by creation timestamp to choose the newest image that matches the kubernetes major/minor/patch version combination.

aniruddha2000:
**What type of PR is this?**
/kind bug

This PR look for Spec.Image instead of Spec.ImageFamily when when Version is specified but Spec.Image and Spec.ImageFamily are not. Also it include the patch version as well instead of current major minor version.
 
Fixes #293 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Root Disk Image lookup should use a filter on image name rather than family and include patch version
```

### DIFF
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -210,8 +210,8 @@ func (m *MachineScope) InstanceImageSpec() *compute.AttachedDisk {
 	if m.Machine.Spec.Version != nil {
 		version = *m.Machine.Spec.Version
 	}
-	image := "capi-ubuntu-1804-k8s-" + strings.ReplaceAll(semver.MajorMinor(version), ".", "-")
-	sourceImage := path.Join("projects", m.ClusterGetter.Project(), "global", "images", "family", image)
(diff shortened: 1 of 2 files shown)


cpanato:
thanks
/approve
/assign @pydctw

pydctw:
I don't have much context on image side of CAPG but reading the issue, how are multiple images with same version but different time stamps are handled? @aniruddha2000 

> we should fall back to using filtering on the name to also include the patch version as well as sorting by creation timestamp to choose the newest image that matches the kubernetes major/minor/patch version combination.

PR #616, opened 2022-06-06
Improve resilience of cluster delete

AverageMarcus:
**What type of PR is this?**

/kind bug

We've experienced some issue where deleting of a CAPG cluster results in the delete reconciliation getting stuck. 

In one scenario the InstanceGroup is removed without all the instances being removed and upon the next reconcile loop the code error's out when trying to find the InstanceGroup. This PR allows that to silently fail as we want to continue with the clean up of other resources anyway.

Another issue we saw was the `.Status.Network.SelfLink` being cleared but subsequent delete reconcile loops attempting to use it to build the firewall rules object - resulting in a nil pointer panic. Rather than referencing the SelfLink, we instead build the partial URL from the known project and network name.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -115,6 +115,11 @@ func (s *ClusterScope) NetworkName() string {
 	return pointer.StringDeref(s.GCPCluster.Spec.Network.Name, "default")
 }
 
+// NetworkLink returns the partial URL for the network.
+func (s *ClusterScope) NetworkLink() string {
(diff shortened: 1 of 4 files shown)


pydctw:
I have observed the issue described during cluster deletion so thanks for the PR 🎉.

/lgtm

cpanato (approved):
thanks
/approve

Issue #622, opened 2022-06-10
GitHub Link Check throw 403 for github.com


hrittikhere:
/kind bug

**What steps did you take and what happened:**
I think there might be some issue with the action implementation for GitHub related links as the link can be perfectly accessed via a [WebUI](https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key). Now after further investigation via curl we can see:

```html
<div id='message'>
<h2>The request is blocked.</h2></div>
<div id='errorref'>
<span>0/ByiYgAAAAD0Jl+OyjIrTI2PolL1vAbHTU5aMjIxMDYwNjE0MDI5ADU5NmQ3OGEyLWNhNWYtNDc5ZC1iY2RjLTA4MzU4MzMxNzRiMg==</span>
</div>
```
![image](https://user-images.githubusercontent.com/67012359/172895559-38a4d295-064e-4164-b68d-942e5f85fdb6.png)

403 for github.com because it blocks requests 

**What did you expect to happen:**
Link Checks for GitHub will pass and not throw a 403

PR #623, opened 2022-06-10
Fix GitHub Actions for Links 

hrittikhere:
/kind bug
/kind documentation

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/622


hrittikhere:
So tests do pass. Now if we can merge this and run all the checks again for docs we can find if our problem is solved or not. For local tests they stand to be solved

cpanato:
/approve
/lgtm

thanks

PR #662, opened 2022-07-19
Fix panic when no failure domain found

alexander-demichev:
**What type of PR is this?**
/kind bug

We are using [externally managed cluster infrastructure](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210203-externally-managed-cluster-infrastructure.md) and nothing sets failure domain on cluster object, this makes controller panic when it's missing. I believe the controller should not panic here and fail in another place in this case.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix panic when no failure domain found
```

### DIFF
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -95,6 +95,9 @@ func (m *MachineScope) Cloud() cloud.Cloud {
 func (m *MachineScope) Zone() string {
 	if m.Machine.Spec.FailureDomain == nil {
 		fd := m.ClusterGetter.FailureDomains()
+		if len(fd) == 0 {
+			return ""
+		}
 		zones := make([]string, 0, len(fd))
 		for zone := range fd {
 			zones = append(zones, zone)


alexander-demichev:
@richardcase gcp cluster controller also tries to populate the failure domain field https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/controllers/gcpcluster_controller.go#L179-L196 but regardless of that, if the payload is invalid it's better to fail somewhere on the GCP side than panic :)

richardcase:
>  if the payload is invalid it's better to fail somewhere on the GCP side than panic :)

Agreed that a panic isn't good. I was trying to highlight that we should review our usage of `Zone()` more generally as it appears in some instances its not needed.

cpanato:
@richardcase are you fine with this change?

richardcase:
@cpanato - yes all good :+1: 

Worth a try.....

/lgtm

cpanato (approved):
/approve

PR #664, opened 2022-07-20
fix: always send autoCreateSubnetnetworks field to create VPC network

fiunchinho:
/kind bug

The controller tries to create a legacy network instead of a VPC network when setting `autoCreateSubnetnetworks` to false, because the field is not sent to the google cloud API.

```release-note
Always send autoCreateSubnetnetworks to gcloud to force VPC network creation
```


cpanato (approved):
/approve
/lgtm

PR #665, opened 2022-07-20
fix: use right name for EnableFlowLogs field

fiunchinho:
/kind bug

It seems like a copy paste error. The field name doesn't match with the field name when marshalling. I think this is a breaking change.

```release-note
Fix EnableFlowLogs field name
```


cpanato:
/ok-to-test

thanks!

cpanato (approved):
/approve
/lgtm

PR #713, opened 2022-09-06
Allow to customize the subnet purpose

fiunchinho:
**What type of PR is this?**

/kind feature

It's useful to define the `purpose` of the subnets. Some features on google cloud require subnets to be created with a specific purpose, like [using envoy for load balancing](https://cloud.google.com/load-balancing/docs/proxy-only-subnets).

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Allow customization of the subnet purpose
```


cpanato (commented):
/ok-to-test

api/v1beta1/types.go
```diff
@@ -148,6 +148,16 @@ type SubnetSpec struct {
...
+
+	// Purpose: The purpose of the resource. This field can be either
+	// PRIVATE_RFC_1918 or INTERNAL_HTTPS_LOAD_BALANCER. A subnetwork with
+	// purpose set to INTERNAL_HTTPS_LOAD_BALANCER is a user-created
+	// subnetwork that is reserved for Internal HTTP(S) Load Balancing. If
+	// unspecified, the purpose defaults to PRIVATE_RFC_1918. The
+	// enableFlowLogs field isn't supported with the purpose field set to
+	// INTERNAL_HTTPS_LOAD_BALANCER.
+	// +optional
+	Purpose *string `json:"purpose,omitempty"`
```
pydctw, line 160: Could we add kubebuilder validation if there are only two valid values?
```
// +kubebuilder:validation:Enum=PRIVATE_RFC_1918;INTERNAL_HTTPS_LOAD_BALANCER
```

pydctw (commented):
Looks good with a minor comment

api/v1beta1/types.go
fiunchinho: The comment was a copy from the google cloud SDK but it seems the comment was not up to date, because it said there were two possible values, then later listed all the possible values.

I updated the comment now and added the kubebuilder validation to only accept the possible values.

pydctw: Thanks for looking into it. Looks good to me.

cpanato:
/approve

thanks!

PR #724, opened 2022-09-20
Subnet role must be passed

fiunchinho:
**What type of PR is this?**

/kind bug

When creating a new [proxy-only subnet](https://cloud.google.com/load-balancing/docs/proxy-only-subnets) using the `REGIONAL_MANAGED_PROXY` purpose, we need to pass the subnet Role.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Set subnet role to ACTIVE
```


cpanato (commented):
/ok-to-test

Issue #906, opened 2023-04-30
Support additional fields for GKE


akshay196:
/kind feature

**Describe the solution you'd like**

CAPG initial support for Managed Cluster (GKE) has added few basic configurations. GCP does support lot more fields for example private cluster, enable workload identity, just to name a few. This is a request to add support for more fields than just basics.

[This spreadsheet](https://docs.google.com/spreadsheets/d/1ETX0JyIL5jiOWPUrxOSOJ8yCzXvSERp2bZ3jPn33Vdo/edit#gid=0) enlists all important GKE  configuration we can support in GCPManaged* kinds.

All additional GKE fields are documented in the spreadsheet linked above. The structure in spreadsheet can be used for defining GCPManaged API specs. Feel free to comment to discuss more about any fields.

**Anything else you would like to add:**

There are other open issues those requested for additional fields too:
- https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/903
- https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/897

Such requests can be tackled by the complete list (spreadsheet) attached here.

If this sounds good then, I am happy to take this forward and decide sub-tasks for better collaboration.

looping @poornima-rafay @niravparikh05

**Progress**

```[tasklist]
### Tasks - PR1
- [x] GCPManagedMachinePool changes (@akshay196) - https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/944
```

```[tasklist]
### Tasks - PR2
- [ ] GCPManagedControlPlane changes - part 1 (Description, ClusterNetwork, DefaultNodeLocation, DefaultMaxPodsPerNode) - ~`PR#936` by Anniruddha~ (@akshay196) - https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1186
```

```[tasklist]
### Tasks - PR3
- [ ] GCPManagedControlPlane changes - part 2 (ClusterSecurity, AddonsConfig, LoggingConfig, MonitoringConfig) (@akshay196) - https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1186
```

```[tasklist]
### Tasks - PR4
- [ ] Update relevant documentation if any
```

aniruddha2000:
@akshay196 I am happy to get involved in the dividing task discussion and take some of them.

richardcase:
That spreadsheet is really helpful @akshay196 :+1: 

[...]

PR #936, opened 2023-05-31
Add fields to the GCPManagedControlPlane types

aniruddha2000:
**What type of PR is this?**
/kind api-change

This PR contains the necessary API changes for GCPManagedControlPlane types to support additional fields.

Part of #906 

- Do we need to rename `ClusterNetwork.Pod` to a more reasonable name? Because it is quite ambiguous to the upstream k8s Pod.
- ``CidrBlock map[string]string `json:"cidrBlock,omitempty"`` must not be greater than 10. What should be the kubebuilder validation?

**TODOs**:

- [ ] squashed commits
- [x] includes documentation

**Release note**:

```release-note
Add necessary API changes for GCPManagedControlPlane types to support additional fields.
```

/cc @akshay196 @richardcase @cpanato


exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+
+	// ImageStreaming whether to use GCFS (Google Container File System).
+	// +optional
+	ImageStreaming bool `json:"ImageStreaming,omitempty"`
+}
+
+// LoggingConfig defines the logging on Cluster.
+type LoggingConfig struct {
+	// Enable define whether enable logging to cluster or not.
+	// +kubebuilder:default=true
```
akshay196, line 193: No need to enable logging by default. Let user enable it explicitly if they wants to. Logging and monitoring operations are [priced](https://cloud.google.com/stackdriver/pricing).
 Can we remove this marker?

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// +optional
+	Enable bool `json:"enable,omitempty"`
+
+	// EnableComponents select components to collect logs. An empty set would disable all logging. The system logs are
+	// minimum required component for enabling log collections. Only honored when enabled=true.
+	// +kubebuilder:validation:Enum=SYSTEM_COMPONENTS;WORKLOADS;APISERVER;SCHEDULER;CONTROLLER_MANAGER
+	// +kubebuilder:validation:MinItems=1
```
akshay196, line 200: This could be empty.
Can we remove this marker?

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// +optional
+	EnableComponents []string `json:"enableComponents,omitempty"`
+}
+
+// MonitoringConfig defines the monitoring on Cluster.
+type MonitoringConfig struct {
+	// Enable is whether enable monitoring to cluster or not.
+	// +kubebuilder:default=true
```
akshay196, line 208: Same comment as above.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// +optional
+	Enable bool `json:"enable,omitempty"`
+
+	// EnableComponents select components to collect metrics. An empty set would disable all monitoring. System metric
+	// is the minimum required component for enabling metric collection. Only honored when enabled=true.
+	// +kubebuilder:validation:Enum=SYSTEM_COMPONENTS;APISERVER;SCHEDULER;CONTROLLER_MANAGER
+	// +kubebuilder:validation:MinItems=1
```
akshay196, line 215: Same comment as logging components.

akshay196 (changes requested):
@aniruddha2000 Few minor comments.

> Do we need to rename ClusterNetwork.Pod to a more reasonable name? Because it is quite ambiguous to the upstream k8s Pod.

ClusterNetwork.Pod and ClusterNetwork.Service names are inspired by [Cluster.ClusterNetwork section](https://github.com/kubernetes-sigs/cluster-api/blob/75c29acbb784ce26daf05d670d949eb44f76b439/api/v1beta1/cluster_types.go#L283-L289) in Cluster-API. However we can rename it to something like PodConfig and ServiceConfig as it encapsulates pod and service networking configuration.

> CidrBlock map[string]string `json:"cidrBlock,omitempty" must not be greater than 10. What should be the kubebuilder validation?

See if `+kubebuilder:validation:MaxProperties` is useful here? (ref: https://book.kubebuilder.io/reference/markers/crd-validation.html)
Otherwise we may add webhook validation.

akshay196:
Regarding apidiff CI check, 
```
   Incompatible changes:
  - GCPManagedControlPlaneSpec: old is comparable, new is not 
```
The struct is comparable if all fields are comparable ([go spec](https://go.dev/ref/spec#Comparison_operators)). So converting fields like `ClusterNetwork` to pointer (`*ClusterNetwork`) would fix it. However `DefaultNodeLocation` is a slice. Slice types are not comparable.

@richardcase Any idea how to resolve this?

richardcase:
So we don't merge too early, without the controller changes (which will be done in another PR):

/hold

richardcase:
As this is for GKE and it's experimental, we can ignore apidiff changes:

/override  pull-cluster-api-provider-gcp-apidiff

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
 	ManagedControlPlaneFinalizer = "gcpmanagedcontrolplane.infrastructure.cluster.x-k8s.io"
 )
 
+// PrivateCluster defines a private Cluster.
+type PrivateCluster struct {
+	// Enabled defines weather a cluster is private or not. If disabled cluster network is public.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
richardcase, line 34: Should we split this into nodes and endpoint?

If a user declares `privateCluster` (which is optional) in their yaml then we can assume they want a private network, so we could add a validation to the webhook to check that one of private nodes or private endpoint is true. 

We could also add a validating webhook, that is

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// +optional
+	ControlPlanePrivateEndpoint bool `json:"controlPlanePrivateEndpoint,omitempty"`
+
+	// DisableDefaultSNAT is disables cluster default sNAT rules. Honored when enabled is true.
+	// +optional
+	DisableDefaultSNAT bool `json:"disableDefaultSNAT,omitempty"`
+}
+
+// AuthorizedNetworkCidrBlock defines the CIDR block configuration.
+type AuthorizedNetworkCidrBlock struct {
```
richardcase, line 56: If someone is specifying a cidr block i'm guessing that at least one of Name or CidrBlock is required.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+
+// AuthorizedNetwork provide an IP-based firewall that controls access to the GKE control plane.
+type AuthorizedNetwork struct {
+	// Enabled defines Whether enable authorized network or not.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
+
+	// CidrBlocks defines the CIDR block configuration.
+	// +optional
+	CidrBlocks AuthorizedNetworkCidrBlock `json:"cidrBlocks,omitempty"`
```
richardcase, line 74: If we have an authorized network then i assume its required to specify cidr blocks?

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	CidrBlock string `json:"cidrBlock,omitempty"`
+}
+
+// ClusterNetworkService defines the range of CIDRBlock list from where it gets the IP address.
+type ClusterNetworkService struct {
+	// CidrBlock is where cluster services will be assigned an IP address from this IP address range. Enter a range
+	// (in CIDR notation) within a network range, a mask, or leave this field blank to use a default range.
+	// This setting is permanent.
+	// +optional
+	CidrBlock string `json:"cidrBlock,omitempty"`
```
richardcase, line 92: Same comment as with the pod cidr block.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	PrivateCluster PrivateCluster `json:"privateCluster,omitempty"`
+
+	// UseIPAliases is whether alias IPs will be used for pod IPs in the cluster. If false, routes will be used for
+	// pod IPs in the cluster.
+	// +optional
+	UseIPAliases bool `json:"useIPAliases,omitempty"`
+
+	// AuthorizedNetwork provide an IP-based firewall that controls access to the GKE control plane.
+	// +optional
+	AuthorizedNetwork AuthorizedNetwork `json:"authorizedNetwork,omitempty"`
```
richardcase, line 108: If this is optional we can make this a pointer.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// +optional
+	Service ClusterNetworkService `json:"service,omitempty"`
+}
+
+// WorkloadIdentityConfig allows workloads in your GKE clusters to impersonate Identity and Access Management (IAM)
+// service accounts to access Google Cloud services.
+type WorkloadIdentityConfig struct {
+	// Enable is to enable the workload identity config.
+	// +optional
+	Enable bool `json:"enable,omitempty"`
```
richardcase, line 124: As a general comment we probably don't need `Enable` fields if the parent struct is optional.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+
+	// WorkloadPool is the workload pool to attach all Kubernetes service accounts to Google Cloud services.
+	// Only relevant when enabled is true
+	// +optional
+	WorkloadPool string `json:"workloadPool,omitempty"`
```
richardcase, line 129: Assuming that this is required if we want to use workload identity. 

As general comment, if a struct is used to represent an optional feature like workload identity then we can make fields required as the parent is optional.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+}
+
+// AuthenticatorGroupConfig is RBAC security group for use with Google security groups in Kubernetes RBAC.
+type AuthenticatorGroupConfig struct {
```
richardcase, line 133: Some comments as previous

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// +optional
+	SecurityGroups string `json:"securityGroups,omitempty"`
+}
+
+// ClusterSecurity defines the cluster security.
+type ClusterSecurity struct {
+	// WorkloadIdentityConfig allows workloads in your GKE clusters to impersonate Identity and Access Management (IAM)
+	// service accounts to access Google Cloud services
+	// +optional
+	WorkloadIdentityConfig WorkloadIdentityConfig `json:"workloadIdentityConfig,omitempty"`
```
richardcase, line 148: As its optional lets use a pointer.

richardcase: Same for below fields for structs

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// +optional
+	EnableLegacyAuthorization bool `json:"enableLegacyAuthorization,omitempty"`
+
+	// IssueClientCertificate is weather to issue a client certificate.
+	// +optional
+	IssueClientCertificate bool `json:"issueClientCertificate,omitempty"`
+}
+
+// AddonsConfig defines the enabled Cluster Addons.
+type AddonsConfig struct {
```
richardcase, line 164: Do we need to consider the other options:

https://cloud.google.com/go/docs/reference/cloud.google.com/go/container/latest/apiv1/containerpb#cloud_google_com_go_container_apiv1_containerpb_WorkloadIdentityConfig

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
richardcase: Instead of requiring the user to specify the strings it might be nice to have individual fields instead as it gives a better UX. For example: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/eks/api/v1beta2/types.go#L30

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
richardcase: Same comment from me, as can have individual fields

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// +optional
+	ClusterNetwork ClusterNetwork `json:"clusterNetwork,omitempty"`
+
+	// ClusterSecurity defines the cluster security.
+	// +optional
+	ClusterSecurity ClusterSecurity `json:"clusterSecurity,omitempty"`
+
+	// AddonsConfig defines the enabled Cluster Addons.
+	// +optional
+	AddonsConfig AddonsConfig `json:"addonsConfig,omitempty"`
```
richardcase, line 246: As its optional we can use a pointer

richardcase: And similar for below fields for structs

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
aniruddha2000: Good Idea!

Just to clarify, are you talking about
```go
        // ControlPlaneCidrBlock is the IP range in CIDR notation to use for the hosted master network. This range must not
	// overlap with any other ranges in use within the cluster's network. Honored when enabled is true.
	// +optional
	ControlPlaneCidrBlock string `json:"controlPlaneCidrBlock,omitempty"`

	// ControlPlaneGlobalAccess is whenever master is accessible globally or not. Honored when enabled is true.
	// +optional
	ControlPlaneGlobalAccess bool `json:"controlPlaneGlobalAccess,omitempty"`
```

these two fields in the Node and last two `ControlPlanePrivateEndpoint` & `DisableDefaultSNAT` in the Network, if I am not wrong.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
aniruddha2000: Yes, adding a MinProperties make sense right?
```diff
+ // +kubebuilder:validation:MinProperties=1
// +kubebuilder:validation:MaxProperties=10
// +optional
CidrBlock map[string]string `json:"cidrBlock,omitempty"`
```

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	// CidrBlock is the list of CIDR blocks associated with Authorized Network.
+	// +optional
+	CidrBlock map[string]string `json:"cidrBlock,omitempty"`
+}
+
+// AuthorizedNetwork provide an IP-based firewall that controls access to the GKE control plane.
+type AuthorizedNetwork struct {
+	// Enabled defines Whether enable authorized network or not.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
aniruddha2000: That's correct, @akshay196 what do you think?

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -27,13 +27,241 @@ const (
...
+	CidrBlocks AuthorizedNetworkCidrBlock `json:"cidrBlocks,omitempty"`
+}
+
+// ClusterNetworkPod the range of CIDRBlock list from where it gets the IP address.
+type ClusterNetworkPod struct {
+	// CidrBlock is where all pods in the cluster are assigned an IP address from this range. Enter a range
+	// (in CIDR notation) within a network range, a mask, or leave this field blank to use a default range.
+	// This setting is permanent.
+	// +optional
+	CidrBlock string `json:"cidrBlock,omitempty"`
```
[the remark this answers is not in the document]
aniruddha2000: In the [spreadsheet](https://docs.google.com/spreadsheets/d/1ETX0JyIL5jiOWPUrxOSOJ8yCzXvSERp2bZ3jPn33Vdo/edit#gid=0) it is `string` and in the CAPI it is `[]string`, does it count the diff? or using the CAPI one make more sense?

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
akshay196: @richardcase Sorry I don't quite understand what you mean by splitting into nodes and endpoint. Can you please explain further or give an example?
The private cluster has 4 properties shown in picture out of which ContolPlaneCidrBlock is mandatory. I clubbed them into private cluster group.

Screenshot from 2023-06-06 22-11-43

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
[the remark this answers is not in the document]
akshay196: @richardcase yes, you're correct. We could remove enable and rely on whether `authorizedNetwork` is specified or not. However separate `Enabled` field would allow user to directly enable/disable authorized network without need to remove cidr blocks.
I am fine either ways.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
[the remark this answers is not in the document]
akshay196: Note that GKE allows defining only single pod address range.

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
aniruddha2000: Sorry, I don't understand this :(

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
richardcase: I'm talking about `EnablePrivateNodes` and `EnablePrivateEndpoint` on the API call: https://cloud.google.com/go/docs/reference/cloud.google.com/go/container/latest/apiv1/containerpb#cloud_google_com_go_container_apiv1_containerpb_PrivateClusterConfig

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
richardcase: Thanks @akshay196 👍

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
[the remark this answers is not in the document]
richardcase: People may get confused if they can supply the pod cidr block on the CAPI Cluster and here.

I wonder if you just take the first if there are multiple and warn the user?

richardcase:
This is great work 👍

akshay196:
New PR: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1186
We may close this.

/close

Issue #982, opened 2023-07-13
Enable Shared VPC networking feature


bighiro:
/kind feature

**Describe the solution you'd like**
Need to enable shared VPC networking feature.
Currently, the api limit the users to create a network or use existing network in the same project where the cluster are getting created. Need a feature to access the network from project hosting the networks and use the network/subnetworks/firewalls that already exists in the host project.

**Anything else you would like to add:**
https://cloud.google.com/vpc/docs/shared-vpc

PR #991, opened 2023-07-27
Add shared VPC networking

### LINKED ISSUE #861: Can i configured GCP Instance Network Settings more freely?
/kind feature

Hi, When creating a cluster, I would like to have more freedom in entering instance information.
Because I need to use a shared VPC from another project. And I don't even need a router.

The way I want it is as follows
1. Using a Shared VPC in another project
or 
2. Using predefined instance templates

Is there any way?
Thanks.

bighiro:
**What type of PR is this?**
/kind feature

Support shared VPC networking.

Fixes #861 #982 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Support shared VPC networking.
Able to deploy the cluster with shared VPC network from another project that hosts the networks.
```

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -93,6 +93,9 @@ type NetworkSpec struct {
 	// +optional
 	Name *string `json:"name,omitempty"`
 
+	// HostProject  is the name of the host project hosting the shared VPC network
+	HostProject *string `json:"hostProject,omitempty"`
+
 	// AutoCreateSubnetworks: When set to true, the VPC network is created
 	// in "auto" mode. When set to false, the VPC network is created in
 	// "custom" mode.
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -94,6 +94,11 @@ func (m *MachineScope) Cloud() cloud.Cloud {
 	return m.ClusterGetter.Cloud()
 }
 
+// NetworkCloud returns initialized network cloud
+func (m *MachineScope) NetworkCloud() cloud.Cloud {
+	return m.ClusterGetter.NetworkCloud()
+}
+
 // Zone returns the FailureDomain for the GCPMachine.
 func (m *MachineScope) Zone() string {
 	if m.Machine.Spec.FailureDomain == nil {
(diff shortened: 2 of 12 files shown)


richardcase:
Thanks for this @bighiro :bow: 

There are a number of jobs failed. Could you do the following:

- `make generate`
- `make lint`
- `make test`

This will highlight some issues that will need to be fixed. Additionally you will need to update the conversions manually as the new field `HostProject` is failing in the roundtrip fuzzing tests. You will need to add something similar to this: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/53a8cb0a8867bf0a87066297008a81ad70385d21/api/v1alpha4/gcpcluster_conversion.go#L41 for the new field so it can be restored.

Reach out here or in our slack channel if you need any help.

akshay196:
Adding comment as I happened to worked on GKE with shared VPC recently.

This PR will affect the GKE implementation too as NetworkSpec is common for un-managed and managed GCP clusters. One thing I have seen is that it becomes mandatory to provide secondary Pod and Service ranges for GKE (both auto-pilot and standard) if the cluster is using shared VPC. [1]

So we need to verify this PR with GKE too if that is also in focus of this PR.

[1] https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#verify_usable_subnets

richardcase:
@bighiro - just folllowing up on if you'd be able to review the comments?

api/v1beta1/types.go
```diff
@@ -93,6 +93,9 @@ type NetworkSpec struct {
 	// +optional
 	Name *string `json:"name,omitempty"`
 
+	// HostProject  is the name of the host project hosting the shared VPC network
```
barbacbd, line 96:
```suggestion
	// HostProject  is the name of the host project hosting the shared VPC network resources.
	// +optional
```

cloud/scope/machine.go
```diff
@@ -94,6 +94,11 @@ func (m *MachineScope) Cloud() cloud.Cloud {
 	return m.ClusterGetter.Cloud()
 }
 
+// NetworkCloud returns initialized network cloud
```
barbacbd, line 97:
```suggestion
// NetworkCloud returns initialized network cloud.
```

cloud/scope/managedcluster.go
```diff
@@ -118,9 +123,22 @@ func (s *ManagedClusterScope) NetworkName() string {
...
 }
 
+// NetworkProject returns the cluster network unique identifier.
+func (s *ManagedClusterScope) NetworkProject() string {
+	return pointer.StringDeref(s.GCPManagedCluster.Spec.Network.HostProject, s.GCPManagedCluster.Spec.Project)
+}
+
+// IsSharedVpc returns the cluster network unique identifier.
+func (s *ManagedClusterScope) IsSharedVpc() bool {
+	if s.NetworkProject() != s.Project() {
```
barbacbd, line 133: nit: 
```suggestion
	return s.NetworkProject() != s.Project
```

cloud/services/compute/subnets/reconcile.go
```diff
@@ -41,6 +41,9 @@ func (s *Service) Reconcile(ctx context.Context) error {
 // Delete deletes cluster subnetwork components.
 func (s *Service) Delete(ctx context.Context) error {
 	logger := log.FromContext(ctx)
+	if s.scope.IsSharedVpc() {
+		return nil
```
barbacbd, line 45: Nit: Log/Alert that this is a shared VPC.

PR #1087, opened 2023-12-01
Fix nodeNetwork GCPManagedMachinePool validations

kahun:
**What type of PR is this?**
/kind bug

If the nodeNetwork field is empty in the GCPManagedMachinePool object, it will throw this error: 
"spec.nodeNetwork.podRangeCidrBlock: Invalid value: "null": spec.nodeNetwork.podRangeCidrBlock in body must be of type string: "null".

**TODOs**:
- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Fix GCPManagedMachinePool nodeNetwork field validation.
```


iamjanr:
Hi @richardcase.
Do you think this PR could be merged?

PR #1098, opened 2023-12-12
Fix GKE update version check

kahun:
**What type of PR is this?**
/kind bug

The GKE update process enters an infinite loop during the reconcile process because it compares desiredMasterVersion to InitialClusterVersion instead of CurrentMasterVersion.

**TODOs**:
- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Use current cluster version when upgrading a GKE cluster instead of the initial version. 
```


richardcase:
Thanks for this @kahun 🙇 With a fix like this its helpful to add a release note instead of using **NONE** because this will then get highlighted when we do a release. I have updated the release note for this PR.

kahun:
> Thanks for this @kahun 🙇 With a fix like this its helpful to add a release note instead of using **NONE** because this will then get highlighted when we do a release. I have updated the release note for this PR.

Thank you for the feedback; I will keep it in mind for future PRs :P

PR #1104, opened 2023-12-19
Fix GKE autoscaling conditions in reconciliation process

kahun:
**What type of PR is this?**
/kind bug

The reconciliation process is broken in GKE. When enableAutoscaling is not defined, the reconciliation process fails because desiredAutoscaling is nil and when it is defined, it fails because the state is unexported.

**TODOs**:
- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Fix GKE autoscaling conditions in reconciliation process.
```


cpanato (commented):
/ok-to-test

cpanato (commented):
/approve

/assign @richardcase 

thanks!

richardcase:
Thanks @kahun 

/lgtm

PR #1134, opened 2024-02-06
Update node pool size even if autoscaling is enabled

kahun:
**What type of PR is this?**
/kind bug

Update node pool size is disabled when autoscaling is enabled but GKE console allows to change node pool size even if autoscaling is enabled.

**TODOs**:
- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Update node pool size even if autoscaling is enabled
```

### DIFF
--- a/cloud/services/container/nodepools/reconcile.go
+++ b/cloud/services/container/nodepools/reconcile.go
@@ -430,12 +430,6 @@ func (s *Service) checkDiffAndPrepareUpdateAutoscaling(existingNodePool *contain
 
 func (s *Service) checkDiffAndPrepareUpdateSize(existingNodePool *containerpb.NodePool) (bool, *containerpb.SetNodePoolSizeRequest) {
 	needUpdate := false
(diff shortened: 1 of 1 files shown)


kahun:
> my question here, is if you use cluster-api to manage your cluster you should not touch the console or change things outside the cluster-api. what impact removing this will cause? is this a current issue?
> 
> cc @richardcase

That's right. The problem now is that if you have autoscaling enabled, you can't increase the size of the node pool with cluster-api.  You'd have to use the console to do that.

And is even worse than that because you actually can change the number of replicas in the machine pool, so the status moves to ScalingUp but the node pool size in gke never gets the update.

richardcase:
@cpanato @kahun - i need to try this out to see the issue. Will do that asap.

richardcase:
@kahun - could we jump on a call to walkthrough this?

kahun:
> @kahun - could we jump on a call to walkthrough this?

Sorry for the delay, whenever you want :+1:

PR #1158, opened 2024-03-04
fix: invalid string for manager's deployment containerPort

damdo:
**What type of PR is this?**
/kind bug

It fixes an invalid containerPort in the manager Deployment's config.
The string was set as an envsubst string but the containerPort field is an int32, and this may break at kustomization/parsing time depending on how the tooling is configured (e.g. for the upstream cluster-api-operator).

**Release note**:
```release-note
fix: invalid string for manager's deployment containerPort
```


cpanato (approved):
thanks

PR #1207, opened 2024-04-11
Allow configuration of API server instance group name

### LINKED ISSUE #1194: Allow customization of API server instance group name
/kind feature

Currently, CAPG creates an instance group named `<infraID>-apiserver-<zone>` for the API server machines. It would be helpful to allow customization of this name to enable migration from systems similar to CAPI onto CAPI itself.

The current value would remain the default, to keep compatibility, and because most users likely won't use this option.

This request comes from [OpenShift adopting Cluster API](https://github.com/openshift/enhancements/pull/1528) and moving away from its legacy Cluster API fork, Machine API. While this feature may impact few users, it's an important enhancement to allow adoption.

OpenShift currently names its API server instance group `<infraID>-master-<zone>`.

Also, this was briefly [discussed on Slack](https://kubernetes.slack.com/archives/C01D1RFEN9G/p1711645409866809?thread_ts=1711556954.886049&cid=C01D1RFEN9G) as a reasonable enhancement.

bfournie:
**What type of PR is this?**

/kind feature

Provides the ability to override the API server instance group name to allow compatibility when using the Openshift Machine API.

Fixes #1194 

Note that a new `LoadBalancerSpec` field has been added to `GCPClusterSpec`. Although this only contains one field it will be populated in the future with additional fields in order to support a LoadBalancer for private clusters.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
A new LoadBalancerSpec field containing the APIServerInstanceGroupTagOverride has been added to GCPClusterSpec.
```


api/v1beta1/gcpcluster_types.go
```diff
@@ -64,6 +64,10 @@ type GCPClusterSpec struct {
 	// supplied then the credentials of the controller will be used.
 	// +optional
 	CredentialsRef *ObjectReference `json:"credentialsRef,omitempty"`
+
+	// LoadBalancerSpec contains configuration for one or more LoadBalancers.
```
barbacbd, line 68: Is this for more than one? It appears to only reference a single Load Balancer.

api/v1beta1/types.go
```diff
@@ -114,6 +114,19 @@ type NetworkSpec struct {
 	LoadBalancerBackendPort *int32 `json:"loadBalancerBackendPort,omitempty"`
 }
 
+// LoadBalancerSpec contains configuration for one or more LoadBalancers.
```
barbacbd, line 117: Same here.

api/v1beta1/gcpcluster_types.go
bfournie: The intent is that `LoadBalancerSpec` will contain all LoadBalancer configuration. Additional fields will be added to this struct including an array to define the scheme (Internal/External) and type (passthrough/proxy) of multiple Load Balancers. This initial `APIServerInstanceGroupTagOverride` applies to the APIServer instance group across any load balancers

api/v1beta1/types.go
bfournie: See above

cloud/services/compute/loadbalancers/reconcile_test.go
```diff
@@ -0,0 +1,167 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
cpanato, line 2:
```suggestion
Copyright 2024 The Kubernetes Authors.
```

cpanato (changes requested):
/ok-to-test

cloud/services/compute/loadbalancers/reconcile_test.go
bfournie: Fixed.

cloud/services/compute/loadbalancers/reconcile_test.go
```diff
@@ -0,0 +1,167 @@
...
+					NamedPorts: []*compute.NamedPort{{Name: "apiserver", Port: 6443}},
+					SelfLink:   "https://www.googleapis.com/compute/v1/projects/proj-id/zones/us-central1-a/instanceGroups/my-cluster-apiserver-us-central1-a",
+				},
+			},
+		},
+		{
+			name: "instanceGroup name is overridden (should create instanceGroup)",
+			scope: func() Scope {
+				tagOverride := "master"
+				clusterScope.GCPCluster.Spec.LoadBalancer = infrav1.LoadBalancerSpec{
```
nrb, line 133: Isn't this changing a variable in a shared scope? If the tests are run in a random order, couldn't this set `clusterScope.GCPCluster.Spec.LoadBalancer` for the rest of the suite?

bfournie: Good catch. Changed to use a separate scope to avoid overwriting.

nrb: Thanks! That'll work.

Another pattern I've seen in tests for other providers is for the function to take in the "golden" one as an argument and return a new copy with the modifications.

Something like  

```go
scope: func(s Scope) Scope {
  s.GCPCluster.Spec.LoadBalancer = infrav1.LoadBalancerSpec{....}
  return s
}
```

and

```go
scope: func(s Scope) Scope {
 return s
}
```

That way, you get to have the top-level one remain clean and passed in on each invocation of `test.scope` to get a fresh, mutable copy.

Up to you if you want to use that here or not.

exp/api/v1beta1/gcpmanagedcluster_types.go
```diff
@@ -59,6 +59,10 @@ type GCPManagedClusterSpec struct {
 	// supplied then the credentials of the controller will be used.
 	// +optional
 	CredentialsRef *infrav1.ObjectReference `json:"credentialsRef,omitempty"`
+
+	// LoadBalancerSpec contains configuration for one or more LoadBalancers.
+	// +optional
+	LoadBalancer infrav1.LoadBalancerSpec `json:"loadBalancer,omitempty"`
```
damdo, line 65: Typo `LoadBalancerSpec` on the godoc vs `LoadBalancer` on the actual field.

cloud/services/compute/loadbalancers/reconcile_test.go
bfournie: Good suggestion. I've reworked it to make the mods only in the test which saves some duplication.

exp/api/v1beta1/gcpmanagedcluster_types.go
bfournie: Thanks for catching that. Fixed.

api/v1beta1/types.go
```diff
@@ -114,6 +114,21 @@ type NetworkSpec struct {
...
+type LoadBalancerSpec struct {
+
+	// APIServerInstanceGroupTagOverride overrides the default setting for the
+	// tag used when creating the API Server Instance Group.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=16
+	// +kubebuilder:validation:Pattern=`(^[1-9][0-9]{0,31}$)|(^[a-z][a-z0-9-]{4,28}[a-z0-9]$)`
+	// +optional
+	APIServerInstanceGroupTagOverride *string `json:"apiServerInstanceGroupTagOverride,omitempty"`
```
patrickdillon, line 127: With the way the API is structured I'm not sure it makes sense for the InstanceGroupTagOverride to belong to the load balancer spec. To me it seems like we should have an instance group spec and define a field in the cluster spec for a a ControlPlaneInstanceGroup. 

I think the issue is, for example, if we add a capability for a secondary load balancer, how do we handle this field potentially being set twice? Instance groups seem more like a dependency of the load balancer.

bfournie: To provide some context on plan for a follow-on PR to enable Internal and External LBs,  the `APIServerInstanceGroupTagOverride` would not actually be set twice. I was going to add a `LoadBalancerConfig`:
```
type LoadBalancingConfig {
 		LBScheme: External|Internal (default=External)
  		LBType: Passthrough|proxy (default=Proxy)
}
```
then LoadBalancerSpec would be
```
type LoadBalancerSpec struct {
	APIServerInstanceGroupTagOverride *string 
    InternalLBConfig *LoadBalancingConfig
    ExternalLBConfig *LoadBalancingConfig
}
```
This keeps the Override with other `APIServerInstanceGroupTagOverride` configuration and allows Internal or External LBs to be created (or maintains current configuration as the External LB will be created if neither config is set)

patrickdillon: I had assumed there would be a separate spec for each LB. Thanks for clarifying and this LGTM

api/v1beta1/types.go
```diff
@@ -114,6 +114,21 @@ type NetworkSpec struct {
 	LoadBalancerBackendPort *int32 `json:"loadBalancerBackendPort,omitempty"`
 }
 
+// LoadBalancerSpec contains configuration for one or more LoadBalancers.
+type LoadBalancerSpec struct {
+
+	// APIServerInstanceGroupTagOverride overrides the default setting for the
+	// tag used when creating the API Server Instance Group.
+	// +kubebuilder:validation:Required
```
patrickdillon, line 122:
```suggestion
	// +kubebuilder:validation:Optional
```

cloud/scope/cluster.go
```diff
@@ -339,8 +344,12 @@ func (s *ClusterScope) HealthCheckSpec() *compute.HealthCheck {
 // InstanceGroupSpec returns google compute instance-group spec.
 func (s *ClusterScope) InstanceGroupSpec(zone string) *compute.InstanceGroup {
 	port := ptr.Deref(s.GCPCluster.Spec.Network.LoadBalancerBackendPort, 6443)
+	tag := infrav1.APIServerRoleTagValue
+	if s.GCPCluster.Spec.LoadBalancer.APIServerInstanceGroupTagOverride != nil {
+		tag = *s.GCPCluster.Spec.LoadBalancer.APIServerInstanceGroupTagOverride
+	}
```
patrickdillon, line 350: nit/suggestion:
```suggestion
	tag := ptr.Deref(s.GCPCluster.Spec.LoadBalancer.APIServerInstanceGroupTagOverride, infrav1.APIServerRoleTagValue)

```

cloud/scope/machine.go
```diff
@@ -128,7 +128,11 @@ func (m *MachineScope) Namespace() string {
 
 // ControlPlaneGroupName returns the control-plane instance group name.
 func (m *MachineScope) ControlPlaneGroupName() string {
-	return fmt.Sprintf("%s-%s-%s", m.ClusterGetter.Name(), infrav1.APIServerRoleTagValue, m.Zone())
+	tag := infrav1.APIServerRoleTagValue
+	if m.ClusterGetter.LoadBalancer().APIServerInstanceGroupTagOverride != nil {
+		tag = *m.ClusterGetter.LoadBalancer().APIServerInstanceGroupTagOverride
+	}
```
patrickdillon, line 134: same nit/suggestion to consider using `ptr.Deref`

api/v1beta1/types.go
```diff
@@ -114,6 +114,21 @@ type NetworkSpec struct {
...
+	// +kubebuilder:validation:MinLength=1
```
nrb, line 123: Honestly not sure how an optional field and a minimum length work together in these generated validations.

api/v1beta1/types.go
bfournie: Good catch, fixed.

cloud/scope/cluster.go
bfournie: Yeah that's a good change, fixed.

api/v1beta1/types.go
bfournie: Removed the MinLength check to avoid any ambiguity.

api/v1beta1/types.go
```diff
@@ -114,6 +114,20 @@ type NetworkSpec struct {
...
+
+	// APIServerInstanceGroupTagOverride overrides the default setting for the
+	// tag used when creating the API Server Instance Group.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=16
+	// +kubebuilder:validation:Pattern=`(^[1-9][0-9]{0,31}$)|(^[a-z][a-z0-9-]{4,28}[a-z0-9]$)`
+	// +optional
+	APIServerInstanceGroupTagOverride *string `json:"apiServerInstanceGroupTagOverride,omitempty"`
+
+	// TODO - additional fields will be added to support Internal LoadBalancer configuration
```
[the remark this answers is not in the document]
bfournie: Yes, removed.

damdo (commented):
/lgtm

Thanks @bfournie !

PR #1222, opened 2024-05-01
Create an Internal Load Balancer if configured

### LINKED ISSUE #1221: Allow for creation of additional load balancer for internal traffic
/kind feature

Currently CAPG creates a External Proxy Load Balancer to load balance traffic to a public IP address. When using CAPI to bootstrap nodes it's necessary to have a second Load Balancer for internal traffic. This request is to add the ability to create a Internal Passthrough Load Balancer.  

The type of load balancers to be created should be specified so that various combinations can be created, i.e. both External and Internal Load Balancers, External only, or Internal only. By default, the External Proxy Load Balancer should only be created as is currently implemented.

bfournie:
**What type of PR is this?**
/kind feature

Provide the ability to configure the types of Load Balancers to be created (Internal and/or External). By default, an External Proxy Load Balancer will be created per the current implementation. If set for an Internal Load Balancer, an Internal Passthrough Load Balancer will be created using resources in the specified region.

**Which issue(s) this PR fixes** :
Fixes #1221

**TODOs**:
- [x ] squashed commits
- [x ] includes documentation
- [ x] adds unit tests

**Release note**:
-->
```release-note
Added a new `LoadBalancerTypes` field to `LoadBalancerSpec` that defines the type of Load Balancers (External or Internal) which should be created. By default a Global External Proxy Load Balancer will be created as is currently implemented. 
Also added an `InternalLoadBalancer` field to optionally configure the name and subnet of the Internal Load Balancer.
```


api/v1beta1/types.go
```diff
@@ -114,6 +134,24 @@ type NetworkSpec struct {
 	LoadBalancerBackendPort *int32 `json:"loadBalancerBackendPort,omitempty"`
 }
 
+// LoadBalancerTypes defines the Load Balancers that should be created.
+type LoadBalancerTypes string
```
barbacbd, line 138:
```suggestion
type LoadBalancerType string
```

api/v1beta1/types.go
```diff
@@ -114,6 +134,24 @@ type NetworkSpec struct {
 	LoadBalancerBackendPort *int32 `json:"loadBalancerBackendPort,omitempty"`
 }
 
+// LoadBalancerTypes defines the Load Balancers that should be created.
```
barbacbd, line 137:
```suggestion
// LoadBalancerType defines the Load Balancer that should be created.
```

cloud/scope/cluster.go
```diff
@@ -290,18 +290,18 @@ func (s *ClusterScope) FirewallRulesSpec() []*compute.Firewall {
...
-		Name:        fmt.Sprintf("%s-%s", s.Name(), infrav1.APIServerRoleTagValue),
+		Name:        fmt.Sprintf("%s-%s", s.Name(), name),
 		AddressType: "EXTERNAL",
 		IpVersion:   "IPV4",
 	}
 }
 
 // BackendServiceSpec returns google compute backend-service spec.
-func (s *ClusterScope) BackendServiceSpec() *compute.BackendService {
+func (s *ClusterScope) BackendServiceSpec(name string) *compute.BackendService {
```
barbacbd, line 302: nit: What name are we intending to pass in here? The backend service name field is combined with the cluster scope name and "name" (maybe something more descriptive for the variable name)?

barbacbd: Note: this applies to all of the altered specs.

api/v1beta1/types.go
```diff
@@ -85,6 +85,26 @@ type Network struct {
 	// created for the API Server.
 	// +optional
 	APIServerForwardingRule *string `json:"apiServerForwardingRule,omitempty"`
+
+	// APIInternalAddress is the IPV4 regional address assigned to the
+	// internal Load Balancer.
+	// +optional
+	APIInternalAddress *string `json:"apiInternalIpAddress,omitempty"`
```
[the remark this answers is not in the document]
bfournie: The main reason I didn't do that was because it is a breaking change, I tried to keep the same API for consistency and just add the additional fields.

If the consensus is that its OK to change this API I can make it an array.

cloud/scope/cluster.go
bfournie: Good point, I'll make that more descriptive.

damdo:
cc. @nrb

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -18,72 +18,228 @@ package loadbalancers
...
 		return err
 	}
+	s.scope.Network().APIServerBackendService = ptr.To[string](backendsvc.SelfLink)
 
+	// Create TargetTCPProxy for Proxy Load Balancer
 	target, err := s.createOrGetTargetTCPProxy(ctx, backendsvc)
 	if err != nil {
 		return err
 	}
+	s.scope.Network().APIServerTargetProxy = ptr.To[string](target.SelfLink)
```
nrb, line 189: Is the target proxy not used at all for the internal load balancer? I could see where we only need one in the case where the external LB is being used as an internal passthrough, but I'm unclear on whether it's required for internal-only.

I'm not sure which line in [this chart](https://cloud.google.com/load-balancing/docs/target-proxies#target_proxy_types) is being implemented, or if that chart even applies.

bfournie: Correct, the target TCP proxy is not required for an internal load balancer which is passthrough only, see https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer
`Internal LoadBalancer Services are implemented by using internal passthrough Network Load Balancers.`

Also https://cloud.google.com/load-balancing/docs/internal

api/v1beta1/types.go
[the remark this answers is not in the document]
cpanato: make sense as we are adding this to add as a struct

wdyt @damdo

jwmay2012:
Thanks for working on this @bfournie. We would really like this as well!

I pulled this down and when running it with `INTERNAL_ONLY` it seems to get stuck on deletion.
It was only deleting global resources like `s.forwardingrules.Delete` but the existing resources were created as `regionalforwardingrules`.

When I added stuff like
```
if err := s.regionalforwardingrules.Delete(ctx, key); err != nil && !gcperrors.IsNotFound(err) {
	log.Error(err, "Error updating a regional forwardingrule", "name", spec.Name)
	return err
}
```
 to `loadbalancers/reconcile.go` functions
 - `deleteForwardingRule()`
 - `deleteBackendService()`
 - `deleteHealthCheck()`
 
 I was then able to reliably get cluster deletion to work without manual intervention or getting stuck.
 
 I'm also noticing the Internal Passthrough LB is getting created with `6443` as the kube api endpoint, but the kubeconfig generated is expecting `443` and never finishes setting up the cluster. I think I understand that passthrough LBs don't allow port mapping so would the solution be to either?--
 - change the `kubeconfig` port from `443` to `6443`
 - change the kube api service and LB from `6443` to `443`

I haven't found the code to change the `kubeconfig` port yet, but I think `ClusterNetwork.APIServerPort` is where I can change the kube api port. 
Not sure what the plans are here, but I'll continue to play round with it.

api/v1beta1/types.go
```diff
@@ -123,6 +161,15 @@ type LoadBalancerSpec struct {
 	// +kubebuilder:validation:Pattern=`(^[1-9][0-9]{0,31}$)|(^[a-z][a-z0-9-]{4,28}[a-z0-9]$)`
 	// +optional
 	APIServerInstanceGroupTagOverride *string `json:"apiServerInstanceGroupTagOverride,omitempty"`
+
+	// LoadBalancerType defines the type of Load Balancer that should be created.
+	// If not set, a Global External Proxy Load Balancer will be created by default.
+	// +optional
+	LoadBalancerType *LoadBalancerType `json:"loadBalancerType,omitempty"`
```
sadasu, line 168: Can the `LoadBalancerType` be defined as `[]LoadBalancerType`. This array can hold `InternalLoadBalancerOnly` and/or `ExternalLoadBalancerOnly`. This also eliminates the need for `DualLoadBalancer` since the array with both Internal and External LoadBalancers would be equivalent to Dual.

bfournie: I had thought about but felt that type of API presents too much ambiguity and potential for error.  For example with that type of array you could have the following potential combinations, all of which are in error:
- `empty` (no LBs defined)
-` INTERNAL/INTERNAL`
- `EXTERNAL/EXTERNAL`

Explicitly defining the Load Balancer configuration options seems more useful.

api/v1beta1/types.go
```diff
@@ -278,3 +325,18 @@ type ObjectReference struct {
...
+	// will be used. For an Internal Load Balancer service the default
+	// name is "api-internal".
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=32
+	// +kubebuilder:validation:Pattern=`(^[1-9][0-9]{0,31}$)|(^[a-z][a-z0-9-]{4,28}[a-z0-9]$)`
+	// +optional
+	Name *string `json:"name,omitempty"`
+
+	// Subnet is the name of the subnet to use for a regional Load Balancer.
+	Subnet *string `json:"subnet,omitempty"`
```
sadasu, line 341: Can the Internal Load Balancer be created without a subnet? Afaik, it can't. So, omitempty is not a valid value.

bfournie: Yes, it must be created with a subnet. Currently it uses the first subnet if one is not defined here, so setting it in the API is optional.
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1222/files#diff-a7274316ed5d60f25cc61e1bcb771a612c3566832a59ecc3a13b279a849c95ebR728-R730

I can change this to require it to be explicitly defined if that seems like the best direction.

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -18,72 +18,230 @@ package loadbalancers
...
+		}
+		if err = s.createInternalLoadBalancer(ctx, name, lbType, instancegroups); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Delete delete cluster control-plane loadbalancer components.
```
sadasu, line 82: Nit: Delete deletes cluster control-plane loadbalancer components.

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -18,72 +18,230 @@ package loadbalancers
...
+func (s *Service) Delete(ctx context.Context) error {
+	lbSpec := s.scope.LoadBalancer()
+	lbType := ptr.Deref(lbSpec.LoadBalancerType, infrav1.ExternalLoadBalancerOnly)
+	if lbType == infrav1.ExternalLoadBalancerOnly || lbType == infrav1.DualLoadBalancer {
+		if err := s.deleteExternalLoadBalancer(ctx); err != nil {
+			return err
```
sadasu, line 88: Should we attempt to also delete Internal Load Balancer when lbType == infrav1.DualLoadBalancer and this delete fails?

api/v1beta1/types.go
sadasu, line 168: I initially thought that maybe this config lends itself well to being a discriminated union with `LoadBalancerType` being discriminated type. But, after reviewing the accompanying changes, I see some advantages to making this field []LoadBalancerType. 

LoadBalancerType(s) can then be populated with ExternalLoadBalancerOnly and/or InternalLoadBalancerOnly thus eliminating the need for `Dual`. This might simplify some of the checks in cloud/services/compute/loadbalancers/reconcile.go.

api/v1beta1/types.go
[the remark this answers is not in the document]
bfournie: OK, how would we handle the API in this case though? Would the existing fields - `APIServerAddress`, `APIServerHealthCheck` etc need to be marked as deprecated but still handled for a few releases until the deprecation is complete? It seems like it will add a lot of complexity to this API.

cloud/services/compute/loadbalancers/reconcile.go
bfournie: > Nit: Delete deletes cluster control-plane loadbalancer components.

Fixed. Thanks.

bfournie:
> Thanks for working on this @bfournie. We would really like this as well!
> 
> I pulled this down and when running it with `INTERNAL_ONLY` it seems to get stuck on deletion. It was only deleting global resources like `s.forwardingrules.Delete` but the existing resources were created as `regionalforwardingrules`.
> 
> When I added stuff like
> 
> ```
> if err := s.regionalforwardingrules.Delete(ctx, key); err != nil && !gcperrors.IsNotFound(err) {
> 	log.Error(err, "Error updating a regional forwardingrule", "name", spec.Name)
> 	return err
> }
> ```
> 
> to `loadbalancers/reconcile.go` functions
> 
> * `deleteForwardingRule()`
> * `deleteBackendService()`
> * `deleteHealthCheck()`
> 
> I was then able to reliably get cluster deletion to work without manual intervention or getting stuck.
> 
> I'm also noticing the Internal Passthrough LB is getting created with `6443` as the kube api endpoint, but the kubeconfig generated is expecting `443` and never finishes setting up the cluster. I think I understand that passthrough LBs don't allow port mapping so would the solution be to either?--
> 
> * change the `kubeconfig` port from `443` to `6443`
> * change the kube api service and LB from `6443` to `443`
> 
> I haven't found the code to change the generated `kubeconfig` port yet, but I think `ClusterNetwork.APIServerPort` is where I can change the kube api port. Not sure what the plans are here, but I'll continue to play round with it.

Thank you for testing this! I wasn't seeing an error on deletion but agree that the regional interfaces were not being explicitly called, so I've changed this to do explicit deletions on the regional resources.

Regarding the port being used, the ControlPlaneEndpoint should be set to 443 if not overridden by APIServerPort https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/cloud/scope/cluster.go#L148-L154 and this *should* be used in the Internal only case here https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1222/files#diff-a7274316ed5d60f25cc61e1bcb771a612c3566832a59ecc3a13b279a849c95ebR235-R240 but perhaps its not. I will investigate this.

api/v1beta1/types.go
sadasu: I think you can achieve those via kubebuilder validations :
 "// +kubebuilder:validation:UniqueItems=" to make sure array entries are unique
 and "// +kubebuilder:default:="External" will make sure it would not be empty.
 
 I know I had suggested making LoadBalancers of type []LoadBalancerType, with the array containing `ExternalLoadBalancerOnly` and/or `InternalLoadBalancerOnly` but looking at the rest of the implementation, I am not sure that is the right path.

sadasu: Are the 4 optional fields added to Network type allowed to be empty when LoadBalancerType is InternalLoadBalancerOnly/DualLoadBalancer ?

sadasu: Also, can the value of LoadBalancerType change once set i.e is it immutable? If so, do we need a validateUpdate() webhook preventing an update for LoadBalancerType?

api/v1beta1/types.go
```diff
@@ -123,6 +161,15 @@ type LoadBalancerSpec struct {
...
 	APIServerInstanceGroupTagOverride *string `json:"apiServerInstanceGroupTagOverride,omitempty"`
+
+	// LoadBalancerType defines the type of Load Balancer that should be created.
+	// If not set, a Global External Proxy Load Balancer will be created by default.
+	// +optional
+	LoadBalancerType *LoadBalancerType `json:"loadBalancerType,omitempty"`
+
+	// InternalLoadBalancer is the configuration for an Internal Passthrough Network Load Balancer.
+	// +optional
+	InternalLoadBalancer *LoadBalancer `json:"internalLoadBalancer,omitempty"`
```
sadasu, line 172: Similar question about this field being immutable.

api/v1beta1/types.go
bfournie: > Are the 4 optional fields added to Network type allowed to be empty when LoadBalancerType is InternalLoadBalancerOnly/DualLoadBalancer ?

Yes, these marked as `+optional` and are string pointers so its OK to not be set

bfournie: > Also, can the value of LoadBalancerType change once set i.e is it immutable? If so, do we need a validateUpdate() webhook preventing an update for LoadBalancerType?

That's a good point, thanks for catching that. I will add a webhook check for it

api/v1beta1/types.go
bfournie: > Similar question about this field being immutable.

I've added a webhook validator on the LoadBalancerSpec struct which will detect changes to any f the fields.

api/v1beta1/types.go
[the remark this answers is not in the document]
damdo: I tend to agree with @bfournie here.
I'd probably stick with a non-breaking change, and then schedule a full API re-review for a new v1beta2 where we can improve the situation and plan the breaking changes we want to make to the API to clean it up.

api/v1beta1/types.go
```diff
@@ -278,3 +325,18 @@ type ObjectReference struct {
...
 	Name string `json:"name"`
 }
+
+// LoadBalancer specifies the configuration of a LoadBalancer.
+type LoadBalancer struct {
+	// Name is the name of the Load Balancer. If not set a default name
+	// will be used. For an Internal Load Balancer service the default
+	// name is "api-internal".
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=32
```
damdo, line 335: Is this a limitation on the AWS side?

api/v1beta1/types.go
damdo: > Yes, it must be created with a subnet. Currently it uses the first subnet if one is not defined here, so setting it in the API is optional.
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1222/files#diff-a7274316ed5d60f25cc61e1bcb771a612c3566832a59ecc3a13b279a849c95ebR728-R730

I'd explicitly write this in the comment for this field.

api/v1beta1/types.go
```diff
@@ -85,6 +85,26 @@ type Network struct {
...
+
+	// APIInternalBackendService is the full reference to the backend service
+	// created for the internal Load Balancer.
+	// +optional
+	APIInternalBackendService *string `json:"apiInternalBackendService,omitempty"`
+
+	// APIInternalForwardingRule is the full reference to the forwarding rule
+	// created for the internal Load Balancer.
+	// +optional
+	APIInternalForwardingRule *string `json:"apiInternalForwardingRule,omitempty"`
```
damdo, line 107: Are these fields all required when the Internal Load Balancer is set?
If not are they defaulted to something in code?

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -18,72 +18,240 @@ package loadbalancers
...
+	// utilization of instances.
+	loadBalancingModeUtilization = loadBalancingMode("UTILIZATION")
+
+	// Connection determines how the traffic load is spread based on the
+	// total number of connections that a backend can handle. This is
+	// only mode available for passthrough Load Balancers.
+	loadBalancingModeConnection = loadBalancingMode("CONNECTION")
+
+	loadBalanceTrafficInternal = "INTERNAL"
+)
```
damdo, line 46: Also here let's adhere to the PascalCase conventions

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -18,72 +18,240 @@ package loadbalancers
...
+	}
+	if internalLBErr != nil {
+		if err := s.deleteInstanceGroups(ctx); err != nil {
+			log.Error(err, "Error deleting instancegroup")
+		}
+		return internalLBErr
+	}
+
+	// Delete shared instancegroups last
+	return s.deleteInstanceGroups(ctx)
```
damdo, line 113: Instead of doing this what do you think if we attempted deleting the instance groups no matter the externalLBErr/internalLBErr, and then aggregate all the errors and return them?

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -18,72 +18,240 @@ package loadbalancers
...
+
+	if err := s.deleteBackendService(ctx, name); err != nil {
+		return err
+	}
+	s.scope.Network().APIServerBackendService = nil
+
+	if err := s.deleteHealthCheck(ctx, name); err != nil {
+		return err
+	}
+	s.scope.Network().APIServerHealthCheck = nil
```
damdo, line 143: I'd add an `fmt.Errorf(` wrap for all these errors for easier debugging when following the log error chain

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -18,72 +18,240 @@ package loadbalancers
...
+
+	if err := s.deleteRegionalBackendService(ctx, name); err != nil {
+		return err
+	}
+	s.scope.Network().APIInternalBackendService = nil
+
+	if err := s.deleteRegionalHealthCheck(ctx, name); err != nil {
+		return err
+	}
+	s.scope.Network().APIInternalHealthCheck = nil
```
damdo, line 169: Error wrapping here too

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -130,60 +298,94 @@ func (s *Service) createOrGetInstanceGroups(ctx context.Context) ([]*compute.Ins
...
 	for _, group := range instancegroups {
-		backends = append(backends, &compute.Backend{
-			BalancingMode: "UTILIZATION",
+		be := &compute.Backend{
+			BalancingMode: string(mode),
 			Group:         group.SelfLink,
-		})
+		}
+		if mode == loadBalancingModeConnection {
+			be.MaxConnections = int64(2 ^ 32)
```
damdo, line 365: Can we add a comment specifying where this value is coming from?

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -192,127 +394,283 @@ func (s *Service) createOrGetBackendService(ctx context.Context, instancegroups
...
+
+// createOrGetRegionalForwardingRule is used to obtain a Regional ForwardingRule.
+func (s *Service) createOrGetRegionalForwardingRule(ctx context.Context, lbname string, backendSvc *compute.BackendService, addr *compute.Address) (*compute.ForwardingRule, error) {
+	log := log.FromContext(ctx)
+	spec := s.scope.ForwardingRuleSpec(lbname)
+	spec.LoadBalancingScheme = string(loadBalanceTrafficInternal)
+	spec.Region = s.scope.Region()
+	spec.BackendService = backendSvc.SelfLink
+	// Ports are used instead or PortRange for passthrough Load Balancer
+	spec.Ports = []string{"6443", "22623"}
```
damdo, line 593: Where are these values coming from?

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -376,3 +757,32 @@ func (s *Service) deleteInstanceGroups(ctx context.Context) error {
...
+			region = s.scope.Region()
+		}
+
+		subnetKey := meta.RegionalKey(subnetSpec.Name, region)
+		subnet, err := s.subnets.Get(ctx, subnetKey)
+		if err != nil {
+			return nil, err
+		}
+		// Return subnet that matches configuration or first one if not configured
+		if cfgSubnet == "" || strings.HasSuffix(subnet.Name, cfgSubnet) {
```
damdo, line 782: What is the reason for using HasSuffix instead of a full match?

damdo (commented):
Thanks @bfournie, great effort with this feature!
I left my review, most are nitpicks and some questions.

Also were you planning on writing E2E tests for this?
I think this is a big enough feature that could warrant that.

api/v1beta1/types.go
bfournie: Updated comment.

bfournie:
> Also were you planning on writing E2E tests for this? I think this is a big enough feature that could warrant that.

Good point, I'll look into writing an E2E test.

api/v1beta1/types.go
```diff
@@ -114,6 +134,24 @@ type NetworkSpec struct {
...
+	// Balancer and will be created if no LoadBalancerType is defined.
+	ExternalLoadBalancerOnly = LoadBalancerType("EXTERNAL_ONLY")
+
+	// InternalLoadBalancerOnly creates a Regional Internal Passthrough Load
+	// Balancer to manage traffic to backends in the configured region.
+	InternalLoadBalancerOnly = LoadBalancerType("INTERNAL_ONLY")
+
+	// DualLoadBalancer creates both External and Internal Load Balancers to provide
+	// separate endpoints for managing both external and internal traffic.
+	DualLoadBalancer = LoadBalancerType("Dual")
```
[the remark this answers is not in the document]
bfournie: Thanks for the comment, changed to use suggested names

api/v1beta1/types.go
bfournie: Similar to the fields for the External Load Balancer (APIServer fields) these will be set to the resources created for the Internal Load Balancer when it is created, including the cluster name.

cloud/services/compute/loadbalancers/reconcile.go
bfournie: These values - `CONNECTION`, `INTERNAL` etc. must match the value that gcp expects. It appears that this is all upper case, from https://cloud.google.com/load-balancing/docs/backend-service#connection_balancing_mode

cloud/services/compute/loadbalancers/reconcile.go
bfournie: Done.

cloud/services/compute/loadbalancers/reconcile.go
bfournie: Good point, this came from an internal implementation but I'll update it with a clarifying comment. TBD.

cloud/services/compute/loadbalancers/reconcile.go
bfournie: Added comment.

cloud/services/compute/loadbalancers/reconcile.go
bfournie: The full name includes the cluster name so it just matches the subnet portion. I can update this to build the name including cluster and subnet for comparision.

bfournie:
> > Also were you planning on writing E2E tests for this? I think this is a big enough feature that could warrant that.
> 
> Good point, I'll look into writing an E2E test.

As we discussed separately, I will create another PR for the e2e test for this functionality as this PR is already quite large. I'll reference this PR in it.

api/v1beta1/types.go
damdo: Got it thanks

cloud/services/compute/loadbalancers/reconcile.go
damdo: Sounds good

damdo (commented):
Thank you for addressing my feedback @bfournie 
I see you are working on a separate PR for e2e testing. SGTM

/lgtm

From my side

/hold

Until at least @cpanato is happy.
/assign @cpanato

cpanato:
i am always happy :)

jsut a question, this will not break any existing clusters that eventually upgrade to this version, correct?

bfournie:
> i am always happy :)
> 
> just a question; this will not break any existing clusters that eventually upgrade to this version, correct?

Correct, it will not break existing clusters on upgrade which will not have LoadBalancerType set so only an External LB will created which is the current behavior, see https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1222/files#diff-a7274316ed5d60f25cc61e1bcb771a612c3566832a59ecc3a13b279a849c95ebR61

cpanato:
/approve
/lgtm

thanks!

cpanato:
will wait for the e2e tests :)

bfournie:
> will wait for the e2e tests :)

Thanks @cpanato! I created a separate PR for the e2e tests with some questions https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1245

Can we merge this first and then focus on the e2e test in https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1245 ?

bfournie:
> /unhold

Thank you

jwmay2012:
I'm having an issue with the `Internal` only loadbalancer type. When cluster-api brings everything up, the LB never passes the healthcheck.
I can get the LB to be healthy by:
- deleting the "frontend/Forwardingrule" and allowing CAPG to recreate it. immediately healthy when it is recreated.
- OR manually creating a second internal passthrough LB, identically configured (or slightly different, but 'correct') to the one created by CAPG. Both LB's become healthy. After deleting the new LB, the original stays healthy.

Seems super weird. Creating a new identical frontend/forwardingrule or recreating the old one causes the healthcheck to start passing and the LB to start accepting requests.

This is for a 1 node cluster.

If there's a better place to post this, I can do that.

bfournie:
> I'm having an issue with the `Internal` only loadbalancer type. When cluster-api brings everything up, the LB never passes the healthcheck. I can get the LB to be healthy by:
> 
> * deleting the "frontend/Forwardingrule" and allowing CAPG to recreate it. immediately healthy when it is recreated.
> * OR manually creating a second internal passthrough LB, identically configured (or slightly different, but 'correct') to the one created by CAPG. Both LB's become healthy. After deleting the new LB, the original stays healthy.
> 
> Seems super weird. Creating a new identical frontend/forwardingrule or recreating the old one causes the healthcheck to start passing and the LB to start accepting requests.
> 
> This is for a 1 node cluster.
> 
> If there's a better place to post this, I can do that.

So this was with just the Internal LB or both External and Internal LB?
We have a test now for both External and Internal. I haven't seen that issue with the 
healthcheck not passing. For Internal only most of my testing has been using this as a provider of Openshift, I haven't seen this in particular but will investigate it.

Probably best to create a Issue under https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues so we can track it.

jwmay2012:
Internal Only LB. No external. I'll work to create an issue soon.

PR #1252, opened 2024-06-03
Delete load balancers before subnets

bfournie:
**What type of PR is this?**
/kind bug

When using an internal load balancer, the internal IP addresses uses a cluster subnet. To avoid "already being used" errors on the subnet deletion, delete the load balancer first.

This is a follow-on from https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1222.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/gcpcluster_controller.go
+++ b/controllers/gcpcluster_controller.go
@@ -231,8 +231,8 @@ func (r *GCPClusterReconciler) reconcileDelete(ctx context.Context, clusterScope
 	log.Info("Reconciling Delete GCPCluster")
 
 	reconcilers := []cloud.Reconciler{
-		subnets.New(clusterScope),
 		loadbalancers.New(clusterScope),
+		subnets.New(clusterScope),
 		firewalls.New(clusterScope),
 		networks.New(clusterScope),
 	}


controllers/gcpcluster_controller.go
```diff
@@ -231,8 +231,8 @@ func (r *GCPClusterReconciler) reconcileDelete(ctx context.Context, clusterScope
 	log.Info("Reconciling Delete GCPCluster")
 
 	reconcilers := []cloud.Reconciler{
-		subnets.New(clusterScope),
 		loadbalancers.New(clusterScope),
+		subnets.New(clusterScope),
 		firewalls.New(clusterScope),
```
cpanato, line 236: should we delete the firewalls before as well?

cpanato (approved):

controllers/gcpcluster_controller.go
bfournie: I don't think that's necessary as there are no conflicting resources being used by the firewalls.

PR #1253, opened 2024-06-04
[release-1.6] Bump setup-envtest to use new controller-tools binaries

damdo:
Bump setup-envtest to use new controller-tools binaries

 
**What type of PR is this?**

/kind cleanup

**Release note**:
```release-note
NONE
```


cpanato (approved):
thanks

PR #1259, opened 2024-06-12
fix: Make gcpmanagedcluster.network a required field

### LINKED ISSUE #1187: Nil pointer dereference, on an invalid spec
/kind bug

[A clear and concise description of what the bug is.]
I created a GCPManagedCluster without the network field and the controller crashed.

The controller shouldn't crash

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/cloud/services/container/clusters/reconcile.go#L256

**Environment:**

- Cluster-api version:
- Minikube/KIND version:

tasdikrahman:
**What type of PR is this?**

This makes the `gcpmanagedcluster.network` field a required field. 
 
/kind feature
 

Fixes #1187 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
fix: Makes GCPManagedCluster Spec's network field a required field
```

### DIFF
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmanagedclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmanagedclusters.yaml
@@ -293,6 +293,7 @@ spec:
                   type: object
                 type: array
             required:
+            - network
             - project
             - region
             type: object
--- a/exp/api/v1beta1/gcpmanagedcluster_types.go
+++ b/exp/api/v1beta1/gcpmanagedcluster_types.go
@@ -41,7 +41,6 @@ type GCPManagedClusterSpec struct {
 	ControlPlaneEndpoint clusterv1.APIEndpoint `json:"controlPlaneEndpoint"`
 
 	// NetworkSpec encapsulates all things related to the GCP network.
-	// +optional
 	Network infrav1.NetworkSpec `json:"network"`
 
 	// AdditionalLabels is an optional set of tags to add to GCP resources managed by the GCP provider, in addition to the


cpanato (commented):
/ok-to-test

damdo:
Would this be considered a breaking API change?

tasdikrahman:
Hey folks. 

@damdo @cpanato thanks for the review. 

> Would this be considered a breaking API change?

I feel so, since we are making it required here in this PR. Let me know if you would like me to change something. Thanks.

But thinking out loud, if the controller would crash when this attribute was not passed, the controller would be implicitly requiring this value from what I can understand.

damdo:
@tasdikrahman 
#1187 is highlighting that this missing causes a nil pointer exception, which makes the controller crash.
I think the right way of doing this would be to first fix that nil pointer by instead throwing an reconciler error instead of assuming Network is always there. (1)

Then this would alert users that the cluster can't be created because the `Network` field is missing, and so at least we would give out some informative detail on how they can fix this.

Then yeah by all means, we can set that to not optional, but it would be an API breaking change, which we might want to think through whether we want that to be happening in v1beta1 or that requires a v1beta2.

Would you be able to open a PR to do (1) for now? Thanks!

tasdikrahman:
@damdo That makes sense, thanks for your comments, I have tried addressing the nil pointer issue here in this PR https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1261/ to begin with.

tasdikrahman:
Hello @damdo , I opened up an issue here https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/1285 to track this change of the Network field being made a required one.

PR #1262, opened 2024-06-14
feat: Add ability to specify MTU when creating Network

### LINKED ISSUE #1246: Support MTU change for VPC
/kind feature

[A clear and concise description of what you want to happen.]

As of now, CAPG doesn't allow to set MTU for VPC which creates a VPC with default MTU as 1460. So option needs to be available to change MTU to different value. 

tasdikrahman:
**What type of PR is this?**

/kind feature

User can set `GCPCluster.Spec.Network.Mtu` for setting the MTU for the network being created. If nothing is passed the MTU will default to 1440 which is the default set here https://cloud.google.com/vpc/docs/mtu

Fixes #1246

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
feat: Add ability to add GCPCluster.Spec.Network.Mtu 
```

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -137,9 +137,13 @@ type NetworkSpec struct {
 	// +optional
 	HostProject *string `json:"hostProject,omitempty"`
 
-	// Mtu: Maximum Transmission Unit (MTU), in bytes, of packets passing through
-	// this interconnect attachment. Only 1440 and 1500 are allowed. If not
-	// specified, the value will default to 1440.
+	// Mtu: Maximum Transmission Unit in bytes. The minimum value for this field is
+	// 1300 and the maximum value is 8896. The suggested value is 1500, which is
+	// the default MTU used on the Internet, or 8896 if you want to use Jumbo
+	// frames. If unspecified, the value defaults to 1460.
+	// More info: https://pkg.go.dev/google.golang.org/api/compute/v1#Network
+	// +kubebuilder:validation:Minimum:=1300
+	// +kubebuilder:validation:Maximum:=8896
 	// +optional
 	Mtu int64 `json:"mtu,omitempty"`
 }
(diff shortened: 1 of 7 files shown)


cpanato (commented):
/ok-to-test

tasdikrahman:
Thanks for the review @cpanato , let me know if you would like me to add any more changes on this PR before approval. Thank you.

cpanato:
thanks

/approve

/assign @damdo

api/v1beta1/types.go
```diff
@@ -136,6 +136,12 @@ type NetworkSpec struct {
 	// HostProject is the name of the project hosting the shared VPC network resources.
 	// +optional
 	HostProject *string `json:"hostProject,omitempty"`
+
+	// Mtu: Maximum Transmission Unit (MTU), in bytes, of packets passing through
+	// this interconnect attachment. Only 1440 and 1500 are allowed. If not
+	// specified, the value will default to 1440.
+	// +optional
+	Mtu int64 `json:"mtu,omitempty"`
```
damdo, line 144: I think we should add openAPI validation for this field only allowing the two values you are mentioning.
Also where are those values coming from? We might want to leave a message mentioning that.

damdo: Also we could add a validation step in the validation code (e.g. webooks).
And a couple of unit test checking that is validated correctly.

damdo (changes requested):

damdo:
/hold
For changes to be added

api/v1beta1/types.go
tasdikrahman: > I think we should add openAPI validation for this field only allowing the two values you are mentioning.
Also where are those values coming from? We might want to leave a message mentioning that.

Hello there, I fixed the validation code to have the minimum and maximum values present. The values of minimum of 1300 and maximum of 8896 are taken from https://pkg.go.dev/google.golang.org/api/compute/v1#Network

> Also we could add a validation step in the validation code (e.g. webooks).
And a couple of unit test checking that is validated correctly.

Will add some unit tests exercising this.

damdo:
Thank you for implementing the requested changes @tasdikrahman 
/lgtm
/unhold

PR #1275, opened 2024-06-27
Set Labels on Disks and ForwardingRules resources

### LINKED ISSUE #1274: AdditionalLabels are not added to all GCP resources
/kind bug

Both the `GCPClusterSpec` and `GCPMachineSpec` have AdditionalLabels fields to add user-defined labels to the resources

e.g.
https://github.com/bfournie/cluster-api-provider-gcp/blob/main/api/v1beta1/gcpcluster_types.go#L55
 ```
      // AdditionalLabels is an optional set of tags to add to GCP resources managed by the GCP provider, in addition to the
        // ones added by default.
        // +optional
        AdditionalLabels Labels `json:"additionalLabels,omitempty"
```

We can see these labels being added to an instance here:
https://github.com/bfournie/cluster-api-provider-gcp/blob/main/cloud/scope/machine.go#L398

However other resources also take the Labels and this field is not being set, for example Disks, ForwardingRules, Addresses etc. (see https://raw.githubusercontent.com/googleapis/google-api-go-client/main/compute/v1/compute-gen.go)

The Labels field should be populated from AdditionalLabels for resources that have this field.

Doing a openshift install using the capg-provider with the `AdditionalLabels` input set in the GCPCluster

Results in the instances getting the label
```
$ gcloud compute instances describe <instance-name>
...
labels:
  foo: bar
```

But no other resources get this label.

**Environment:**

Openshift 4.16 using capg provider.

### PR DESCRIPTION
**Pull request by bfournie**
**What type of PR is this?**

/kind bug

The Labels field should be set on Disks and ForwardingRules resources from the machine or cluster `AdditionalLabels` field, similar to how it is set on Instances.

Fixes #1274 

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
This fix populates the Disk and ForwardingRules resources' `Labels` field from the machine/cluster `AdditionalLabels` field.
```

### DIFF
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -337,6 +337,7 @@ func (s *ClusterScope) ForwardingRuleSpec(lbname string) *compute.ForwardingRule
 		IPProtocol:          "TCP",
 		LoadBalancingScheme: "EXTERNAL",
 		PortRange:           portRange,
+		Labels:              s.AdditionalLabels(),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


cpanato (commented):
/ok-to-test

cpanato (approved):
thanks

PR #1277, opened 2024-06-28
Use SetLabels() to set ForwardingRules labels

### LINKED ISSUE #1276: ForwardingRules' labels must be added via the SetLabels() method
/kind bug

In https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1275, the ForwardingRules `Labels`  field was set in the resource definition. It turns out that some resources, including ForwardingRules, require the` SetLabels() ` method to add the labels. This is the comment for ForwardingRules in https://github.com/googleapis/google-api-go-client/blame/main/compute/v1/compute-gen.go
```
       // Labels: Labels for this resource. These can only be added or modified
        // by the setLabels method. 
        Labels map[string]string `json:"labels,omitempty"
```

This is different than other resources, for example Instances and Disks, in which the Label can be set in the resource and later modified using `SetLabels()`. This is the description in compute-gen.go
```
        // Labels: Labels to apply to this instance. These can be later modified
        // by the setLabels method.
        Labels map[string]string `json:"labels,omitempty"`
```

So, the end result is that the change to add Labels to the resources when creating it works fine for Disks but does not work for ForwardingRules which require `SetLabels() `to be used after the resource is create.

Will need to change ForwardingRules to use `SetLabels() `.

**Environment:**
Openshift 4.16 using cluster-api-provider-gcp as provider.

bfournie:
**What type of PR is this?**
/kind bug

The ForwardingRules resource requires that the SetLabels() method be used to add the labels after the resource has been created. This resource is different than, for example, Instances or Disks. See Issue #1276 for more information.

**Which issue(s) this PR fixes** 
Fixes #1276 

This is the comment for `ForwardingRules` in https://github.com/googleapis/google-api-go-client/blame/main/compute/v1/compute-gen.go
```
       // Labels: Labels for this resource. These can only be added or modified
       // by the setLabels method. 
        Labels map[string]string `json:"labels,omitempty"
```

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
"NONE"
```

### DIFF
--- a/cloud/services/compute/loadbalancers/reconcile.go
+++ b/cloud/services/compute/loadbalancers/reconcile.go
@@ -579,6 +579,18 @@ func (s *Service) createOrGetForwardingRule(ctx context.Context, lbname string, 
 		}
 	}
 
+	// Labels on ForwardingRules must be added after resource is created
+	labels := s.scope.AdditionalLabels()
+	if !labels.Equals(forwarding.Labels) {
+		setLabelsRequest := &compute.GlobalSetLabelsRequest{
+			LabelFingerprint: forwarding.LabelFingerprint,
+			Labels:           labels,
+		}
+		if err = s.forwardingrules.SetLabels(ctx, key, setLabelsRequest); err != nil {
+			return nil, err
+		}
+	}
+
 	return forwarding, nil
 }
 
@@ -618,6 +630,18 @@ func (s *Service) createOrGetRegionalForwardingRule(ctx context.Context, lbname 
 
 		forwarding, err = s.regionalforwardingrules.Get(ctx, key)
 		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Labels on ForwardingRules must be added after resource is created
+	labels := s.scope.AdditionalLabels()
+	if !labels.Equals(forwarding.Labels) {
+		setLabelsRequest := &compute.RegionSetLabelsRequest{
+			LabelFingerprint: forwarding.LabelFingerprint,
+			Labels:           labels,
+		}
+		if err = s.regionalforwardingrules.SetLabels(ctx, key, setLabelsRequest); err != nil {
 			return nil, err
 		}
 	}
--- a/cloud/services/compute/loadbalancers/service.go
+++ b/cloud/services/compute/loadbalancers/service.go
@@ -44,6 +44,14 @@ type forwardingrulesInterface interface {
 	Get(ctx context.Context, key *meta.Key, options ...k8scloud.Option) (*compute.ForwardingRule, error)
 	Insert(ctx context.Context, key *meta.Key, obj *compute.ForwardingRule, options ...k8scloud.Option) error
 	Delete(ctx context.Context, key *meta.Key, options ...k8scloud.Option) error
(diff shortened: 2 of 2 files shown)


cpanato (commented):
/ok-to-test

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -579,6 +580,19 @@ func (s *Service) createOrGetForwardingRule(ctx context.Context, lbname string,
...
+	labels := s.scope.AdditionalLabels()
+	if !reflect.DeepEqual(labels, forwarding.Labels) {
+		setLabelsRequest := &compute.GlobalSetLabelsRequest{
+			LabelFingerprint: forwarding.LabelFingerprint,
+			Labels:           labels,
+		}
+		err = s.forwardingrules.SetLabels(ctx, key, setLabelsRequest)
+		if err != nil {
+			return nil, err
+		}
```
damdo, line 593: How about
```suggestion
		if err := s.forwardingrules.SetLabels(ctx, key, setLabelsRequest); err != nil {
			return nil, err
		}
```

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -579,6 +580,19 @@ func (s *Service) createOrGetForwardingRule(ctx context.Context, lbname string,
 		}
 	}
 
+	// Labels on ForwardingRules must be added after resource is created
+	labels := s.scope.AdditionalLabels()
+	if !reflect.DeepEqual(labels, forwarding.Labels) {
```
damdo, line 585: Just to confirm, this is a map right? So ordering should not be an issue while diffing correct?

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -622,6 +636,19 @@ func (s *Service) createOrGetRegionalForwardingRule(ctx context.Context, lbname
...
+		setLabelsRequest := &compute.RegionSetLabelsRequest{
+			LabelFingerprint: forwarding.LabelFingerprint,
+			Labels:           labels,
+		}
+		err = s.regionalforwardingrules.SetLabels(ctx, key, setLabelsRequest)
+		if err != nil {
+			return nil, err
+		}
```
damdo, line 649: Same here :)

cloud/services/compute/loadbalancers/reconcile.go
bfournie: Yes it is a map. There's actually a predefined function https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/api/v1beta1/labels.go#L29-L31 that I changed to use.

cloud/services/compute/loadbalancers/reconcile.go
bfournie: Fixed.

cloud/services/compute/loadbalancers/reconcile.go
bfournie: Thanks, fixed.

damdo (commented):
/lgtm

/assign @cpanato

PR #1313, opened 2024-09-18
Add support for Spot VMs

jwmay2012:
**What type of PR is this?**
/kind feature

Adds support for GCP Spot VMs, which are distinct from `preemptible` vms. Spot VMs are newer.
https://cloud.google.com/compute/docs/instances/spot
They no longer have a mandatory 24 hr limit.

Modeled PR after the [addition of preemptible vms](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/300). 

Documentation that specifies details on `provisioningModel`, `SPOT` and `STANDARD`--
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#provisioning_model
https://cloud.google.com/compute/docs/instances/create-use-spot#create

Relevant docs about field logic--
```
"provisioningModel": "SPOT" in the Compute Engine API

---------------

If the output includes the provisioningModel field set to SPOT, the VM is a Spot VM.

Otherwise, if the output includes the provisioningModel field set to standard or if the output omits the provisioningModel field:
- If the output includes the preemptible field set to true, the VM is a preemptible VM.
- Otherwise, the VM is a standard VM.
```

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [ ] adds unit tests
No tests exist for preemptible vms. This change seems pretty simple and straight forward and doesn't cause any changes unless the new field is used.

**Release note**:
```release-note
Add support for Spot VMs. `GCPMachineTemplate.Spec.ProvisioningModel = "Spot"`
```

### DIFF
--- a/api/v1beta1/gcpmachine_types.go
+++ b/api/v1beta1/gcpmachine_types.go
@@ -217,6 +217,16 @@ type CustomerEncryptionKey struct {
 	SuppliedKey *SuppliedKey `json:"suppliedKey,omitempty"`
 }
 
+// ProvisioningModel is a type for Spot VM enablement.
+type ProvisioningModel string
+
+const (
+	// ProvisioningModelStandard specifies the VM type to NOT be Spot.
+	ProvisioningModelStandard ProvisioningModel = "Standard"
+	// ProvisioningModelSpot specifies the VM type to be Spot.
+	ProvisioningModelSpot ProvisioningModel = "Spot"
+)
+
 // GCPMachineSpec defines the desired state of GCPMachine.
 type GCPMachineSpec struct {
 	// InstanceType is the type of instance to create. Example: n1.standard-2
@@ -301,6 +311,13 @@ type GCPMachineSpec struct {
 	// Preemptible defines if instance is preemptible
 	// +optional
 	Preemptible bool `json:"preemptible,omitempty"`
+
+	// ProvisioningModel defines if instance is spot.
+	// If set to "Standard" while preemptible is true, then the VM will be of type "Preemptible".
+	// If "Spot", VM type is "Spot". When unspecified, defaults to "Standard".
+	// +kubebuilder:validation:Enum=Standard;Spot
+	// +optional
+	ProvisioningModel *ProvisioningModel `json:"provisioningModel,omitempty"`
 
 	// IPForwarding Allows this instance to send and receive packets with non-matching destination or source IPs.
 	// This is required if you plan to use this instance to forward routes. Defaults to enabled.
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -402,6 +402,16 @@ func (m *MachineScope) InstanceSpec(log logr.Logger) *compute.Instance {
 			Preemptible: m.GCPMachine.Spec.Preemptible,
 		},
 	}
+	if m.GCPMachine.Spec.ProvisioningModel != nil {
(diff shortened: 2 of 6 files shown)


cloud/scope/machine.go
```diff
@@ -402,6 +402,16 @@ func (m *MachineScope) InstanceSpec(log logr.Logger) *compute.Instance {
 			Preemptible: m.GCPMachine.Spec.Preemptible,
 		},
 	}
+	if m.GCPMachine.Spec.ProvisioningModel != nil {
```
salasberryfin, line 405: Should we also explicitly set the `ProvisioningModel` to `Standard` when not null or is it implicitly set by GCP?

salasberryfin:
The CAPG book contains a [section on using Preemptible VMs](https://cluster-api-gcp.sigs.k8s.io/topics/preemptible-vms) (book source files can be found [here](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/tree/main/docs/book/src)). Perhaps it makes sense to also document Spot VMs and what happens with each combination of `Preemptible` + `ProvisioningModel`, as you specified on type definition. What do you think?

cloud/scope/machine.go
jwmay2012: I do believe GCP defaults to `STANDARD`. All documentation seems to be focused around specifying `SPOT` and otherwise ignoring the field. 

I also figured it was lower risk to maintain the existing behavior when the field is not set, allowing the GCP default to continue to be used.

jwmay2012:
> The CAPG book contains a [section on using Preemptible VMs](https://cluster-api-gcp.sigs.k8s.io/topics/preemptible-vms) (book source files can be found [here](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/tree/main/docs/book/src)). Perhaps it makes sense to also document Spot VMs and what happens with each combination of `Preemptible` + `ProvisioningModel`, as you specified on type definition. What do you think?

Makes sense. I'll add the specifics of `Preemptible` + `ProvisioningModel` to the docs I've already made. I believe `Preemptible` gets ignored and `ProvisioiningModel` overrides. This is implicit in the documentation, but I'll confirm first.

jwmay2012:
Tested combinations of `preemptible: true` and `provisioningModel: Spot`. Spot takes priority.

api/v1beta1/gcpmachine_types.go
```diff
@@ -302,6 +312,13 @@ type GCPMachineSpec struct {
 	// +optional
 	Preemptible bool `json:"preemptible,omitempty"`
 
+	// ProvisioningModel defines if instance is spot.
+	// If set to "Standard" while preemptible is true, then the VM will be of type "Preemptible".
+	// If "Spot", VM type is "Spot". When unspecified, defaults to "Standard".
+	// +kubebuilder:validation:Enum=Standard;Spot
+	// +optional
```
barbacbd, line 319: nit: Technically this is fine since it is alphabetical but if any other instance types come up in the future this could change. 

```suggestion
        // +kubebuilder:default=Standard
	// +optional
```

salasberryfin (approved):
Thanks for working on this @jwmay2012

cpanato (approved):
/lgtm

/hold for another eye from @richardcase 
/assing @richardcase

richardcase:
Also from my side:

/lgtm

And so lets

/unhold

PR #1315, opened 2024-09-19
Add support for InstanceTerminationAction for instance preemption.

jwmay2012:
**What type of PR is this?**
/kind feature

Allows specifying for instances to be deleted instead of just stopped when the instance is preempted.

https://cloud.google.com/compute/docs/instances/create-use-spot#rest_1
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#instance_termination_action

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Add support for InstanceTerminationAction for instance preemption. `GCPMachine.Spec.InstanceTerminationAction = "Delete"`
```


salasberryfin:
/ok-to-test

Thanks for working on this @jwmay2012

jwmay2012:
I tested this code and it works. It's only supposed to be used with Spot VMs. Not STANDARD or the older `preemptible` VMs.
Normal behavior is the VM gets stopped, and the Node registers as UNKNOWN or shutdown/failed. 

The result with `InstanceTerminationAction="Delete"` is the Node gets deleted from the workload cluster along with the instance, but the `Machine` from the management cluster is still present and revives a Zombie instance. If that Spot instance was deleted via the console, it can come back online running. If that Spot instance was deleted via a preemption, it comes back `Stopped`.  In both cases, the previous `Node` never returns and you end up with a `Machine` without a `Node`.

🤷🏻‍♂️ 

This isn't as helpful as I was hoping it would be.  I was hoping this would allow Spot machines to get automatically re-created when a VM is preempted. But no.

It might be worth it to have this option exposed, but I don't know if I can consider it useful given how CAPI reacts (or doesn't react) to it. 

Up to you :D

PR #1357, opened 2024-11-06
Bump CAPI to v1.9.z 

damdo:
**What type of PR is this?**
/kind feature

CAPI v1.9 is out now, so this PRs bumps the CAPI version used by CAPG to it.

Bumps CAPI to v1.9.z based on https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/providers/migrations/v1.8-to-v1.9.md

api-diff failing is expected as the API changed to a typed go generics function.

**Checklist**:

- [x] squashed commits
- [x] adds or updates e2e tests

**Release note**:
```release-note
Bump CAPI to v1.9.z
```


damdo:
TODO: try and add something like https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/85759ceebd3a679b599b2dd9ded48eb945fbdc28/test/e2e/shared/common.go#L68 in the AfterSuite of the e2e tests to dump details before the teardown.

Makefile
```diff
@@ -32,8 +32,11 @@ export GOPROXY
 # Active module mode, as we use go modules to manage dependencies
 export GO111MODULE=on
 
+# Go version
+GOLANG_VERSION := 1.22.8
```
cpanato, line 36: 1.22.11 use the latest as we already have in the dockerfile https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/Dockerfile#L16

cpanato: oh i think capi is using 1.23 but we can do that in a follow up

cpanato (changes requested):
a small nit

thanks for working on this

cc @justinsb

Makefile
damdo: Updated to 1.22.11 thanks @cpanato

cpanato (approved):
thanks!

Makefile
damdo: @cpanato BTW it looks like CAPI 1.9.4 is still using go 1.22 https://github.com/kubernetes-sigs/cluster-api/blob/79e6731b0d20a0ca2dad5e4f96d2a532aa82b54b/Makefile#L26

salasberryfin (approved):
Thanks @damdo

PR #1367, opened 2024-11-13
chore: support registry and tag overrides in makefile

salasberryfin:
/kind cleanup

This change adds support for `STAGING_REGISTRY`/`PROD_REGISTRY` and `RELEASE_TAG` overrides in the `Makefile`. The current configuration does not allow setting a different value for these parameters via environment variables. This is a limitation in, for example, an air-gapped environment where a user may want to create their own release of the provider and reference an image stored in a custom registry.

It should not affect any existing use cases as the default value is unchanged.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


damdo (commented):
/lgtm

/assign @cpanato @richardcase

cpanato (approved):

PR #1376, opened 2024-11-22
Enable dual stack

barbacbd:
**What type of PR is this?**
/kind api-change
/kind feature

Adding the ability to enable dual stacks.
    
  Subnets are one of the resources that control the ability to use a dual stack (IPV4
  and IPV6). These are required to be set on resource creation, and they cannot be
  altered post creation.
    
  The user has the option to set the stack type for each subnet (IPV4_ONLY, IPV4_IPV6, or
  IPV6_ONLY). The default will be IPV4_ONLY.

Fixes # https://issues.redhat.com/browse/CORS-3762

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
 Adding the ability to enable dual stacks.
    
Subnets are one of the resources that control the ability to use a dual stack (IPV4 and IPV6). These are required to be set on resource creation, and they cannot be altered post creation.
    
The user has the option to set the stack type for each subnet (IPV4_ONLY, IPV4_IPV6, or IPV6_ONLY). The default will be IPV4_ONLY.

When creating compute.Subnetwork structs, set the StackType to the same value as the StackType in the SubnetSpecs.
```

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -240,6 +240,22 @@ type SubnetSpec struct {
 	// +kubebuilder:default=PRIVATE_RFC_1918
 	// +optional
 	Purpose *string `json:"purpose,omitempty"`
+
+	// StackType: The stack type for the subnet. If set to IPV4_ONLY, new VMs in
+	// the subnet are assigned IPv4 addresses only. If set to IPV4_IPV6, new VMs in
+	// the subnet can be assigned both IPv4 and IPv6 addresses. If not specified,
+	// IPV4_ONLY is used. This field can be both set at resource creation time and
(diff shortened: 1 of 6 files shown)


damdo (commented):
Thanks for this PR @barbacbd 

It looks reasonable to me
/lgtm

Let's see how it behaves in testing.

Also
/assign @cpanato @richardcase

cpanato:
the workload upgrade was failing before, we need to fix that :)

ozhankaraman:
Could we include this patch to latest minor or major release, this is an important fix for clusters dual stack feature?

PR #1382, opened 2024-12-11
rm caBundle

reyvonger:
What this PR does / why we need it:

Starting with Kubernetes 1.31 it won't be possible anymore to continuously apply CRDs that are setting caBundle to an invalid value (in our case Cg==). The solution is to simply drop the caBundle field (it was never actually required by kube-apiserver).
For more details see: https://kubernetes.slack.com/archives/C0EG7JC6T/p1722441161968339

Let me know if you have any questions.

We'll want to backport this into all supported releases. It's never great to set the caBundle to an invalid value, even before Kubernetes 1.31.

This is an effort driven by CAPI: https://github.com/kubernetes-sigs/cluster-api/pull/10972

### DIFF
--- a/config/crd/patches/webhook_in_gcpclusters.yaml
+++ b/config/crd/patches/webhook_in_gcpclusters.yaml
@@ -10,9 +10,6 @@ spec:
     webhook:
       conversionReviewVersions: ["v1", "v1beta1"]
       clientConfig:
-        # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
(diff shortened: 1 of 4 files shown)


salasberryfin:
/ok-to-test

Thanks @reyvonger. From the Slack thread you shared, I understand not setting `caBundle` would be safe for any Kubernetes version and not only `>=v1.31`, so there's no harm in back-porting this. Is this correct?

reyvonger:
Yes

cpanato:
thanks

/approve
/lgtm

/cherrypick release-1.8

PR #1385, opened 2024-12-17
feat: add support for identity service server and updating identity service 

afarbos:
**What type of PR is this?**

/kind feature

The goal of this PR is to expose the server used by identity enabling easier authentication without prior cluster access or secret manipulation and update of the identity service config.

Doc: https://cloud.google.com/kubernetes-engine/docs/how-to/oidc

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
GKEManagedControlPlane: Add support for identity service server in status and updating identity service 
```


afarbos:
I ran locally: `make generate`, `make verify` and `make test` all pass.
I also verified everything works using tilt.

salasberryfin:
Thanks @afarbos. Could you please rebase your commits to squash them into logical changes? This helps maintain a clean history and simplifies reverts, if needed.

afarbos:
> Thanks @afarbos. Could you please rebase your commits to squash them into logical changes? This helps maintain a clean history and simplifies reverts, if needed.

sounds good, done! 

3 commits:
- fix local dev
- fix add update logic for identity server
- feat expose identity service server

cloud/services/container/clusters/kubeconfig.go
```diff
@@ -41,34 +41,35 @@ const (
...
 			log,
-		); createErr != nil {
-			return fmt.Errorf("creating kubeconfig secret: %w", createErr)
+		); err != nil {
+			return nil, fmt.Errorf("creating kubeconfig secret: %w", err)
 		}
-	} else if updateErr := s.updateCAPIKubeconfigSecret(ctx, configSecret); updateErr != nil {
-		return fmt.Errorf("updating kubeconfig secret: %w", err)
+	} else if kubeConfig, err = s.updateCAPIKubeconfigSecret(ctx, configSecret); err != nil {
+		return nil, fmt.Errorf("updating kubeconfig secret: %w", err)
```
afarbos, line 69: note: this also fix the error it should have been `updateErr`, now everything is `err`

cloud/services/container/clusters/reconcile.go
```diff
@@ -482,3 +498,70 @@ func compareMasterAuthorizedNetworksConfig(a, b *containerpb.MasterAuthorizedNet
...
+		metadata:
+			name: default
+			namespace: kube-public
+		spec:
+			server: https://192.168.0.1:6443
+	*/
+
+	if !s.scope.GCPManagedControlPlane.Spec.EnableIdentityService {
+		// Identity service is not enabled, skipping
+		return "", nil
```
salasberryfin, line 531: If identity service is not enabled, `Status.IdentityServer` will be nil. Is this safe for scenarios where we're not using this feature?

cloud/services/container/clusters/reconcile.go
```diff
@@ -482,3 +498,70 @@ func compareMasterAuthorizedNetworksConfig(a, b *containerpb.MasterAuthorizedNet
...
+	}
+
+	dynamicClient, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return "", err
+	}
+
+	resourceID := schema.GroupVersionResource{
+		Group:    "authentication.gke.io",
+		Version:  "v2alpha1",
```
salasberryfin, line 550: Should we parameterize this for simplified maintenance in the future?

salasberryfin: I'm assuming it's subject to change periodically as new API versions are released.

cloud/services/container/clusters/reconcile.go
afarbos: Yes, since the server per say does not exist in this scenario. Additionally, I set the string to be optional here https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1385/files#diff-51a950bdfb5cfcf3f2ad6518b8e25bbe6ff584d534f6b9ceeb3aaece9895f30eR177 so it would just not be here.

To be clear this is a string, so it would be empty not nil. In case, you were worried about panic.

afarbos:
@salasberryfin friendly bump

afarbos:
> PR needs rebase.

rebased

afarbos:
rebased

salasberryfin:
Hey @afarbos, thanks for the contributions (and the patience) and sorry for the delay in responding. The only issue I see with this is that we're not actively testing managed cluster functionality so, would it be okay if we wait for this https://github.com/kubernetes/k8s.io/issues/7665 request to be resolved so we can actually test that GKE provisioning is functioning properly?

afarbos:
> You will need to rebase this to fix the `gcp-apidiff`.

ack, rebased. 👀

cloud/services/container/clusters/reconcile.go
```diff
@@ -168,6 +173,11 @@ func (s *Service) Reconcile(ctx context.Context) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
+	err = s.reconcileIdentityService(ctx, kubeConfig, &log)
```
justinsb, line 176: Nit: I always recommend the one-line form when we don't need the error later:

```
if err := s.reconcileIdentityService(..); err != nil {
```

(No need to fix, just my 2c)

cloud/services/container/clusters/reconcile.go
```diff
@@ -481,6 +491,13 @@ func (s *Service) checkDiffAndPrepareUpdate(existingCluster *containerpb.Cluster
 		log.V(4).Info("Master authorized networks config update check", "desired", desiredMasterAuthorizedNetworksConfig)
 	}
 
+	desiredEnableIdentityService := s.scope.GCPManagedControlPlane.Spec.EnableIdentityService
+	if desiredEnableIdentityService != existingCluster.GetIdentityServiceConfig().GetEnabled() {
+		needUpdate = true
+		clusterUpdate.DesiredIdentityServiceConfig = &containerpb.IdentityServiceConfig{Enabled: desiredEnableIdentityService}
```
justinsb, line 497: FYI I believe that _some_ (OK, most) fields cannot be updated in "one shot".  https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/ClusterUpdate says "Exactly one update can be applied to a cluster with each request, so at most one field can be provided."

I think the easiest way to handle this is probably to build the UpdateClusterRequest as we are doing here, but then to break it down into one-field-at-a-time requests when we actually go to call UpdateCluster

(I don't know whether we want to handle in this PR - or maybe it is handled somewhere else and I missed it - but it is a classic gotcha that I'm sure we'll hit!)

cloud/services/container/clusters/reconcile.go
```diff
@@ -516,3 +533,78 @@ func compareMasterAuthorizedNetworksConfig(a, b *containerpb.MasterAuthorizedNet
 	}
 	return true
 }
+
+// reconcileIdentityService set the identity service server in the status of the GCPManagedControlPlane.
+func (s *Service) reconcileIdentityService(ctx context.Context, kubeConfig clientcmd.ClientConfig, log *logr.Logger) error {
+	identityServiceServer, err := s.getIdentityServiceServer(ctx, kubeConfig)
+	if err != nil {
+		err = fmt.Errorf("failed to retrieve identity service: %w", err)
+		log.Error(err, "Failed to retrieve identity service server")
```
justinsb, line 542: Nit: I personally think we shouldn't do this, we should rely on the caller logging, but I'm guessing this happens more often than we would like

(Another thought, likely not for this PR - we should decide whether we should pass the logr in, vs getting it from the ctx)

cloud/services/container/clusters/reconcile.go
salasberryfin: This is an interesting limit I did not know about. We're actually updating the `UpdateClusterRequest` with all of the changes we detect and then calling `UpdateCluster()`. Should we be having issues with GCP rejecting multiple changes at the same time? I don't recall seeing this. Does it mean that only one of the updated fields is applied and an update of multiple fields needs as many reconciliations?

cloud/services/container/clusters/reconcile.go
salasberryfin: I second this, it is cleaner to have the caller logging the error and the called method only return the error here.

I also agree on the logger being passed as an argument, which we're doing already, and I think is overly convoluted, but I suggest we discuss this in a separate issue and open a stand-alone PR to tidy it up.

cloud/services/container/clusters/reconcile.go
afarbos: I can update to:
```
# Update alone, only one update at a time is supported 
if !needUpdate && desiredEnableIdentityService != existingCluster.GetIdentityServiceConfig().GetEnabled() {
```
That should be done everywhere after. does that look better?

cloud/services/container/clusters/reconcile.go
afarbos: I follow the current patterns, I am happy to change. I am not 100% sure what we want to remove.
Should I just return?
```
		return err
```

PR #1392, opened 2025-01-14
Add AllowGlobalAccess support for InternalLoadBalancer.

jwmay2012:
**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -371,4 +371,12 @@ type LoadBalancer struct {
 	// required for the Load Balancer, if not defined the first configured subnet will be
 	// used.
 	Subnet *string `json:"subnet,omitempty"`
+
+	// AllowGlobalAccess. If set to true, clients can access the internal
+	// passthrough Network Load Balancers, the regional internal Application Load
+	// Balancer, and the regional internal proxy Network Load Balancer from all
+	// regions. If false, only allows access from the local region the load
+	// balancer is located at. Note that for INTERNAL_MANAGED forwarding rules,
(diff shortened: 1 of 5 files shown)


api/v1beta1/types.go
```diff
@@ -371,4 +371,12 @@ type LoadBalancer struct {
...
 	// used.
 	Subnet *string `json:"subnet,omitempty"`
+
+	// AllowGlobalAccess. If set to true, clients can access the internal
+	// passthrough Network Load Balancers, the regional internal Application Load
+	// Balancer, and the regional internal proxy Network Load Balancer from all
+	// regions. If false, only allows access from the local region the load
+	// balancer is located at. Note that for INTERNAL_MANAGED forwarding rules,
+	// this field cannot be changed after the forwarding rule is created.
+	AllowGlobalAccess bool `json:"allowGlobalAccess,omitempty"`
```
barbacbd, line 381: It is hard to tell based on the comments in the file, but I feel like this field should exist somewhere else or this struct name needs to change (or the comments should be altered slightly). The struct appears to only be used for Internal Load Balancers but the way that the comments are worded make this struct seem like it should be for external and internal Load Balancers. Allow global access only appears to be related to internal LBs, so something isn't matched in the description(s) here. 

nit: not necessarily a problem with this PR but maybe we should address it.

jwmay2012:
Closing, as already implemented by https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1469

PR #1434, opened 2025-02-25
feat: align GCPManagedControlPlane with CAPI >=v1.9

### LINKED ISSUE #1433: Align managed control plane to latest CAPI changes
/kind bug
/area gke

# Description

To align `GCPManagedControlPlane` specification and status with CAPI >=v1.9 we need to introduce new control plane version fields and set existing ones for deprecation in coming releases. The list of changes, as agreed on Slack (thanks @damdo @richardcase @cpanato):
- Introduce new `Version` fields in `GCPManagedControlPlaneSpec` and `GCPManagedControlPlaneStatus` as per the updated CAPI contract for control planes.
- Mark `ControlPlaneVersion` and `CurrentVersion` as deprecated with a message saying to use `Version` instead.
- Have a validation webhook that disallows setting `Version` and `ControlPlaneVersion`.
- Update all templates and e2e tests to use `Version`.

In the future, as the deprecated field is ultimately removed, we can add a conversion from `ControlPlaneVersion` to `Version`.

salasberryfin:
/kind bug

To align `GCPManagedControlPlane` specification and status with CAPI >=v1.9 we need to introduce new control plane version fields and set existing ones for deprecation in coming releases. The list of changes, as agreed on Slack (thanks @damdo @richardcase @cpanato):
- Introduce new `Version` fields in `GCPManagedControlPlaneSpec` and `GCPManagedControlPlaneStatus` as per the updated CAPI contract for control planes.
- Mark `ControlPlaneVersion` and `CurrentVersion` as deprecated with a message saying to use `Version` instead.
- Have a validation webhook that disallows setting `Version` and `ControlPlaneVersion`.
- Update all templates and e2e tests to use `Version`.

Fixes #1433 

In the future, as the deprecated field is ultimately removed, we can add a conversion from `ControlPlaneVersion` to `Version`.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
action required
!! ACTION REQUIRED BEFORE UPGRADING !!
If you are using the GCPManagedControlPlane to provision GKE clusters and you do not have a spec.Version specified in such resource (or you are using spec.ControlPlaneVersion), you will need to either:
a) explicitly set such `spec.Version` field before upgrading CAPG (if you are already using spec.ControlPlaneVersion, please, use spec.Version instead)
or b) disable the MachineSetPreflightChecks in your cluster either:
b1) by setting this core CAPI feature gate to `false`
b2) or by disabling it via the relevant annotation on all the machineSets belonging to said cluster (follow this guide on how to do this: https://cluster-api.sigs.k8s.io/tasks/experimental-features/machineset-preflight-checks).
This is necessary as core CAPI 1.9 introduces a feature gate change, setting MachineSetPreflightChecks=true, which in turn relies on the presence of spec.Version and status.Version on the GCPManagedControlPlane object.
These fields will be deprecated in a future release.
```

### DIFF
--- a/cloud/services/container/clusters/reconcile.go
+++ b/cloud/services/container/clusters/reconcile.go
@@ -96,7 +96,9 @@ func (s *Service) Reconcile(ctx context.Context) (ctrl.Result, error) {
 	}
 
 	log.V(2).Info("gke cluster found", "status", cluster.GetStatus())
-	s.scope.GCPManagedControlPlane.Status.CurrentVersion = convertToSdkMasterVersion(cluster.GetCurrentMasterVersion())
+	controlPlaneVersion := convertToSdkMasterVersion(cluster.GetCurrentMasterVersion())
+	s.scope.GCPManagedControlPlane.Status.CurrentVersion = controlPlaneVersion
+	s.scope.GCPManagedControlPlane.Status.Version = &controlPlaneVersion
 
 	switch cluster.GetStatus() {
 	case containerpb.Cluster_PROVISIONING:
@@ -271,8 +273,8 @@ func (s *Service) createCluster(ctx context.Context, log *logr.Logger) error {
 			},
 		},
 	}
-	if s.scope.GCPManagedControlPlane.Spec.ControlPlaneVersion != nil {
-		cluster.InitialClusterVersion = convertToSdkMasterVersion(*s.scope.GCPManagedControlPlane.Spec.ControlPlaneVersion)
+	if initialClusterVersionFromSpec := s.scope.GetControlPlaneVersion(); initialClusterVersionFromSpec != nil {
+		cluster.InitialClusterVersion = convertToSdkMasterVersion(*initialClusterVersionFromSpec)
 	}
 	if s.scope.GCPManagedControlPlane.Spec.ClusterNetwork != nil {
 		cn := s.scope.GCPManagedControlPlane.Spec.ClusterNetwork
@@ -434,8 +436,8 @@ func (s *Service) checkDiffAndPrepareUpdate(existingCluster *containerpb.Cluster
 		}
 	}
 	// Master version
-	if s.scope.GCPManagedControlPlane.Spec.ControlPlaneVersion != nil {
-		desiredMasterVersion := convertToSdkMasterVersion(*s.scope.GCPManagedControlPlane.Spec.ControlPlaneVersion)
+	if desiredMasterVersionFromSpec := s.scope.GetControlPlaneVersion(); desiredMasterVersionFromSpec != nil {
+		desiredMasterVersion := convertToSdkMasterVersion(*desiredMasterVersionFromSpec)
 		existingClusterMasterVersion := convertToSdkMasterVersion(existingCluster.GetCurrentMasterVersion())
 		if desiredMasterVersion != existingClusterMasterVersion {
 			needUpdate = true
--- a/exp/api/v1beta1/gcpmanagedcontrolplane_types.go
+++ b/exp/api/v1beta1/gcpmanagedcontrolplane_types.go
@@ -146,8 +146,16 @@ type GCPManagedControlPlaneSpec struct {
 	// ControlPlaneVersion represents the control plane version of the GKE cluster.
 	// If not specified, the default version currently supported by GKE will be
 	// used.
+	//
+	// Deprecated: This field will soon be removed and you are expected to use Version instead.
+	//
 	// +optional
 	ControlPlaneVersion *string `json:"controlPlaneVersion,omitempty"`
+	// Version represents the control plane version of the GKE cluster.
+	// If not specified, the default version currently supported by GKE will be
+	// used.
+	// +optional
+	Version *string `json:"version,omitempty"`
 	// Endpoint represents the endpoint used to communicate with the control plane.
 	// +optional
 	Endpoint clusterv1.APIEndpoint `json:"endpoint"`
@@ -183,8 +191,15 @@ type GCPManagedControlPlaneStatus struct {
 	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
 
 	// CurrentVersion shows the current version of the GKE control plane.
+	//
+	// Deprecated: This field will soon be removed and you are expected to use Version instead.
+	//
 	// +optional
 	CurrentVersion string `json:"currentVersion,omitempty"`
+
+	// Version represents the version of the GKE control plane.
+	// +optional
+	Version *string `json:"version,omitempty"`
 }
 
 // +kubebuilder:object:root=true
@@ -194,6 +209,7 @@ type GCPManagedControlPlaneStatus struct {
 // +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=".metadata.labels.cluster\\.x-k8s\\.io/cluster-name",description="Cluster to which this GCPManagedControlPlane belongs"
 // +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.ready",description="Control plane is ready"
 // +kubebuilder:printcolumn:name="CurrentVersion",type="string",JSONPath=".status.currentVersion",description="The current Kubernetes version"
+// +kubebuilder:printcolumn:name="Version",type="string",JSONPath=".status.version",description="The Kubernetes version of the GKE control plane"
 // +kubebuilder:printcolumn:name="Endpoint",type="string",JSONPath=".spec.endpoint",description="API Endpoint",priority=1
 
 // GCPManagedControlPlane is the Schema for the gcpmanagedcontrolplanes API.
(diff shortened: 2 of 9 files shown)


exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -147,7 +147,14 @@ type GCPManagedControlPlaneSpec struct {
 	// If not specified, the default version currently supported by GKE will be
 	// used.
 	// +optional
+	//
+	// Deprecated: This field will soon be removed and you are expected to use Version instead.
```
damdo, line 151: I think the optional stylistically (and maybe even from an implementation standpoint) goes next to the field, whereas the Deprecated can also go above in the godoc comment.
Better be safe, example here:
https://github.com/kubernetes-sigs/cluster-api/blob/7a71f4b0ed71cb92fa751e48057df6bb672319bb/api/v1beta1/cluster_types.go#L533-L539

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -184,7 +191,13 @@ type GCPManagedControlPlaneStatus struct {
 
 	// CurrentVersion shows the current version of the GKE control plane.
 	// +optional
+	//
+	// Deprecated: This field will soon be removed and you are expected to use Version instead.
```
damdo, line 195: Same here

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -184,7 +191,13 @@ type GCPManagedControlPlaneStatus struct {
...
 	CurrentVersion string `json:"currentVersion,omitempty"`
+
+	// Version represents the version of the GKE control plane.
+	// +optional
+	Version *string `json:"version,omitempty"`
 }
```
damdo, line 201: We want to populate this `status.Version` with whatever we currently populate the `status.CurrentVersion` with.
Then When in the future we remove the `status.CurrentVersion`, we will only have the `status.Version` to set.

One instance I saw of this is here: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/d80bfabadd3f611f5282d3ee7c2e963f3ffc4621/cloud/services/container/clusters/reconcile.go#L99

But there may be others.

damdo (changes requested):
Hey @salasberryfin 
Thanks a lot for working on implementing this!

Mostly LGTM, left a couple of things we need to change though.

damdo:
@salasberryfin Also we will likely need to add an `action required` in the release-note for this PR,
similar to the one I added for CAPA here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5209#issue-2637715307
(you should be able to mostly copy-paste it).

cc. @richardcase

damdo:
@salasberryfin also we might need to silence the linter for [errors like](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/actions/runs/13518827906/job/37773293001?pr=1434) : `cloud/services/container/clusters/reconcile.go:515:5: SA1019: spec.ControlPlaneVersion is deprecated: This field will soon be removed and you are expected to use Version instead. (staticcheck)`, as we still want to set/get those fields internally even if they are deprecated, until we remove them.

cloud/services/container/clusters/reconcile.go
```diff
@@ -506,3 +506,14 @@ func compareMasterAuthorizedNetworksConfig(a, b *containerpb.MasterAuthorizedNet
 	}
 	return true
 }
+
+// get the desired master version
+func getControlPlaneVersionFromSpec(spec *infrav1exp.GCPManagedControlPlaneSpec) *string {
```
richardcase, line 511: This could live as a function on the scope to make it a little neater.

richardcase (commented):
Great work @salasberryfin 🎉 

Just a couple of minor comments.

damdo:
No worries, thanks @salasberryfin

salasberryfin:
I think the latest commit addresses all of the comments @damdo @richardcase.

cpanato:
@salasberryfin there are a few lints to tackle, thanks for working on this

salasberryfin:
Thanks @cpanato. The exclude rule I added covered some of the cases where the linter found errors but missed some of them. Should be fixed now.

exp/api/v1beta1/gcpmanagedcontrolplane_webhook_test.go
```diff
@@ -72,21 +72,21 @@ func TestGCPManagedControlPlaneDefaultingWebhook(t *testing.T) {
...
 		{
 			name:         "with autopilot enabled",
 			resourceName: "cluster1",
 			resourceNS:   "default",
 			spec: GCPManagedControlPlaneSpec{
-				ClusterName:         "cluster1_autopilot",
-				ControlPlaneVersion: &vV1_27_1,
-				EnableAutopilot:     true,
+				ClusterName:     "cluster1_autopilot",
+				Version:         &vV1_27_1,
```
cpanato, line 86: can be this updated?

salasberryfin:
Also added the action required to the release note. Thanks @damdo!

alexander-demicev (approved):
/lgtm

cpanato (approved):
thanks!

/approve

/hold for @richardcase @damdo

exp/api/v1beta1/gcpmanagedcontrolplane_webhook_test.go
salasberryfin: I'll update this in a separate PR, if that's okay.

exp/api/v1beta1/gcpmanagedcontrolplane_webhook.go
```diff
@@ -98,6 +98,15 @@ func (r *GCPManagedControlPlane) ValidateCreate() (admission.Warnings, error) {
 			r.Spec.LoggingService, "can't be set when autopilot is enabled"))
 	}
 
+	if r.Spec.ControlPlaneVersion != nil {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "ControlPlaneVersion"),
+			r.Spec.LoggingService, "spec.ControlPlaneVersion is deprecated and will soon be removed: please use spec.Version"))
```
damdo, line 103: @salasberryfin Does this result in blocking any `GCPManagedControlPlane` CREATE if the `spec.ControlPlaneVersion` is set?
If so, I think this would be an implicit API breaking change, we should probably warn yes, but not error, at least until we remove the field from the API, when we bump it to v1beta2.

cc. @richardcase

damdo (commented):
@salasberryfin Mostly LGTM, I left one blocking question though. LMK thanks!

exp/api/v1beta1/gcpmanagedcontrolplane_webhook_test.go
cpanato: yes

cloud/services/container/clusters/reconcile.go
```diff
@@ -479,6 +481,17 @@ func (s *Service) checkDiffAndPrepareUpdate(existingCluster *containerpb.Cluster
 	return needUpdate, &updateClusterRequest
 }
 
+// get the desired master version
+func (s *Service) getControlPlaneVersionFromSpec() *string {
```
richardcase, line 485: We should probably add this to the scope instead. The reason being it can be re-used anywhere that a version is needed.

exp/api/v1beta1/gcpmanagedcontrolplane_webhook.go
richardcase: I think this needs to be changed to an admission warning instead.

richardcase: Returning an error would be a breaking change.

salasberryfin: Thanks for catching this!

cloud/services/container/clusters/reconcile.go
salasberryfin: Sorry I did not get this in the original review. Should be fixed now!

salasberryfin:
Thanks for the reviews. The webhook now returns a warning if only `spec.controlPlaneVersion` is used and an error if both `spec.controlPlaneVersion` and `spec.Version` are set. New tests have been added to verify the validation webhook on these scenarios.

damdo:
Hm looks like the E2Es failed https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-gcp/1434/pull-cluster-api-provider-gcp-e2e-test/1897214121795915776

Let's retest for good measure in case it is a flake, otherwise we might need to look into it.

/retest

salasberryfin:
Thanks @damdo, looks like it was a flake.

damdo (commented):
/lgtm

Thanks @salasberryfin for working on this

/unhold

PR #1446, opened 2025-03-17
chore: update tiltfile for immutable images

justinsb:
The tiltfile was broken (trying to write process.txt).

Sync up with the values in the (main) cluster-api tiltfile.


justinsb:
Do people still use tilt?  I'm wondering if it's worth fixing...

damdo (commented):
/lgtm

Thanks @justinsb

PR #1469, opened 2025-05-03
option to enable global access for internal passthrough load balancers

### LINKED ISSUE #1468: Add an option to enable global access for internal passthrough load balancers
/kind feature

Add an option to enable global access for internal passthrough load balancers.

When Cluster API (CAPI) is deployed in a different region or outside of the Google Cloud Platform (GCP), reconciliation fails because CAPI cannot reach the control plane API. Enabling the global access option on the internal passthrough load balancer would resolve this issue.

hrbasic:
/kind feature

When Cluster API (CAPI) is deployed in a different region or outside of the Google Cloud Platform (GCP), reconciliation fails because CAPI cannot reach the control plane API. Enabling the global access option on the internal passthrough load balancer would resolve this issue.

**Which issue(s) this PR fixes** 
Fixes #1468

**Release note**:
```release-note
Option for enabling global access for internal passthrough load balancer.
```

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -357,6 +357,17 @@ type ObjectReference struct {
 	Name string `json:"name"`
 }
 
+// InternalAccess defines the access for the Internal Passthrough Load Balancer.
+type InternalAccess string
+
+const (
+	// InternalAccessRegional restricts traffic to clients within the same region as the internal load balancer.
+	InternalAccessRegional = InternalAccess("Regional")
+
+	// InternalAccessGlobal allows traffic from any region to access the internal load balancer.
+	InternalAccessGlobal = InternalAccess("Global")
+)
+
 // LoadBalancer specifies the configuration of a LoadBalancer.
 type LoadBalancer struct {
 	// Name is the name of the Load Balancer. If not set a default name
@@ -371,4 +382,17 @@ type LoadBalancer struct {
 	// required for the Load Balancer, if not defined the first configured subnet will be
 	// used.
 	Subnet *string `json:"subnet,omitempty"`
+
+	// InternalAccess defines the access for the Internal Passthrough Load Balancer.
+	// It determines whether the load balancer allows global access,
+	// or restricts traffic to clients within the same region as the load balancer.
+	// If unspecified, the value defaults to "Regional".
+	//
+	// Possible values:
+	//   "Regional" - Only clients in the same region as the load balancer can access it.
+	//   "Global" - Clients from any region can access the load balancer.
+	// +kubebuilder:validation:Enum=Regional;Global
+	// +kubebuilder:default=Regional
+	// +optional
+	InternalAccess InternalAccess `json:"internalAccess,omitempty"`
 }
--- a/cloud/services/compute/loadbalancers/reconcile.go
+++ b/cloud/services/compute/loadbalancers/reconcile.go
@@ -601,6 +601,10 @@ func (s *Service) createOrGetRegionalForwardingRule(ctx context.Context, lbname 
 	spec.LoadBalancingScheme = string(loadBalanceTrafficInternal)
 	spec.Region = s.scope.Region()
 	spec.BackendService = backendSvc.SelfLink
+	lbSpec := s.scope.LoadBalancer()
+	if lbSpec.InternalLoadBalancer != nil && lbSpec.InternalLoadBalancer.InternalAccess == infrav1.InternalAccessGlobal {
+		spec.AllowGlobalAccess = true
+	}
 	// Ports is used instead or PortRange for passthrough Load Balancer
 	// Configure ports for k8s API to match the external API which is the first port of range
 	var ports []string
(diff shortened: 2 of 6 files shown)


api/v1beta1/types.go
```diff
@@ -371,4 +371,10 @@ type LoadBalancer struct {
 	// required for the Load Balancer, if not defined the first configured subnet will be
 	// used.
 	Subnet *string `json:"subnet,omitempty"`
+
+	// AllowGlobalAccess allows clients from any region
+	// to access Internal Passthrough Network Load Balancer.
+	// Defaults to false.
+	// +optional
+	AllowGlobalAccess *bool `json:"allowGlobalAccess,omitempty"`
```
richardcase, line 379:
```suggestion
	AllowGlobalAccess bool `json:"allowGlobalAccess,omitempty"`
```

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -601,6 +601,10 @@ func (s *Service) createOrGetRegionalForwardingRule(ctx context.Context, lbname
 	spec.LoadBalancingScheme = string(loadBalanceTrafficInternal)
 	spec.Region = s.scope.Region()
 	spec.BackendService = backendSvc.SelfLink
+	lbSpec := s.scope.LoadBalancer()
+	if lbSpec.InternalLoadBalancer != nil {
+		spec.AllowGlobalAccess = ptr.Deref(lbSpec.InternalLoadBalancer.AllowGlobalAccess, false)
```
richardcase, line 606:
```suggestion
		spec.AllowGlobalAccess = lbSpec.InternalLoadBalancer.AllowGlobalAccess
```

richardcase (commented):
Thanks for this @hrbasic 🙇 

The pointer isn't needed on the bool field if you don't want to distinguish between "not set" and "false". 

Apart from that, looks good to me.

api/v1beta1/types.go
damdo, line 379: The use of Booleans in new Kubernetes (and CAPI) APIs is strongly discouraged.
Reference in the Kubernetes API conventions: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

The suggested replacement is to use an Enum Type. So for example in this case you could define something like:
`internalAccess`, which can take two values: `Regional | Global`

richardcase: Thats a good point @damdo. And KAL will pick that up when we add it.

I don't feel overly opinionated on the use of a bool here, but an enum will be more future-proof for sure.

hrbasic: Thank you for your review and comments. I've updated the code to use an enum instead of a bool, as you suggested. Please let me know if there's anything else you'd like me to address.

api/v1beta1/types.go
```diff
@@ -357,6 +357,17 @@ type ObjectReference struct {
 	Name string `json:"name"`
 }
 
+// InternalAccess defines the access for the Internal Passthrough Load Balancer.
+type InternalAccess string
+
+const (
+	// Regional restricts traffic to clients within the same region as the load balancer.
+	Regional = InternalAccess("Regional")
```
damdo, line 365:
```suggestion
	InternalAccessRegional = InternalAccess("Regional")
```

api/v1beta1/types.go
```diff
@@ -357,6 +357,17 @@ type ObjectReference struct {
...
+
+	// Global allows traffic from any region to access the load balancer.
+	Global = InternalAccess("Global")
```
damdo, line 368:
```suggestion
	InternalAccessGlobal = InternalAccess("Global")
```

api/v1beta1/types.go
```diff
@@ -357,6 +357,17 @@ type ObjectReference struct {
 	Name string `json:"name"`
 }
 
+// InternalAccess defines the access for the Internal Passthrough Load Balancer.
+type InternalAccess string
+
+const (
+	// Regional restricts traffic to clients within the same region as the load balancer.
```
damdo, line 364:
```suggestion
	// InternalAccessRegional restricts traffic to clients within the same region as the internal load balancer.
```

api/v1beta1/types.go
```diff
@@ -357,6 +357,17 @@ type ObjectReference struct {
...
+	Regional = InternalAccess("Regional")
+
+	// Global allows traffic from any region to access the load balancer.
```
damdo, line 367:
```suggestion
	// InternalAccessGlobal allows traffic from any region to access the internal load balancer.
```

damdo (commented):
Thanks for the change to Enum, I left some nits on the consts naming but mostly LGTM

api/v1beta1/types.go
hrbasic: Thank you for the suggestion. I've updated the code accordingly.

PR #901, opened 2023-04-18
Adding GCPMachinePool Functionality

BrennenMM7:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This PR Implements GCPMachinePool using [Managed Instance Groups](https://cloud.google.com/compute/docs/instance-groups/). This is needed to support the CAPI MachinePools object. Feature will allow cluster operators to deploy GCP MIG's to their CAPG clusters to take advantage of both Google Autoscaling of Compute Instances or utilizing cluster autoscaler to dynamically provision node resources. 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #297,#1005

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Feature: adds MachinePool/GCPMachinePool Objects to deploy Managed Instance Group services. 
```

PR #1478, opened 2025-05-22
Adding GCPMachinePool Functionality (continued)

justinsb:
- **Adding GCPMachinePool Functionality**
  

- **make generate-manifests**
  

- **Update controller for new controller-runtime**
  

- **MetadataItem: key cannot be optional**
  

- **Tweak error message**


justinsb:
When I looked at #901 previously I thought there were three things that I thought maybe we should address or split out:

1. Should we drain?  https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/901#discussion_r1999708518
2. Should GCPMachinePoolMachine just be a GCPMachine?  https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/901#discussion_r1999710916
3. Do we want to address the kubeadm dependency here vs splitting it out in another PR?  (I personally very much want to loosen the coupling, but I feel like maybe it is a separate PR): https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/901#discussion_r1999711928

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,291 @@
...
+
+	c, err := ctrl.NewControllerManagedBy(mgr).
+		WithOptions(options).
+		For(&infrav1exp.GCPMachinePool{}).
+		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(mgr.GetScheme(), log, r.WatchFilterValue)).
+		Watches(
+			&expclusterv1.MachinePool{},
+			handler.EnqueueRequestsFromMapFunc(machinePoolToInfrastructureMapFunc(gvk)),
+		).
+		// watch for changes in KubeadmConfig to sync bootstrap token
```
salasberryfin, line 89: We should let users select the bootstrap provider of their choice. This dependency used to exist in other infrastructure providers (e.g. Azure) but it is now removed https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4931/files#diff-090609f25d4709ca183257330305b845613ce137e49ce1f6c71512e338607bff

justinsb: Good news - in #1506 we no longer watch KubeadmConfig, we watch the MachinePool instead!

PR #1445, opened 2025-03-17
run gke e2e on pr ci

salasberryfin:
**What type of PR is this?**

/kind other

**What this PR does / why we need it**:

GKE E2E are not running on new PRs. Let's modify the ginkgo filter to run managed cluster tests and verify if existing templates pass CI checks with the current configuration.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

PR #1479, opened 2025-05-26
test: use unique cluster name to avoid collisions

salasberryfin:
/kind failing-test

Use unique cluster names for each test case to prevent name collisions. Reusing cluster resource names creates a situation where the `Cluster` object may be updated by a different cluster flavor, which may imply that the two definitions are not compatible and a webhook will disallow the change. Thanks @damdo for investigating and getting to the bottom of this.

Fixes #

The same change should be applied to GKE test cases. We can do that in the existing PR https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1445, when we're ready to successfully run these tests.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


damdo (commented):
It works!
Thanks Carlos 

/lgtm

PR #1487, opened 2025-06-10
test: use gcp ccm in e2e cluster templates

salasberryfin:
/kind cleanup

The in-tree GCE cloud provider is no longer supported on newer versions on Kubernetes and we need to move to using CCM https://github.com/kubernetes/cloud-provider-gcp

This PR includes bumping Kubernetes version to v1.32.5 (authored by @damdo) and, with this bump, Calico needs to be updated, as the version we've been using until now is not considered compatible with Kubernetes v1.32.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


test/e2e/data/ccm/gce-cloud-controller-manager.yaml
```diff
@@ -1,47 +1,58 @@
...
+        effect: NoSchedule
+      - key: node-role.kubernetes.io/control-plane
+        effect: NoSchedule
+      serviceAccountName: cloud-controller-manager
       containers:
-      - args:
-        - --allocate-node-cidrs=true
-        - --cidr-allocator-type=RangeAllocator
+      - name: cloud-controller-manager
+        image: gcr.io/k8s-staging-cloud-provider-gcp/cloud-controller-manager:master
```
upodroid, line 44: can we use an image from registry.k8s.io/cloud-provider-gcp please? https://explore.ggcr.dev/?repo=registry.k8s.io%2Fcloud-provider-gcp%2Fcloud-controller-manager

salasberryfin: This was originally (before this PR) set to use `gcr.io`. @cpanato should we keep it as is or should we move to using k8s registry?

damdo: gcr.io has been deprecated in favour of registry.k8s.io, better to switch if that still works.
https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/

Fun fact this blog post is from @upodroid himself!

cpanato: Here we are using the staging one and using the master tag, which changes every commit to the main branch. Was that because in the past we used to test with the CI artifacts, which also use k8s from the master branch, now that we are using for a specific k8s version we can use ccm in the registry and see if we have that tag available for the k8s version (minor) we are using.

but I am fine to keep as is for now, we can follow up on that. I still have work to do in the ccm to get that in shape

cpanato:
lets re-run a few more times before merging

/test pull-cluster-api-provider-gcp-e2e-test

cpanato:
lets see if with this the ci-artifacts still run, if not we can follow up later

/test ?

PR #1506, opened 2025-08-16
Minimal MachinePool support

justinsb:
- **Initial spike: GCPMachinePool**
  

- **GCPMachinePool: generated code/manifests**
 
This continues the work started by @BrennenMM7 in https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/901 .  I also combined in the support from cluster-api-provider-aws to see what we want to borrow from that, and will whittle the code we don't need from cluster-api-provider-aws away.

```release-note

NONE
```

### DIFF
--- a/cloud/scope/machine_test.go
+++ b/cloud/scope/machine_test.go
@@ -78,21 +78,6 @@ func TestMachineLocalSSDDiskType(t *testing.T) {
 
 // TestInstanceNetworkInterfaceAliasIPRangesSpec tests the InstanceNetworkInterfaceAliasIPRangesSpec function
 func TestInstanceNetworkInterfaceAliasIPRangesSpec(t *testing.T) {
-	// Register the GCPMachine and GCPMachineList in a schema.
-	schema, err := infrav1.SchemeBuilder.Register(&infrav1.GCPMachine{}, &infrav1.GCPMachineList{}).Build()
-	assert.Nil(t, err)
-
-	// Create a controller fake client.
-	testClient := fake.NewClientBuilder().WithScheme(schema).Build()
-
-	// Test machine parameter
-	failureDomain := "us-central1-a"
-	testMachine := clusterv1.Machine{
-		Spec: clusterv1.MachineSpec{
-			FailureDomain: failureDomain,
-		},
-	}
-
 	t.Run("should return nil for empty alias IP ranges", func(t *testing.T) {
 		testGCPMachine := infrav1.GCPMachine{
 			Spec: infrav1.GCPMachineSpec{
@@ -100,17 +85,7 @@ func TestInstanceNetworkInterfaceAliasIPRangesSpec(t *testing.T) {
 			},
 		}
 
-		testScopeParams := MachineScopeParams{
-			Client:     testClient,
-			Machine:    &testMachine,
-			GCPMachine: &testGCPMachine,
-		}
-
-		testMachineScope, err := NewMachineScope(testScopeParams)
-		assert.Nil(t, err)
-		assert.NotNil(t, testMachineScope)
-
-		result := testMachineScope.InstanceNetworkInterfaceAliasIPRangesSpec()
+		result := InstanceNetworkInterfaceAliasIPRangesSpec(testGCPMachine.Spec.AliasIPRanges)
 		assert.Nil(t, result)
 	})
 
@@ -126,17 +101,7 @@ func TestInstanceNetworkInterfaceAliasIPRangesSpec(t *testing.T) {
 			},
 		}
 
-		testScopeParams := MachineScopeParams{
-			Client:     testClient,
-			Machine:    &testMachine,
-			GCPMachine: &testGCPMachine,
-		}
-
-		testMachineScope, err := NewMachineScope(testScopeParams)
-		assert.Nil(t, err)
-		assert.NotNil(t, testMachineScope)
-
-		result := testMachineScope.InstanceNetworkInterfaceAliasIPRangesSpec()
+		result := InstanceNetworkInterfaceAliasIPRangesSpec(testGCPMachine.Spec.AliasIPRanges)
 		assert.NotNil(t, result)
 		assert.Len(t, result, 1)
 		assert.Equal(t, "10.0.0.0/24", result[0].IpCidrRange)
@@ -159,17 +124,7 @@ func TestInstanceNetworkInterfaceAliasIPRangesSpec(t *testing.T) {
 			},
 		}
 
-		testScopeParams := MachineScopeParams{
-			Client:     testClient,
-			Machine:    &testMachine,
-			GCPMachine: &testGCPMachine,
-		}
-
-		testMachineScope, err := NewMachineScope(testScopeParams)
-		assert.Nil(t, err)
-		assert.NotNil(t, testMachineScope)
-
-		result := testMachineScope.InstanceNetworkInterfaceAliasIPRangesSpec()
+		result := InstanceNetworkInterfaceAliasIPRangesSpec(testGCPMachine.Spec.AliasIPRanges)
 		assert.NotNil(t, result)
 		assert.Len(t, result, 2)
 		assert.Equal(t, "10.0.0.0/24", result[0].IpCidrRange)
@@ -190,17 +145,7 @@ func TestInstanceNetworkInterfaceAliasIPRangesSpec(t *testing.T) {
 			},
 		}
 
-		testScopeParams := MachineScopeParams{
-			Client:     testClient,
-			Machine:    &testMachine,
-			GCPMachine: &testGCPMachine,
-		}
-
-		testMachineScope, err := NewMachineScope(testScopeParams)
-		assert.Nil(t, err)
-		assert.NotNil(t, testMachineScope)
-
-		result := testMachineScope.InstanceNetworkInterfaceAliasIPRangesSpec()
+		result := InstanceNetworkInterfaceAliasIPRangesSpec(testGCPMachine.Spec.AliasIPRanges)
 		assert.NotNil(t, result)
 		assert.Len(t, result, 1)
 		assert.Equal(t, "10.100.0.0/24", result[0].IpCidrRange)
--- a/cloud/scope/machinepool.go
+++ b/cloud/scope/machinepool.go
@@ -342,7 +342,8 @@ func (m *MachinePoolScope) InstanceTemplateResource(ctx context.Context) (*compu
 	instance.Disks = append(instance.Disks, m.InstanceAdditionalDiskSpec()...)
 	instance.Metadata = InstanceAdditionalMetadataSpec(m.GCPMachinePool.Spec.AdditionalMetadata)
 	instance.ServiceAccounts = append(instance.ServiceAccounts, instanceServiceAccountsSpec(m.GCPMachinePool.Spec.ServiceAccount))
-	instance.NetworkInterfaces = append(instance.NetworkInterfaces, InstanceNetworkInterfaceSpec(m.ClusterGetter, m.GCPMachinePool.Spec.PublicIP, m.GCPMachinePool.Spec.Subnet))
+	var aliasIPRanges []infrav1.AliasIPRange // Not supported by MachinePool
+	instance.NetworkInterfaces = append(instance.NetworkInterfaces, InstanceNetworkInterfaceSpec(m.ClusterGetter, m.GCPMachinePool.Spec.PublicIP, m.GCPMachinePool.Spec.Subnet, aliasIPRanges))
 	instance.GuestAccelerators = instanceGuestAcceleratorsSpec(m.GCPMachinePool.Spec.GuestAccelerators)
 	if len(instance.GuestAccelerators) > 0 {
 		instance.Scheduling.OnHostMaintenance = onHostMaintenanceTerminate
--- a/exp/api/v1beta1/gcpmachinepool_types.go
+++ b/exp/api/v1beta1/gcpmachinepool_types.go
@@ -45,10 +45,6 @@ type GCPMachinePoolSpec struct {
 	// +optional
 	Subnet *string `json:"subnet,omitempty"`
 
-	// // ProviderID is the unique identifier as specified by the cloud provider.
-	// // +optional
-	// ProviderID *string `json:"providerID,omitempty"`
-
 	// ImageFamily is the full reference to a valid image family to be used for this machine.
 	// +optional
 	ImageFamily *string `json:"imageFamily,omitempty"`
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -326,7 +326,7 @@ func instanceAdditionalDiskSpec(ctx context.Context, spec []infrav1.AttachedDisk
 }
 
 // InstanceNetworkInterfaceSpec returns compute network interface spec.
-func InstanceNetworkInterfaceSpec(cluster cloud.ClusterGetter, publicIP *bool, subnet *string) *compute.NetworkInterface {
+func InstanceNetworkInterfaceSpec(cluster cloud.ClusterGetter, publicIP *bool, subnet *string, aliasIPRanges []infrav1.AliasIPRange) *compute.NetworkInterface {
 	networkInterface := &compute.NetworkInterface{
 		Network: path.Join("projects", cluster.NetworkProject(), "global", "networks", cluster.NetworkName()),
 	}
@@ -344,18 +344,18 @@ func InstanceNetworkInterfaceSpec(cluster cloud.ClusterGetter, publicIP *bool, s
 		networkInterface.Subnetwork = path.Join("projects", cluster.NetworkProject(), "regions", cluster.Region(), "subnetworks", *subnet)
 	}
 
-	networkInterface.AliasIpRanges = m.InstanceNetworkInterfaceAliasIPRangesSpec()
+	networkInterface.AliasIpRanges = InstanceNetworkInterfaceAliasIPRangesSpec(aliasIPRanges)
 
 	return networkInterface
 }
 
 // InstanceNetworkInterfaceAliasIPRangesSpec returns a slice of Alias IP Range specs.
-func (m *MachineScope) InstanceNetworkInterfaceAliasIPRangesSpec() []*compute.AliasIpRange {
-	if len(m.GCPMachine.Spec.AliasIPRanges) == 0 {
+func InstanceNetworkInterfaceAliasIPRangesSpec(spec []infrav1.AliasIPRange) []*compute.AliasIpRange {
+	if len(spec) == 0 {
 		return nil
 	}
-	aliasIPRanges := make([]*compute.AliasIpRange, 0, len(m.GCPMachine.Spec.AliasIPRanges))
-	for _, alias := range m.GCPMachine.Spec.AliasIPRanges {
+	aliasIPRanges := make([]*compute.AliasIpRange, 0, len(spec))
+	for _, alias := range spec {
 		aliasIPRange := &compute.AliasIpRange{
 			IpCidrRange:         alias.IPCidrRange,
 			SubnetworkRangeName: alias.SubnetworkRangeName,
@@ -505,7 +505,7 @@ func (m *MachineScope) InstanceSpec(log logr.Logger) *compute.Instance {
 
 	instance.Metadata = InstanceAdditionalMetadataSpec(m.GCPMachine.Spec.AdditionalMetadata)
 	instance.ServiceAccounts = append(instance.ServiceAccounts, instanceServiceAccountsSpec(m.GCPMachine.Spec.ServiceAccount))
-	instance.NetworkInterfaces = append(instance.NetworkInterfaces, InstanceNetworkInterfaceSpec(m.ClusterGetter, m.GCPMachine.Spec.PublicIP, m.GCPMachine.Spec.Subnet))
+	instance.NetworkInterfaces = append(instance.NetworkInterfaces, InstanceNetworkInterfaceSpec(m.ClusterGetter, m.GCPMachine.Spec.PublicIP, m.GCPMachine.Spec.Subnet, m.GCPMachine.Spec.AliasIPRanges))
 	instance.GuestAccelerators = instanceGuestAcceleratorsSpec(m.GCPMachine.Spec.GuestAccelerators)
 	if len(instance.GuestAccelerators) > 0 {
 		instance.Scheduling.OnHostMaintenance = onHostMaintenanceTerminate


justinsb:
This PR is WIP while I whittle down the unneeded code from cluster-api-provider-aws and generally make this reviewable.  But I am uploading as this is a checkpoint that works (in a limited way!)

config/rbac/role.yaml
```diff
@@ -106,6 +107,17 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - infrastructure.cluster.x-k8s.io
+  resources:
+  - gcpmachinepools
+  verbs:
+  - delete
```
justinsb, line 115: This is not folder into the same block as e.g. gcpmachines because we don't need create.  I'm not sure that we need `create` on e.g. gcpclusters either, but ... that's a separate issue

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,330 @@
...
+			providerID = fmt.Sprintf("gce://%s/%s/%s", tokens[1], tokens[3], tokens[5])
+		} else {
+			return ctrl.Result{}, fmt.Errorf("unexpected instance URL format: %s", instance.Instance)
+		}
+
+		providerIDList[i] = providerID
+	}
+
+	// FUTURE: do we need to verify that the instances are actually running?
+	machinePoolScope.GCPMachinePool.Spec.ProviderIDList = providerIDList
```
justinsb, line 271: I really don't like this very much (it shouldn't be spec, it requires us to poll the cloud API), but it seems to be the MachinePool contract.

justinsb:
So I think I finally have this working with an e2e test in #1539 and (hopefully) in a mergeable state.

The big thing for the e2e test was populating spec.providerIDList.  I don't love that contract, but it is the MachinePool contract.

pkg/logger/logger.go
```diff
@@ -0,0 +1,121 @@
...
+	Warn(msg string, keysAndValues ...any)
+	Trace(msg string, keysAndValues ...any)
+	Error(err error, msg string, keysAndValues ...any)
+	WithValues(keysAndValues ...any) *Logger
+	WithName(name string) *Logger
+	GetLogger() logr.Logger
+}
+
+// Logger is a concrete logger using logr underneath.
+type Logger struct {
```
bochengchu: Just curious why we are adding a new logger here? And is it going to replace the loggers in other controllers?

justinsb: Yeah, I added this because CAPA does it, but I also think it's a good call @bochengchu  - it feels orthogonal.  I think I will remove it.

main.go
```diff
@@ -205,6 +207,19 @@ func setupReconcilers(ctx context.Context, mgr ctrl.Manager) error {
 		return fmt.Errorf("setting up GCPCluster controller: %w", err)
 	}
 
+	if feature.Gates.Enabled(capifeature.MachinePool) {
+		setupLog.Info("Enabling MachinePool reconcilers")
+		gcpMachinePoolConcurrency := gcpMachineConcurrency // FUTURE: Use our own flag while feature-gated?
```
chrischdi: +1 to use a seperate flag :-)

exp/api/v1beta1/gcpmachinepool_types.go
```diff
@@ -0,0 +1,276 @@
...
+	// Not meaningful by MachinePool (?)
+	// // AliasIPRanges let you assign ranges of internal IP addresses as aliases to a VM's network interfaces.
+	// // +optional
+	// AliasIPRanges []infrav1.AliasIPRange `json:"aliasIPRanges,omitempty"`
+
+	// Not meaningful for MachinePool
+	// // ProviderID is the unique identifier as specified by the cloud provider.
+	// // +optional
+	// ProviderID *string `json:"providerID,omitempty"`
+
```
chrischdi: ProviderIDList is the replacement here :-)

```suggestion
```

exp/api/v1beta1/gcpmachinepool_types.go
```diff
@@ -0,0 +1,276 @@
...
+	// the controller, and that manual intervention is required. Examples
+	// of terminal errors would be invalid combinations of settings in the
+	// spec, values that are unsupported by the controller, or the
+	// responsible controller itself being critically misconfigured.
+	//
+	// Any transient errors that occur during the reconciliation of MachinePools
+	// can be added as events to the MachinePool object and/or logged in the
+	// controller's output.
+	// +optional
+	FailureMessage *string `json:"failureMessage,omitempty"`
```
chrischdi: The corresponding fields in CAPI where this bubbles up to are deprecated, should we remove them here too?

exp/api/v1beta1/gcpmachinepool_types.go
```diff
@@ -0,0 +1,276 @@
...
+	// +optional
+	Ready bool `json:"ready"`
+
+	// Replicas is the most recently observed number of replicas
+	// +optional
+	Replicas int32 `json:"replicas"`
+
+	// Conditions defines current service state of the GCPMachinePool.
+	// +optional
+	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
```
chrischdi: Should we think about using metav1.Conditions? (considering clusterv1.Conditions gets removed we could already now start using metav1.Conditions and don't have to migrate later on)?

It's a question of consistency too, we maybe want to use clusterv1.Conditions, but I don't see a real reason why we can't use metav1.Conditions already now.

exp/api/v1beta1/gcpmachinepool_types.go
```diff
@@ -0,0 +1,276 @@
...
+
+// GetConditions returns the observations of the operational state of the GCPMachinePool resource.
+func (r *GCPMachinePool) GetConditions() clusterv1.Conditions {
+	return r.Status.Conditions
+}
+
+// SetConditions sets the underlying service state of the GCPMachinePool to the predescribed clusterv1.Conditions.
+func (r *GCPMachinePool) SetConditions(conditions clusterv1.Conditions) {
+	r.Status.Conditions = conditions
+}
```
chrischdi: If we do metav1, we would have to rewrite these and fulfill the new signature / names / interface

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,330 @@
...
+	err := r.Client.Get(ctx, req.NamespacedName, gcpMachinePool)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Fetch the CAPI MachinePool
+	machinePool, err := getOwnerMachinePool(ctx, r.Client, gcpMachinePool.ObjectMeta)
```
chrischdi: Note: CAPI Bump has a util `util.GetOwnerMachinePool(ctx, r.Client, gcpMachinePool.ObjectMeta)

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,330 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if machinePool == nil {
+		log.Info("MachinePool Controller has not yet set OwnerRef")
```
chrischdi, line 89: Should we handle here that if the GCPMachinePool has a DeletionTimestamp set, that we remove finalizers (in case we have)?

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,330 @@
...
+
+	// Fetch the Cluster.
+	clusterObj, err := capiutils.GetClusterFromMetadata(ctx, r.Client, machinePool.ObjectMeta)
+	if err != nil {
+		log.Info("MachinePool is missing cluster label or cluster does not exist")
+		return reconcile.Result{}, nil
+	}
+	log = log.WithValues("cluster", klog.KObj(clusterObj))
+
+	if capiutils.IsPaused(clusterObj, gcpMachinePool) {
```
chrischdi: Should we use the `paused.EnsurePausedCondition` (or if we stay at clusterv1.Conditions, have a fork of that func instead)?

Example from MDs:

```go
	if isPaused, requeue, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, deployment); err != nil || isPaused || requeue {
		return ctrl.Result{}, err
	}
```

(note: there should be an issue or proposal in core CAPI around this.)

exp/api/v1beta1/gcpmachinepool_types.go
chrischdi: Citing from the WIP contract. 

> The use of failureReason and failureMessage should not be used for new InfraMachinePool implementations. In other areas of the Cluster API, starting from the v1beta2 contract version, there is no more special treatment for provider’s terminal failures within Cluster API.

https://deploy-preview-12971--kubernetes-sigs-cluster-api.netlify.app/developer/providers/contracts/infra-machinepool#inframachinepool-terminal-failures

PR: https://github.com/kubernetes-sigs/cluster-api/pull/12971

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,330 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta1"
+	"sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions"
```
chrischdi: should we use the proper aliases and enforce them via golangci lint? E.g. 	`v1beta1conditions "sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions"`

Helps later on to migrate.

`golangci-lint --fix` automates the required changes.

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,330 @@
...
+	// 	// FUTURE: If we are in a failed state, delete the secret regardless of instance state
+
+	// 	return ctrl.Result{}, nil
+	// }
+
+	// if !machinePoolScope.Cluster.Status.InfrastructureReady {
+	// 	log.Info("Cluster infrastructure is not ready yet")
+	// 	conditions.MarkFalse(machinePoolScope.GCPMachinePool, expinfrav1.MIGReadyCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
+	// 	return ctrl.Result{}, nil
+	// }
```
chrischdi: Should either be commented in or removed?

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,330 @@
...
+
+	// FUTURE: do we need to verify that the instances are actually running?
+	machinePoolScope.GCPMachinePool.Spec.ProviderIDList = providerIDList
+	machinePoolScope.GCPMachinePool.Status.Replicas = int32(len(providerIDList))
+	machinePoolScope.GCPMachinePool.Status.Ready = true
+
+	// Requeue so that we can keep the spec.providerIDList and status in sync with the MIG.
+	// This is important for scaling up and down, as the CAPI MachinePool controller relies on
+	// the providerIDList to determine which machines belong to the MachinePool.
+	return ctrl.Result{RequeueAfter: 1 * time.Minute}, nil
```
chrischdi, line 278: So we always return with a RequeAfter of 1 minute.

Making syncPeriod basically to 1m.

Should we maybe return without a requeueafter if there is no work to do (e.g. existing providerIDs == replicas?)

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,330 @@
...
+		return err
+	}
+
+	// remove finalizer
+	controllerutil.RemoveFinalizer(machinePoolScope.GCPMachinePool, expinfrav1.MachinePoolFinalizer)
+
+	return nil
+}
+
+func (r *GCPMachinePoolReconciler) getInfraCluster(ctx context.Context, cluster *clusterv1.Cluster, gcpMachinePool *expinfrav1.GCPMachinePool) (*infrav1.GCPCluster, *scope.ClusterScope, error) {
```
chrischdi:
```suggestion
func (r *GCPMachinePoolReconciler) createScope(ctx context.Context, cluster *clusterv1.Cluster, gcpMachinePool *expinfrav1.GCPMachinePool) (*scope.ClusterScope, error) {
```

Should this be createScope or so? We never use the InfraCluster returned from this func.

cloud/scope/machinepool.go
```diff
@@ -0,0 +1,518 @@
...
+func (m *MachinePoolScope) PatchCAPIMachinePoolObject(ctx context.Context) error {
+	return m.capiMachinePoolPatchHelper.Patch(
+		ctx,
+		m.MachinePool,
+	)
+}
+
+// Close the MachinePoolScope by updating the machinepool spec, machine status.
+func (m *MachinePoolScope) Close() error {
+	return m.PatchObject(context.TODO())
```
chrischdi, line 158: Could we have a proper context here? Or would it break some interface?

cloud/scope/machinepool.go
```diff
@@ -0,0 +1,518 @@
...
+	return tags
+}
+
+// Role returns the machine role from the labels.
+func (m *MachinePoolScope) Role() string {
+	// FUTURE: Or template labels?
+	_, isControlPlane := m.MachinePool.Labels[clusterv1.MachineControlPlaneLabel]
+
+	if isControlPlane {
+		// FUTURE: Extract constants
```
chrischdi: TODO for this PR or should we have a follow-up or tracking issue? :-) (also for other `// Future` comments)

cloud/services/compute/instancegroupmanagers/instancegroupmanagers_reconcile.go
```diff
@@ -0,0 +1,165 @@
...
+			return nil, fmt.Errorf("getting instanceGroupManager %v: %w", selfLink, err)
+		}
+
+		log.V(2).Info("Creating instanceGroupManager")
+		if err := s.instanceGroupManagers.Insert(ctx, igmKey, desired); err != nil {
+			log.Error(err, "creating instanceGroupManager")
+			return nil, fmt.Errorf("creating instanceGroupManager %v: %w", selfLink, err)
+		}
+
+		actual, err = s.instanceGroupManagers.Get(ctx, igmKey)
```
chrischdi, line 126: Shouldn't we be able to return here? (or should we have a comment why we have to go through the below?)

cloud/services/compute/instancegroupmanagers/instancegroupmanagers_reconcile.go
```diff
@@ -0,0 +1,165 @@
...
+	if desired.InstanceTemplate != actual.InstanceTemplate {
+		log.V(2).Info("updating instanceTemplate for instanceGroupManager", "desired.instanceTemplate", desired.InstanceTemplate, "actual.instanceTemplate", actual.InstanceTemplate)
+		if err := s.instanceGroupManagers.SetInstanceTemplate(ctx, igmKey, &compute.InstanceGroupManagersSetInstanceTemplateRequest{
+			InstanceTemplate: desired.InstanceTemplate,
+		}); err != nil {
+			log.Error(err, "updating instanceTemplate for instanceGroupManager")
+			return nil, fmt.Errorf("updating instanceTemplate for instanceGroupManager %v: %w", selfLink, err)
+		}
+
+		actual.TargetSize = desired.TargetSize
```
chrischdi:
```suggestion
```

Duplicate to above?

cloud/services/compute/instancetemplates/instancetemplate_reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+	instanceTemplate, instanceTemplateKey, err := s.createOrGetInstanceTemplate(ctx)
+	if err != nil {
+		return nil, err
+	}
+	log.V(2).Info("binding to instanceTemplate", "selfLink", instanceTemplate.SelfLink)
+
+	return instanceTemplateKey, nil
+}
+
+// Delete delete machine instance.
```
chrischdi: Does it do that?
```suggestion
// Delete deletes all machine templates for a instance group.
```

cloud/services/compute/instancetemplates/instancetemplate_reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/filter"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"google.golang.org/api/compute/v1"
+
+	"sigs.k8s.io/cluster-api-provider-gcp/cloud/gcperrors"
+	"sigs.k8s.io/cluster-api-provider-gcp/pkg/gcp"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// Reconcile reconcile machine instance.
```
chrischdi: also commentsh ere are wrong :-)

cloud/services/compute/instancegroupmanagers/instancegroupmanagers_reconcile.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/filter"
+	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"google.golang.org/api/compute/v1"
+	"sigs.k8s.io/cluster-api-provider-gcp/cloud/gcperrors"
+	"sigs.k8s.io/cluster-api-provider-gcp/pkg/gcp"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// Reconcile reconcile machine instance.
```
chrischdi: Also needs proper comments

cloud/services/compute/instancetemplates/instancetemplate_reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("marshalling instance template to json: %w", err)
+	}
+	encoded := append([]byte(baseKey.Name), desiredJSON...)
+	hash := sha256.Sum256(encoded)
+	hashHex := hex.EncodeToString(hash[:])
+
+	namePrefix := baseKey.Name
+	suffix := hashHex[:16]
+	name := namePrefix + suffix
```
chrischdi, line 134: Theoretically, what if we hit a colission? It would use the old template right?

Should we have the full hash on a tag or something so we could compare? Or other ways to detect and react?

exp/api/v1beta1/gcpmachinepool_types.go
justinsb: It's not a function of spec, so I'm not sure that providerIDList belongs in status (expensive to keep up to date, and still never up to date)

It's also interesting that we don't need it!

exp/api/v1beta1/gcpmachinepool_types.go
justinsb: I'll remove - thanks.

exp/controllers/gcpmachinepool_controller.go
justinsb: I see you've done more of the CAPI migration.  I'll try to keep up.  But we should consider whether we prioritize features or version bumps.

exp/api/v1beta1/gcpmachinepool_types.go
justinsb: I'll give it a go.

exp/controllers/gcpmachinepool_controller.go
justinsb: It's a good point, we probably aren't cleaning up correctly...

exp/controllers/gcpmachinepool_controller.go
justinsb: I'll fix this instance.

Can we add a test (separately)?

exp/controllers/gcpmachinepool_controller.go
justinsb: I don't think we can ever terminate, if we are indeed populating the providerIDList (I'm not sure it's a good idea to have the providerIDList TBH)

exp/api/v1beta1/gcpmachinepool_types.go
justinsb: Ah I see, we have providerIDList.  I will remove providerID - good call.

I'll try to keep my feelings about providerIDList separate 😂

cloud/scope/machinepool.go
justinsb: Do you want to open an issue?  We have a FooScope pattern and none of them take context :-(

cloud/services/compute/instancegroupmanagers/instancegroupmanagers_reconcile.go
justinsb: Good call

cloud/services/compute/instancegroupmanagers/instancegroupmanagers_reconcile.go
justinsb: Will add a comment.  It doesn't cost us much to do a few extra comparisons, and it keeps the code simpler (one code path)

main.go
justinsb: I suggest we don't do this yet (while the controller is experimental)

exp/api/v1beta1/gcpmachinepool_types.go
justinsb: I think it worked well - nice suggestion, saves us from having to go the v1beta2.conditions route

exp/controllers/gcpmachinepool_controller.go
justinsb: After a rebase it worked well - thanks!

exp/controllers/gcpmachinepool_controller.go
justinsb: Because this happens in a bunch of places in capg, do you want to open an issue so we can fix it throughout capg - I'm not sure I follow what we want to do - set a condition?

exp/controllers/gcpmachinepool_controller.go
justinsb: (This is not fixed yet)

exp/controllers/gcpmachinepool_controller.go
justinsb: Removed!

cloud/scope/machinepool.go
justinsb: I'm just going to remove them; we'll find them again in the future.

cloud/services/compute/instancetemplates/instancetemplate_reconcile.go
justinsb: Fixed - thanks!

cloud/services/compute/instancetemplates/instancetemplate_reconcile.go
justinsb: Thanks

cloud/services/compute/instancegroupmanagers/instancegroupmanagers_reconcile.go
justinsb: Fixed - thanks

cloud/services/compute/instancetemplates/instancetemplate_reconcile.go
justinsb: Added the full hash as a label.  We can at least detect it.  If it ever happens (it seems quite unlikely at 64 bits) we can add stronger comparison logic.

exp/controllers/gcpmachinepool_controller.go
justinsb: We need the machinePoolScope to be able to clean up.  (At the very least we need the cluster name).  So if the MachinePool is deleted first, we are in big trouble.  It feels like a design issue, but that's not expected for an experimental feature in CAPI, so do you want to open an issue?

cloud/scope/machinepool.go
salasberryfin: I added issue https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/1573 and we can address this in a follow-up PR to fix all other affected scope methods.

exp/controllers/gcpmachinepool_controller.go
salasberryfin: Shouldn't this case be controlled by the call to `reconcileDelete()`? Is it possible that `GCPMachinePool` is deleted first?

exp/api/v1beta1/conditions_consts.go
```diff
@@ -70,4 +76,22 @@ const (
...
+	MIGDeletionInProgress = "ManagedInstanceGroupDeletionInProgress"
+
+	// InstanceTemplateReadyCondition represents the status of an AWSMachinePool's associated Launch Template.
+	InstanceTemplateReadyCondition clusterv1.ConditionType = "InstanceTemplateReady"
+	// InstanceTemplateNotFoundReason is used when an associated Launch Template can't be found.
+	InstanceTemplateNotFoundReason = "InstanceTemplateNotFound"
+	// InstanceTemplateCreateFailedReason used for failures during Launch Template creation.
+	InstanceTemplateCreateFailedReason = "InstanceTemplateCreateFailed"
+	// InstanceTemplateReconcileFailedReason used for failures during Launch Template reconciliation.
+	InstanceTemplateReconcileFailedReason = "InstanceTemplateReconcileFailed"
```
salasberryfin: Looks like a few of these new conditions are not yet used:
`MIGNotFoundReason`
`InstanceTemplateNotFoundReason`
`InstanceTemplateCreateFailedReason`
Are we planning on using them down the line or can they just be deleted?

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,344 @@
...
+func (r *GCPMachinePoolReconciler) reconcileDelete(ctx context.Context, machinePoolScope *scope.MachinePoolScope) error {
+	log := log.FromContext(ctx)
+
+	log.Info("Handling deleted GCPMachinePool")
+
+	if err := instancegroupmanagers.New(machinePoolScope).Delete(ctx); err != nil {
+		log.Error(err, "Error deleting instanceGroupManager")
+		r.Recorder.Eventf(machinePoolScope.GCPMachinePool, corev1.EventTypeWarning, "FailedDelete", "Failed to delete instancegroupmanager: %v", err)
+
+		// record.Warnf(machineScope.GCPMachine, "GCPMachineReconcile", "Reconcile error - %v", err)
```
salasberryfin:
```suggestion
```

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,344 @@
...
+			Message: fmt.Sprintf("Error deleting instanceGroupManager: %v", err),
+		})
+		return err
+	}
+
+	if err := instancetemplates.New(machinePoolScope).Delete(ctx); err != nil {
+		log.Error(err, "Error deleting instanceTemplates")
+		r.Recorder.Eventf(machinePoolScope.GCPMachinePool, corev1.EventTypeWarning, "FailedDelete", "Failed to delete instance template: %v", err)
+
+		// record.Warnf(machineScope.GCPMachine, "GCPMachineReconcile", "Reconcile error - %v", err)
```
salasberryfin:
```suggestion
```

salasberryfin (changes requested):

cloud/services/compute/instancegroupmanagers/instancegroupmanagers_reconcile.go
damdo: Are you going to add the comment here or elsewhere? TY

exp/api/v1beta1/gcpmachinepool_types.go
```diff
@@ -0,0 +1,236 @@
...
+	// AdditionalMetadata is an optional set of metadata to add to an instance, in addition to the ones added by default by the
+	// GCP provider.
+	// +listType=map
+	// +listMapKey=key
+	// +optional
+	AdditionalMetadata []capg.MetadataItem `json:"additionalMetadata,omitempty"`
+
+	// // IAMInstanceProfile is a name of an IAM instance profile to assign to the instance
+	// // +optional
+	// // IAMInstanceProfile string `json:"iamInstanceProfile,omitempty"`
```
damdo: Looks like this is commented out, do we want to drop it?

exp/api/v1beta1/gcpmachinepool_types.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// Subnet is a reference to the subnetwork to use for this instance. If not specified,
+	// the first subnetwork retrieved from the Cluster Region and Network is picked.
+	// +optional
+	Subnet *string `json:"subnet,omitempty"`
+
+	// Not meaningful by MachinePool (?)
+	// // AliasIPRanges let you assign ranges of internal IP addresses as aliases to a VM's network interfaces.
+	// // +optional
+	// AliasIPRanges []infrav1.AliasIPRange `json:"aliasIPRanges,omitempty"`
```
damdo: Looks like this is commented out, do we want to drop it?

exp/api/v1beta1/gcpmachinepool_webhook.go
```diff
@@ -0,0 +1,83 @@
...
+func (*gcpMachinePoolWebhook) ValidateDelete(_ context.Context, obj runtime.Object) (admission.Warnings, error) {
+	r, ok := obj.(*GCPMachinePool)
+	if !ok {
+		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a GCPMachinePool but got a %T", obj))
+	}
+
+	gcpMachinePoolLog.Info("Validating GCPMachinePool delete", "name", r.Name)
+
+	return nil, nil
+}
```
damdo, line 89: Are we planning to add in the validation logic at a later stage?

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,344 @@
...
+	log := log.FromContext(ctx)
+
+	// Fetch the GCPMachinePool .
+	gcpMachinePool := &expinfrav1.GCPMachinePool{}
+	err := r.Client.Get(ctx, req.NamespacedName, gcpMachinePool)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
```
damdo: Personally I like error wrapping when possible. But that's a personal preference.

exp/controllers/gcpmachinepool_controller.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	// Fetch the CAPI MachinePool
+	machinePool, err := exputil.GetOwnerMachinePool(ctx, r.Client, gcpMachinePool.ObjectMeta)
+	if err != nil {
+		return reconcile.Result{}, err
```
damdo, line 86: Optionally wrap

exp/controllers/gcpmachinepool_controller.go
damdo: How are other providers that have implemented MachinePools doing this?

damdo (commented):
Thanks a lot Justin!
This is looking good, left a couple of comments but it looks like we are already in great shape! TY

exp/api/v1beta1/conditions_consts.go
justinsb: I can remove them.  They seem to be what others CAPI providers use, so I was thinking it was part of the implicit contract, but I guess if they are part of the contract they should be in CAPI itself, so we can just add them as needed

exp/api/v1beta1/gcpmachinepool_types.go
justinsb: Removed them.

exp/api/v1beta1/gcpmachinepool_types.go
justinsb: Removed

exp/api/v1beta1/gcpmachinepool_webhook.go
justinsb: I will add a few TODOs.  I think we probably want to share logic with the machine validation.  Does that work?

exp/controllers/gcpmachinepool_controller.go
justinsb: My rules of thumb:
* if it is a call to a third party API we should wrap.
* if it's a call to an internal method, we expect our code to return a good error and we don't need to wrap, unless we have some additional information to add that is not passed to the method.

So in this case, I added wrapping to the call to r.Client.Get (even though that's a kube library I still consider it "third party" here).  But I think if we want better information around errors from GetOwnerMachinePool we should add that inside GetOwnerMachinePool.

One day I want to try to use the additional information in the `log` context, it frustrates me that errors don't include that automatically!  But not in this PR :-)

exp/controllers/gcpmachinepool_controller.go
justinsb: So in this case, we should improve GetOwnerMachinePool if we want more info (IMO)

exp/controllers/gcpmachinepool_controller.go
justinsb: Same hack it looks like:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/0cdd2b16abbcaa02424d4993567b74f0956b92c2/exp/controllers/awsmachinepool_controller.go#L449-L457

cloud/services/compute/instancegroupmanagers/instancegroupmanagers_reconcile.go
justinsb: I added it here, but expanding it to include "cheap":

```
		// We fetch the created instanceGroupManager to return the complete object.
		// Also, we continue to go through the comparisons below so we have one code path,
		// (there are no API calls so this is cheap),
		// although in practice we don't expect to need further updates.
```

cloud/scope/machine_test.go
```diff
@@ -159,17 +124,7 @@ func TestInstanceNetworkInterfaceAliasIPRangesSpec(t *testing.T) {
...
-			Machine:    &testMachine,
-			GCPMachine: &testGCPMachine,
-		}
-
-		testMachineScope, err := NewMachineScope(testScopeParams)
-		assert.Nil(t, err)
-		assert.NotNil(t, testMachineScope)
-
-		result := testMachineScope.InstanceNetworkInterfaceAliasIPRangesSpec()
+		result := InstanceNetworkInterfaceAliasIPRangesSpec(testGCPMachine.Spec.AliasIPRanges)
```
justinsb, line 127: Self-reviewing.  If we want I could break this change (InstanceNetworkInterfaceAliasIPRangesSpec) out into a separate PR, to make this PR more clearly scoped only to MachinePool

exp/api/v1beta1/gcpmachinepool_webhook.go
damdo: Sounds good thanks!

exp/controllers/gcpmachinepool_controller.go
damdo: ack ty

cloud/scope/machine_test.go
damdo: Fine like this, IMO

exp/controllers/gcpmachinepool_controller.go
damdo: Got it TY. I think this is fine for now as we are experimental, but we can see if we can revisit this in the future :)

exp/controllers/gcpmachinepool_controller.go
damdo: DW, Fine as is

PR #1507, opened 2025-08-16
chore: refactor InstanceAdditionalDiskSpec for reuse in MachinePool

justinsb:
```release-note

NONE
```

### DIFF
--- a/cloud/scope/machine_test.go
+++ b/cloud/scope/machine_test.go
@@ -1,6 +1,7 @@
 package scope
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -12,6 +13,8 @@ import (
 // This test verifies that if a user selects "local-ssd"
 // as a disk type then the MachineScope correctly detects it as such.
 func TestMachineLocalSSDDiskType(t *testing.T) {
+	ctx := context.Background()
+
 	// Register the GCPMachine and GCPMachineList in a schema.
 	schema, err := infrav1.SchemeBuilder.Register(&infrav1.GCPMachine{}, &infrav1.GCPMachineList{}).Build()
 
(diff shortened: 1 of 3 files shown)


cloud/scope/machine_test.go
```diff
@@ -12,6 +13,8 @@ import (
 // This test verifies that if a user selects "local-ssd"
 // as a disk type then the MachineScope correctly detects it as such.
 func TestMachineLocalSSDDiskType(t *testing.T) {
+	ctx := context.Background()
```
cpanato, line 16: should we use the T.Context() here?

cpanato (commented):
A clarification question: Is there any impact on the existing running clusters? When the user upgrades to a version that has this change

cloud/scope/machine_test.go
justinsb: Oh wow!  I have been wanting this for a long time.  I had missed that it was introduced - but it was introduced in 1.24.

I think we're pinned to go 1.23 in go.mod, but it gets auto-bumped if we move to CAPI 1.11 (#1509)

We can also bump the go toolchain in go.mod - do we want to do that?

(I totally agree that when we bump to go 1.24 we should use T.Context here and probably in many other places!)

justinsb:
Trying to track down why that apidiff job is hanging ... hopefully it is just a networking flake pulling go dependencies, so I have proposed a timeout in https://github.com/kubernetes/test-infra/pull/35469 (and thank you @damdo for approving that!).  But of course I did touch this recently, so it may need a deeper fix ... TBD

Assuming I can get that test passing, I _think_ that this one should not block on the CAPI update (and thus nor should it block on T.Context), because it is in support of MachinePool, not the CAPI upgrade.  I am assuming we will support MachinePool before CAPI 1.11, because they are logically separate, but happy to discuss that assumption: I'd rather get MachinePool in sooner, but it also looks like we can bump to CAPI 1.11 without _too_ much trouble (I'm working on that PR next!)

cloud/scope/machine_test.go
damdo: Agree let's defer that after we get CAPI 1.11 and go 1.24 in

damdo (commented):
So this is a pre-factor right?

/approve
/lgtm

/hold
Feel free to unhold when ready to merge it

PR #1509, opened 2025-08-23
Upgrade to CAPI v1.11

justinsb:
We upgrade CAPI libraries to 1.11.0, however we continue to use v1beta1 semantics and versions (except in our tests).

This should be an internal version bump, not user facing.

```release-note
Upgrade to go 1.24, k8s v1.33, CAPI v1.11, controller-runtime v0.21 
```


pkg/capiutils/predicates.go
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
cpanato:
```suggestion
Copyright 2025 The Kubernetes Authors.
```

pkg/capiutils/utils.go
```diff
@@ -0,0 +1,99 @@
+package capiutils
```
cpanato, line 17: missing boilerplate

go.mod
```diff
@@ -1,6 +1,8 @@
 module sigs.k8s.io/cluster-api-provider-gcp
 
-go 1.23.7
+go 1.24.0
```
cpanato, line 3: as this project is not used as a lib i would use 1.24.6 directly

justinsb: Done - annoyingly (?) go mod tidy then drops the toolchain directive, so that is why that is now missing.

pkg/capiutils/utils.go
justinsb: Added!

justinsb:
> Thanks @justinsb.
> 
> Looks like this bump also requires a number of changes to remove deprecated functions. Let me know if I can help collaborating on this PR.

Thank you @salasberryfin - any collaboration would be wonderful.  I propose keeping this PR fairly minimal - just bumping to CAPG 1.11.0 but not yet adopting v1beta2 APIs.

I think the deprecated functions are ones that I introduced and marked deprecated!  They are needed (temporarily) while we continue to use v1beta1 APIs internally.  Moving everything to v1beta2 would (I think) be a more invasive code change (I'll happily review though if you want to add that - I think you need to bump CAPI to 1.11 first which is what I'm trying to do in this PR).  The CAPI v1beta1 packages are marked deprecated also, and yes - we need to move to v1beta2.

go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/cluster-api-provider-gcp
 
-go 1.23.7
+go 1.24.6
```
damdo: I'd prefer we did bump the minimum required minor version, but not patch version.
This has caused a LOT of pain downstream, and not only in our product. (i.e. just doing `go 1.24.0` instead of `1.24.z`)

Core CAPI also has my same take for CAPI v1.11, as you can see [here](https://github.com/kubernetes-sigs/cluster-api/blob/a68028a33c1769865e671dbb3c61b33243b59bea/go.mod#L3)

hack/tools/go.mod
```diff
@@ -1,37 +1,40 @@
 module sigs.k8s.io/cluster-api-provider-gcp/hack/tools
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.24.2
+toolchain go1.24.6
```
damdo, line 5: This I am happy with

test/e2e/config/gcp-ci.yaml
```diff
@@ -78,16 +78,16 @@ providers:
...
 variables:
-  KUBERNETES_VERSION: "v1.32.5"
-  KUBERNETES_VERSION_MANAGEMENT: "v1.32.5"
+  KUBERNETES_VERSION: "v1.33.2"
+  KUBERNETES_VERSION_MANAGEMENT: "v1.33.2"
   ETCD_VERSION_UPGRADE_TO: "3.5.16-0"
   COREDNS_VERSION_UPGRADE_TO: "v1.11.3"
   KUBERNETES_IMAGE_UPGRADE_FROM: "projects/k8s-staging-cluster-api-gcp/global/images/cluster-api-ubuntu-2204-v1-32-5-nightly"
-  KUBERNETES_IMAGE_UPGRADE_TO: "projects/k8s-staging-cluster-api-gcp/global/images/cluster-api-ubuntu-2204-v1-31-0-nightly"
+  KUBERNETES_IMAGE_UPGRADE_TO: "projects/k8s-staging-cluster-api-gcp/global/images/cluster-api-ubuntu-2204-v1-33-2-nightly"
```
damdo: Is this available already?

test/e2e/data/shared/v1beta1/metadata.yaml
```diff
@@ -1,6 +1,9 @@
 apiVersion: clusterctl.cluster.x-k8s.io/v1alpha3
 kind: Metadata
 releaseSeries:
+  - major: 1
+    minor: 11
+    contract: v1beta1
```
damdo, line 6: Great, although I think we are missing the `/metadata.yaml` bump: https://github.com/justinsb/cluster-api-provider-gcp/blob/b1d9c054273c19a90f1fe2f71411fe2fdb1fb8dd/metadata.yaml#L45-L46

go.mod
justinsb: Would be great to figure this out, that's why I sent #1515 because I'm confused!  Do you have any links to the problems caused?  I remember a problem a few years ago in k/k where we found a regression in go on arm ~x86~ 32 bit (IIRC).

That said, I do think we should be using toolchain directives

```
go 1.24

toolchain go1.24.6
```

On other projects users insist that we run the latest golang because otherwise their vulnerability scanners flag it.  I can see a case for not running 1.25.0 while 1.24.x is supported, but if we're using 1.24.x I think we want to be using 1.24.6 to avoid getting flagged by every security team.

hack/tools/go.mod
justinsb: Saw this after I wrote the above reply for the top-level go.mod?  Can I apply the same there?

test/e2e/config/gcp-ci.yaml
justinsb: I've discovered it isn't :-). I bumped down to 1.33.2 so 🤞

I guess technically this could be done separately - I do think we should try to figure out whether we want to bump CAPI minors in one big PR.  I'm trying not to so that the PR is relatively small, but it is expanding :-). We should move to use v1beta2 CAPI APIs and we should move our CRDs to the v1beta2 contract, but if I did that in one PR then ... I wouldn't want to review it, personally :-)

test/e2e/data/shared/v1beta1/metadata.yaml
justinsb: I mentioned it here: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1509#issuecomment-3236326815

I worry about this ("update this file only when a new major or minor version is released"): https://github.com/justinsb/cluster-api-provider-gcp/blob/b1d9c054273c19a90f1fe2f71411fe2fdb1fb8dd/metadata.yaml#L5

Also, that I would have to say `contract: v1beta1` but I _think_ that we want to try to get to `contract: v1beta2` (or should at least discuss when we want to do that!)

go.mod
damdo: I am happy with using the `go toolchain xx` directive bumped to the latest patch version of the version specified in the `go xx` directive (staying to a .0 patch release for the whole time).

> Do you have any links to the problems caused?

The problem with bumping the patch version of the `go xx` directive is that it forces all importers of this module to use that same or a higher patch version to build it.

Which it is fine in most of the projects if they can keep up with the go patch release cadence, but it is hard in environments/projects where the go compiler is bumped once every minor go release, or at least where the bumping cadence is not that high for _reasons_. 
Downstream we explicitly disable toolchain for achieving reproducible/consistent builds, so we can't rely on the go compiler "dynamic version upgrade" feature (or whatever it is called 😄)

hack/tools/go.mod
damdo: Sure yes!

test/e2e/data/shared/v1beta1/metadata.yaml
justinsb: It looks like I have to set it for the tests to pass so this might be moot!

go.mod
justinsb: Agreed, I like that approach!

justinsb:
It _looks_ like I can't upgrade kubernetes versions, I don't think we support 1.33 yet.  I _think_ this is unrelated to the CAPI bump, and is related to the cloudprovider extraction and deprecation of some kubelet flags.  I guess we should deal with that separately (?)

damdo:
> The tests are now passing (apart from apidiff, because we are changing the API). But this is testing with kube 1.32, not kube 1.33. I believe https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1526 is the problem, but https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1527 will also show that there's another problem where we can't support 1.33 without upgrading to CAPI 1.11 first...

@justinsb thanks for looking into these. IMO the best path forward would be to first merge 1526 and then rebase this PR to use the 1526 fix and try using k8s 1.33 again. So we have everything in here.

justinsb:
> > The tests are now passing (apart from apidiff, because we are changing the API). But this is testing with kube 1.32, not kube 1.33. I believe #1526 is the problem, but #1527 will also show that there's another problem where we can't support 1.33 without upgrading to CAPI 1.11 first...
> 
> @justinsb thanks for looking into these. IMO the best path forward would be to first merge 1526 and then rebase this PR to use the 1526 fix and try using k8s 1.33 again. So we have everything in here.

That sounds like a great path - I'll try to get at least #1526 merged, and I'll see what's going on with #1527 as I do think there's one last surprise for me!

PR #1515, opened 2025-08-28
docs: document branching strategy

justinsb:
Documenting our branches and branch "philosophy"

```release-note

NONE
```

### DIFF
--- a/docs/book/src/developers/releasing.md
+++ b/docs/book/src/developers/releasing.md
@@ -63,6 +63,23 @@ Example versions:
 - Patch release: `v0.1.1`
 - Major release: `v1.0.0`
 
+The major and minor versions mirror the version of Cluster-API.  Thus CAPG 1.11.x releases will use CAPI versions 1.11
+
+The patch version is incremented for each release.  Thus the first release of CAPG using CAPI 1.11 will 1.11.0, the second will be 1.11.1, etc
+
+## Branches
+
+We maintain release branches to facilitate backporting fixes.  These branches are named `release-<major>.<minor>`.  Patch releases are tagged from the corresponding release branch.
+
+Feature development happens on the `main` branch, and [nightlies](nightlies.md) are released from the `main` branch also.
+While we aim to keep the `main` branch healthy, it is intended for development and testing, and should not be used in production.
+
+As such, we will aim to keep the main branch up to date with the latest versions of all dependencies (go toolchain, CAPI, other go module dependencies).
+
+Release branches lock to a particular minor version of CAPI, but will update other dependencies on these branches while they are supported.
+Currently we maintain only one release branch (and the `main` branch).  Once a branch is no longer supported there will be no new releases
+of the corresponding minor, including no further security fixes.
+
 ## Expected artifacts
 
 1. A release yaml file `infrastructure-components.yaml` containing the resources needed to deploy to Kubernetes


docs/book/src/developers/releasing.md
```diff
@@ -63,6 +63,23 @@ Example versions:
...
+The patch version is incremented for each release.  Thus the first release of CAPG using CAPI 1.11 will 1.11.0, the second will be 1.11.1, etc
+
+## Branches
+
+We maintain release branches to facilitate backporting fixes.  These branches are named `release-<major>.<minor>`.  Patch releases are tagged from the corresponding release branch.
+
+Feature development happens on the `main` branch, and [nightlies](nightlies.md) are released from the `main` branch also.
+While we aim to keep the `main` branch healthy, it is intended for development and testing, and should not be used in production.
+
+As such, we will aim to keep the main branch up to date with the latest versions of all dependencies (go toolchain, CAPI, other go module dependencies).
```
justinsb, line 77: I think this is right, we should basically auto-bump everything on main

docs/book/src/developers/releasing.md
```diff
@@ -63,6 +63,23 @@ Example versions:
...
+
+Release branches lock to a particular minor version of CAPI, but will update other dependencies on these branches while they are supported.
```
justinsb, line 79: It sounds like we have not updated go versions on existing branches, but I think we should

damdo: You mean _patch_ level Go version bumps right?

docs/book/src/developers/releasing.md
damdo: I think we mostly already do with the dependabot setup we have. Of course we have some constraints as CAPI is normally couple with one specific k8s version and one specific go version, and we try and keep that the same in CAPI providers

docs/book/src/developers/releasing.md
```diff
@@ -63,6 +63,23 @@ Example versions:
 - Patch release: `v0.1.1`
 - Major release: `v1.0.0`
 
+The major and minor versions mirror the version of Cluster-API.  Thus CAPG 1.11.x releases will use CAPI versions 1.11
+
+The patch version is incremented for each release.  Thus the first release of CAPG using CAPI 1.11 will 1.11.0, the second will be 1.11.1, etc
```
damdo, line 68: That has been true for now, but there might come a time where we need to release CAPG minor versions more frequently than CAPI minor releases.

This what many of the main CAPI providers out there already do. For example, to my knowledge, CAPA, CAPV, CAPZ don't have this CAPI to CAPI Provider minor version roping, and can and may release multiple provider versions in between core CAPI minor releases.

docs/book/src/developers/releasing.md
justinsb: So I think it is something we should decide.  I think we want to be running a supported version of golang at all times, and I think we want to bump minors if our minor is no longer supported.

Should we say something like "for the go toolchain, we will bump to the latest supported patch within the same minor; we will bump to the latest minor if our minor goes unsupported.

(I've realized I can also say "particular minor version of CAPI - and thus of kubernetes dependencies including controller-runtime")

It might be that we've simply relied on CAPI bumping go toolchain versions in the past, and we want to continue to do that.  If that's our intent - if we just want to follow CAPI versions - then I can document that instead...

docs/book/src/developers/releasing.md
justinsb: I'll try to find some wording that makes it clear that we have this property today, but that we might not always have it in future.  (It's a really human friendly property IMO, not having to constantly remember the translation factor!)

docs/book/src/developers/releasing.md
justinsb: Agreed, I'm hoping we can write down our intent and then use it for automation - both on doing the bumps and perhaps on reviewing PRs.  It comes up in particular on go version bumps: I think that on the main branch we should bump go toolchains automatically, to the latest, even if that puts us onto a new minor or ahead of CAPI.  Sometimes there will be e.g. new `go vet` primitives and the PR won't merge without other changes, sometimes we might find a go regression (in theory).

I can see an alternative strategy where we bump main to the latest _patch_ versions within a minor, matching CAPI on the minor.  (And for any libraries which aren't in CAPI, like say a GCP library we *do* bump minors).  I think that would work as well: CAPI is pretty prompt at bumping.

docs/book/src/developers/releasing.md
damdo: SGTM thanks Justin!

docs/book/src/developers/releasing.md
justinsb: (Edited the above to fix incomplete thought!)

damdo: I think this could quickly expand to a wider discussion on how many releases we support and what is our release cadence :) (which we might want to have at some point, maybe in a CAPG Community call)

I am happy either way 
1) with us bumping the `go toolchain xx` directive and keep that up to date with the latest supported patch release, as long as we don't touch the `go xx` directive, as that's the minimum required version that can be used to build the project and I wouldn't want that to change
2) Follow what core CAPI does

My personal preference would be to follow core CAPI as closely as possible. They spent already a significant about of time oiling and perfecting the cadence and process and I think they are now at a point where folks are happy with it and it ties in well with the Go and the K8s processes. Also I think that's what the community expects anyway, and also that means we are always well compatible with that.

docs/book/src/developers/releasing.md
damdo: I think I covered this in https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1515#discussion_r2309691771

I'd personally strive to follow CAPI as close as possible.

Also I am happy with
> strategy where we bump main to the latest patch versions within a minor, matching CAPI on the minor. (And for any libraries which aren't in CAPI, like say a GCP library we do bump minors). I think that would work as well: CAPI is pretty prompt at bumping

docs/book/src/developers/releasing.md
salasberryfin: I think we've been able to follow this until now because we haven't released that often, which meant we could just wait for a CAPI bump to create a new minor release. As we get more people involved with the project, I'd vote for increasing our release cadence. In this case, it will eventually force us to separate CAPG minor versions from CAPI's.

salasberryfin (commented):
Thanks @justinsb!

docs/book/src/developers/releasing.md
damdo: We discussed this also in the office hours and we are leaning towards giving us more flexibility for the future and not commit to strict numbering going forward. cc. @cpanato too

chrischdi: Note: I just want to share some knowledge, no hard opinions on what to do for CAPG :-)

Note: CAPV today also ~ follows CAPI release versions, but nowadays with an offset of 3 minors, and it always uses the `.1` CAPI release to include compatibility to the next k8s release.

IMHO it makes sense to follow the CAPI release cadence, which helps for planning but of course its not set in stone and sometimes a project wants to have an additional minor due to reasons.

Following CAPI's release cadence might be challenging though and could require to test early for the next release.

At the end CAPI's release cadence (of x.x.1 relases) follows k8s release cadence.

docs/book/src/developers/releasing.md
chrischdi: CAPI ~ follows its k8s dependency regarding to the go version. This does not mean a provider has to. 

> Note: If the Go minor version of one of our supported branches goes out of support, we should consider bumping to a newer Go minor version according to our backport policy. Usually this is done after k/k bumps the Go version in their release branches.
> 
> The right Go minor version for a Cluster API branch can be looked up as follows (based on an example):
> 
> Cluster API v1.10 has Kubernetes v1.32 as a dependency: https://github.com/kubernetes-sigs/cluster-api/blob/v1.10.0/go.mod
> Kubernetes v1.32 uses Go 1.23.8 (can be seen by checking the corresponding kubekins image in https://github.com/kubernetes/test-infra/blob/1a5662a/images/kubekins-e2e-v2/variants.yaml#L19)
> => Cluster API v1.10 can be bumped to Go 1.23


source: https://github.com/kubernetes-sigs/cluster-api/tree/main/docs/release/role-handbooks/release-lead#continuously-bump-the-go-version

docs/book/src/developers/releasing.md
chrischdi: In core CAPI we use trivy & govulncheck (via github action) to get notified on CVE's for the release branches and bump these (if required and viable) on CVE's to include the fixes in (potential) patch releases :-)

docs/book/src/developers/releasing.md
damdo: Thanks for the detail @chrischdi 
Good to know it is not a mandatory requirement 😄 
I'd still like to see them going hand in hand with respect to that, to avoid unwanted incompatibilities and surprises 😬

docs/book/src/developers/releasing.md
damdo: We could think about doing that here too yeah.
I think we need to understand how that overlaps with dependabot.
Also we might be able to configure dependabot in a way so it bumps the `go toolchain` versions to higher patch versions when needed but without touching the main `go` directive.. 🤔 I know @cpanato is an expert on these things

salasberryfin: I'd also be in favor of following core CAPI. I think it makes the project easier to maintain long term as we implement a process that has been proved to work and don't have to focus on describing (and applying) our own.

It also means that future developers/maintainers that are new to CAPG but have experience in the CAPI ecosystem will have an easier time figuring out the intricacies of the project.

salasberryfin (commented):
Should we revive this?

cc @justinsb @damdo

damdo:
Yes let's! 👍

docs/book/src/developers/releasing.md
```diff
@@ -63,6 +63,23 @@ Example versions:
...
+We maintain release branches to facilitate backporting fixes.  These branches are named `release-<major>.<minor>`.  Patch releases are tagged from the corresponding release branch.
+
+Feature development happens on the `main` branch, and [nightlies](nightlies.md) are released from the `main` branch also.
+While we aim to keep the `main` branch healthy, it is intended for development and testing, and should not be used in production.
+
+As such, we will aim to keep the main branch up to date with the latest versions of all dependencies (go toolchain, CAPI, other go module dependencies).
+
+Release branches lock to a particular minor version of CAPI, but will update other dependencies on these branches while they are supported.
+Currently we maintain only one release branch (and the `main` branch).  Once a branch is no longer supported there will be no new releases
+of the corresponding minor, including no further security fixes.
```
damdo, line 81: @justinsb Can we have a section mentioning our support window? To minor releases. So then we can link it up here where we say `Once a branch is no longer supported`. TY!

damdo:
/assign

to finish this off and merge

PR #1516, opened 2025-08-29
chore: don't use Requeue in controller-runtime

justinsb:
It is deprecated in newer versions of controller-runtime,
because it should not be used.

```release-note

NONE
```


justinsb:
We're going to need this to bump controller-runtime, but we might want to cherry-pick to 1.10 (and it makes #1509 smaller)

salasberryfin:
Thanks @justinsb 

/lgtm

/cherrypick release-1.10

PR #1517, opened 2025-08-29
chore: fix go vet errors around conditions.MarkFalse

justinsb:
Newer go versions flag this (non-constant used as format message).  As with #1516 we may want to cherry-pick (?), and it makes #1509 smaller

```release-note
chore: fix go vet errors around conditions.MarkFalse
```

(Technically the messages do change, and I could replace it with a "%s" or "%v" if we want the same messages.  I think this is clearer and the message is not part of our contract, but I'm happy to switch to %s if desired)


salasberryfin:
I think any help with reducing the size of https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1509 is nice, so let's cherry-pick this one too.

/cherrypick release-1.10

PR #1526, opened 2025-09-05
templates: don't pass cloud-provider to kube-apiserver

justinsb:


templates/cluster-template-clusterclass.yaml
```diff
@@ -121,8 +121,8 @@ spec:
         clusterConfiguration:
           apiServer:
             timeoutForControlPlane: 20m
-            extraArgs:
-              cloud-provider: gce
+            #extraArgs:
```
cpanato: can we remove this instead of commenting if that is not required anymore?

justinsb:
Changed comments to removal

damdo (commented):
/approve
/lgtm

PR #1527, opened 2025-09-05
tests: test with kubernetes 1.33.2

justinsb:
Test with kubernetes 1.33, which is notable because it drops support for in-tree cloudprovider.

```release-note

NONE
```


justinsb:
OK so this fails without #1506 https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-gcp/1527/pull-cluster-api-provider-gcp-e2e-test/1963803065215619072.

I have now cherry-picked #1506 onto this branch because I believe it should pass (🤞 )

/hold as this PR is now doing multiple things!

damdo:
Indeed thanks for the heads up

/unhold

PR #1531, opened 2025-09-17
Provide a MinPortsPerVM for Nat Router

barbacbd:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**

** This is a regression for the openshift installer.

Instead of filling this value in, the default will be 64 (so that this is backwards compatible), otherwise the user can provide this information in the cluster spec.

** MinPortsPerVM: Minimum number of ports allocated to a VM from this NAT config. If not set, a default number of ports is allocated to a VM. This is rounded up to the nearest power of 2. For example, if the value of this field is 50, at least 64 ports are allocated to a VM.

Fixes #

https://issues.redhat.com/browse/OCPBUGS-61876

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
MinPortsPerVM is the minimum number of ports allocated to a VM from this NAT config. If not set, a default number of ports is allocated to a VM. This is rounded up to the nearest power of 2. For example, if the value of this field is 50, at least 64 ports are allocated to a VM. 
```

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -147,6 +147,16 @@ type NetworkSpec struct {
 	// +kubebuilder:default:=1460
 	// +optional
 	Mtu int64 `json:"mtu,omitempty"`
+
+	// MinPortsPerVM: Minimum number of ports allocated to a VM from this NAT
+	// config. If not set, a default number of ports is allocated to a VM. This is
+	// rounded up to the nearest power of 2. For example, if the value of this
+	// field is 50, at least 64 ports are allocated to a VM.
+	// +kubebuilder:validation:Minimum:=2
+	// +kubebuilder:validation:Maximum:=65536
+	// +kubebuilder:default:=64
+	// +optional
+	MinPortsPerVM int64 `json:"minPortsPerVm,omitempty"`
 }
 
 // LoadBalancerType defines the Load Balancer that should be created.
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -237,6 +237,7 @@ func (s *ClusterScope) NatRouterSpec() *compute.Router {
 				Name:                          fmt.Sprintf("%s-%s", networkSpec.Name, "nat"),
 				NatIpAllocateOption:           "AUTO_ONLY",
 				SourceSubnetworkIpRangesToNat: "ALL_SUBNETWORKS_ALL_IP_RANGES",
+				MinPortsPerVm:                 s.GCPCluster.Spec.Network.MinPortsPerVM,
 			},
 		},
 	}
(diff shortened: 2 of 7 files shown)


api/v1beta1/types.go
```diff
@@ -147,6 +147,15 @@ type NetworkSpec struct {
 	// +kubebuilder:default:=1460
 	// +optional
 	Mtu int64 `json:"mtu,omitempty"`
+
+	// MinPortsPerVM: Minimum number of ports allocated to a VM from this NAT
+	// config. If not set, a default number of ports is allocated to a VM. This is
+	// rounded up to the nearest power of 2. For example, if the value of this
+	// field is 50, at least 64 ports are allocated to a VM.
+	// +kubebuilder:validation:Minimum:=64
```
damdo: @barbacbd Do we need to set this minimum value?
Wouldn't ~`0`~ `1` for example be a valid number for the internal GCP platform validation?

barbacbd: @damdo 2^0 would be 1 right? Which would increase to 64. The minimum allowed by GCP is 64 I believe.

damdo: Why would it increase to 64? If by definition 1 is POW 2, that should be ok right?
Might be worth trying. If GCP errors with less than 64, then we can set the Minimum validation like you did :)

damdo: TY

api/v1beta1/types.go
```diff
@@ -147,6 +147,16 @@ type NetworkSpec struct {
 	// +kubebuilder:default:=1460
 	// +optional
 	Mtu int64 `json:"mtu,omitempty"`
+
+	// MinPortsPerVM: Minimum number of ports allocated to a VM from this NAT
+	// config. If not set, a default number of ports is allocated to a VM. This is
```
justinsb, line 152: Nit: we might want to specify that this is configured on the Nat Router, I wasn't entirely sure what this doing otherwise.  But it's only a comment, so not a blocker IMO.

cloud/scope/cluster.go
```diff
@@ -237,6 +237,7 @@ func (s *ClusterScope) NatRouterSpec() *compute.Router {
 				Name:                          fmt.Sprintf("%s-%s", networkSpec.Name, "nat"),
 				NatIpAllocateOption:           "AUTO_ONLY",
 				SourceSubnetworkIpRangesToNat: "ALL_SUBNETWORKS_ALL_IP_RANGES",
+				MinPortsPerVm:                 s.GCPCluster.Spec.Network.MinPortsPerVM,
```
justinsb, line 240: So how does defaulting work here with existing objects?  I guess worst-case (default not applied) the code does the same thing it did before (and 64 is also the default on the NatRouter, I believe, so we shouldn't be changing the actual value in GCP either)

justinsb:
/test pull-cluster-api-provider-gcp-apidiff

I worry that I've broken our apidiff test - or at least broken it printing nice error messages...

PR #1532, opened 2025-09-18
CORS-4230: Add SkipFirewallRuleCreation

barbacbd:
**What type of PR is this?**

/kind feature
/kind api-change

Add API changes to SkipFirewallRuleCreation. When true, the firewall rules will not be created. When this is the case, the firewall rules should exist prior to creating the network. This will allow ServiceAccounts to skip the rules: compute.firewalls.create.
The firewall service will no longer create firewall rules when the skip firewall rule creation is enabled OR when it is a shared vpc.

Fixes #

https://issues.redhat.com/browse/CORS-4230

**TODOs**:

- [ X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add API changes to SkipFirewallRuleCreation. When true, the firewall rules will not be created. When this is the case, the firewall rules should exist prior to creating the network. This will allow ServiceAccounts to skip the rules: compute.firewalls.create.

Update the services and interfaces. The firewall service will no longer create firewall rules when the skip firewall rule creation is enabled OR when it is a shared vpc.
```


damdo (commented):
/ok-to-test

api/v1beta1/types.go
```diff
@@ -137,6 +137,10 @@ type NetworkSpec struct {
 	// +optional
 	HostProject *string `json:"hostProject,omitempty"`
 
+	// SkipFirewallRuleCreation should be set to true when no firewall rules should be
+	// created by the provider.
+	SkipFirewallRuleCreation *bool `json:"skipFirewallRuleCreation,omitempty"`
+
```
damdo: How about creating a `Firewall` type that in the future can be expanded if necessary?
```go
// Firewall configuration.
// +optional
Firewall FirewallSpec `json:"firewall,omitempty"`
```

And for the inner struct maybe have a field called something along the lines of `RulesManagement`?
```go
// FirewallSpec contains configuration for the firewall.
type FirewallSpec struct {
	// RulesManagement determines the management policy for firewall rules.
	// "On": The controller will create and manage firewall rules.
	// "Off": The controller will not touch any firewall rules. If this is
	//        changed to "Off" after rules have been created, they will not be
	//        deleted.
	// Defaults to "On".
	// +optional
	// +kubebuilder:default:="On"
	RulesManagement RulesManagementPolicy `json:"rulesManagement,omitempty"`
}

// RulesManagementPolicy is a string enum type for managing firewall rules.
// +kubebuilder:validation:Enum=On;Off
type RulesManagementPolicy string

const (
	// RulesManagementOn indicates that the controller should create and manage
	// firewall rules. This is the default behavior.
	RulesManagementOn RulesManagementPolicy = "On"

	// RulesManagementOff indicates that the controller should not create or manage
	// any firewall rules. If rules already exist, they will be left as-is.
	RulesManagementOff RulesManagementPolicy = "Off"
)
```

damdo: cc. @JoelSpeed for a sanity check

JoelSpeed: Seems like a reasonable suggestion, though maybe you want to change it to be `rules: Managed | Unmanaged` so that in the future you could potentially add a `PartiallyManaged` or other third option to the enum

damdo (commented):
Thanks!

/approve

/assign @justinsb @salasberryfin @cpanato

cloud/services/compute/firewalls/reconcile.go
```diff
@@ -28,8 +28,8 @@ import (
 // Reconcile reconcile cluster firewall compoenents.
 func (s *Service) Reconcile(ctx context.Context) error {
 	log := log.FromContext(ctx)
-	if s.scope.IsSharedVpc() {
-		log.V(2).Info("Shared VPC enabled. Ignore Reconciling firewall resources")
+	if s.scope.SkipFirewallRuleCreation() {
```
justinsb, line 31: BTW I like this a lot, even if SkipFirewallRuleCreation just called IsSharedVpc, it is much more self-documenting to have self-descriptive functions like this and put the logic in those functions, rather than put the logic in the caller.

api/v1beta1/types.go
```diff
@@ -107,6 +107,33 @@ type Network struct {
...
+type FirewallSpec struct {
+	// RulesManagement determines the management policy for firewall rules.
+	// "Managed": The controller will create and manage firewall rules.
+	// "Unmanaged": The controller will not touch any firewall rules. If this is
+	//              changed to "Off" after rules have been created, they will not be
+	//              deleted.
+	// Defaults to "Managed".
+	// +optional
+	// +kubebuilder:default:="Managed"
+	RulesManagement RulesManagementPolicy `json:"rulesManagement,omitempty"`
```
justinsb: So one nit: do we want people to have to write `rulesManagement: Managed` or `rules: Managed`

I think @JoelSpeed suggested `rules`, and I like it better from the perspective of not repeating the word "managed", but OTOH if we do end up adding extensible rules, maybe we want to use `firewall.rules` for that?

barbacbd: @justinsb would you prefer `rules` or `policy` ?

barbacbd: @justinsb I do have a similar PR https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1538 that does introduce rules.

JoelSpeed: Thinking about ways this could go

* Users want no rules management at all, will DIY it all
* Users want the default rules only
* Users want only a custom set of rules
* Users want default and custom rules (?)

So for those four choice to be valid, we are going to want a discriminated union (IMO). That could look something like:

```
firewall:
  rules: DefaultManaged | Unmanaged | CustomManaged
  customManaged:
    defaultRules: Include | Exclude
    rules:
    - ...
```

Would we expect the custom rules to be on the same level as the rules/policy selector? If so, we wouldn't want the custom rules to be allowed when the rules are unmanaged

We would also need a way to say exclude the default rules still, in which case, you have an oddity of `rules: Managed` but `defaultRules: Exclude` and `customRules: []` meaning that it does nothing? We could validate around this but its less intuitive IMO than doing the DU approach described above

Thoughts?

api/v1beta1/types.go
```diff
@@ -107,6 +107,33 @@ type Network struct {
 	APIInternalForwardingRule *string `json:"apiInternalForwardingRule,omitempty"`
 }
 
+// FirewallSpec contains configuration for the firewall.
+type FirewallSpec struct {
+	// RulesManagement determines the management policy for firewall rules.
+	// "Managed": The controller will create and manage firewall rules.
+	// "Unmanaged": The controller will not touch any firewall rules. If this is
+	//              changed to "Off" after rules have been created, they will not be
+	//              deleted.
```
JoelSpeed: If you kubectl describe, does this formatting print out correctly?

api/v1beta1/types.go
```diff
@@ -137,6 +164,10 @@ type NetworkSpec struct {
 	// +optional
 	HostProject *string `json:"hostProject,omitempty"`
 
+	// Firewall configuration.
+	// +optional
+	Firewall FirewallSpec `json:"firewall,omitempty"`
```
JoelSpeed: This is a struct so needs `omitzero`, or to be a pointer. Is this project using go 1.24?

cloud/scope/managedcluster.go
```diff
@@ -129,6 +129,12 @@ func (s *ManagedClusterScope) NetworkProject() string {
 	return ptr.Deref(s.GCPManagedCluster.Spec.Network.HostProject, s.Project())
 }
 
+// SkipFirewallRuleCreation returns whether the spec indicates that firewall rules
+// should be created or not.
+func (s *ManagedClusterScope) SkipFirewallRuleCreation() bool {
+	return (s.GCPManagedCluster.Spec.Network.Firewall.RulesManagement == infrav1.RulesManagementUnmanaged) || s.IsSharedVpc()
```
JoelSpeed: How would you explain this relationship in the API documentation/validations?

barbacbd: There are no custom rules in this PR so this should be taken only with the information in this PR. If the installation includes a shared VPC or if the user has explicitly said firewall rules are unmanaged then the firewall rules should not be created. It is important to note that this will ONLY apply to the default rules.

api/v1beta1/types.go
barbacbd: It is using go 1.24 I believe.

api/v1beta1/types.go
barbacbd: Honestly I haven't tried. This is the same formatting someone used on a suggested entry of mine in the past.

api/v1beta1/types.go
barbacbd: I think we could keep a value specifically for the default rules that are included in CAPG
```
// DefaultRulesManagement is the policy that applies to the default rules only. When the management policy is set to Unmanaged or the installation is using a shared VPC, the default firewall rules will not be created. 
// Defaults to "Managed".
// +optional
// +kubebuilder:default:="Managed"
DefaultRulesManagement RulesManagementPolicy `json:"defaultRulesManagement,omitempty"`
```

We should also have a field (this is currently in another PR) for the custom firewall rules. This includes the firewall rules that a caller would like to have created by CAPG. I don't believe that this list needs a value of Managed or Unmanaged. When the caller provides any firewall rules in this field they should be considered managed. I don't believe that there is a need to fill in this field with firewall rules and then state that they want these to be managed or created. I think that this is a fair assumption since we have no other need for the firewall rules. If there was another reason for specifying the firewall rules that could make sense but we have no other use.
```
CustomFirewallRules []FirewallRule
```

api/v1beta1/types.go
JoelSpeed: Would you be able to test, I'm not sure it will actually be preserved and make end up looking ugly/weird

cloud/scope/managedcluster.go
JoelSpeed: I don't see anything in the API about not being compatible with shared VPC?

api/v1beta1/types.go
JoelSpeed: So this needs omitzero
```suggestion
	Firewall FirewallSpec `json:"firewall,omitempty,omitzero"`
```

api/v1beta1/types.go
JoelSpeed: We need to make a decision on the direction of this field.

I think as it currently stands, `rulesManagement` with a future `customRules` sibling isn't going to be obvious that the custom rules are always managed.

So we can either rename the rules management to `defaultRules` and then have `customRules` as separate, or go with the union approach I suggested. Personally I'd lean towards the union as I feel it is more declarative about the users intent

cloud/scope/managedcluster.go
JoelSpeed: We should update he godoc of the API to mention this

PR #1533, opened 2025-09-24
Wait for API pod to be healthy before registering control plane node in instance group

### LINKED ISSUE #1407: Internal Load Balancer Implementation W/Passthrough LB Does Not Appear to Support More Than Single Node Control Plane Cluster
/kind bug

The implementation for Internal Load Balancer uses Passthrough Network LB which causes issues while spinning up a control plane with more than one node.  When subsequent control plane nodes try to join they will fail during preflight checks due to the fact that it uses Direct Server Return and requests are routed to the same host without taking into account whether its a healthy backend as this is prior to static pods being deployed so nothing can respond to the requests.  I'm not 100% certain if this is a bug due to the implementation choice or if there is a work around for this issue however it seems like using an Interrnal Proxy Network Load Balancer is a better fit for this use case as it will only route to healthy backends which addresses this issue.

Reference in docs:
If the client VM is a backend VM of the load balancer, connections sent to the IP address of the load balancer's forwarding rule are always answered by the client/backend VM. This happens regardless of whether the backend VM is healthy. It happens for all traffic sent to the load balancer's IP address, not just traffic on the protocol and ports specified in the load balancer's internal forwarding rule.

Expected a control plane with more than one node to be provisioned successfully

**Environment:**
N/A This is implementation detail of feature rather then specific to a release

- Cluster-api version:
- Minikube/KIND version:

bochengchu:
/kind bug

Fixes #1407

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Wait for API pod to be healthy before registering control plane node in instance group
```

### DIFF
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -27,6 +27,7 @@ import (
 	"github.com/go-logr/logr"
 
 	"github.com/pkg/errors"
+
 	"golang.org/x/mod/semver"
 	"google.golang.org/api/compute/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -47,6 +48,7 @@ type MachineScopeParams struct {
 	ClusterGetter cloud.ClusterGetter
 	Machine       *clusterv1.Machine
 	GCPMachine    *infrav1.GCPMachine
+	IsFirst       bool
 }
 
 // NewMachineScope creates a new MachineScope from the supplied parameters.
@@ -73,6 +75,7 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
 		GCPMachine:    params.GCPMachine,
 		ClusterGetter: params.ClusterGetter,
 		patchHelper:   helper,
+		IsFirst:       params.IsFirst,
 	}, nil
 }
 
@@ -83,6 +86,7 @@ type MachineScope struct {
 	ClusterGetter cloud.ClusterGetter
 	Machine       *clusterv1.Machine
 	GCPMachine    *infrav1.GCPMachine
+	IsFirst       bool
 }
 
 // ANCHOR: MachineGetter
@@ -134,6 +138,21 @@ func (m *MachineScope) ControlPlaneGroupName() string {
 // IsControlPlane returns true if the machine is a control plane.
 func (m *MachineScope) IsControlPlane() bool {
 	return IsControlPlaneMachine(m.Machine)
+}
+
+// IsFirstMachine returns true if the machine is the first machine in the cluster.
+func (m *MachineScope) IsFirstMachine() bool {
+	return m.IsFirst
+}
+
+// IsAPIServerHealthy returns true if the machine's API server pod is healthy.
+func (m *MachineScope) IsAPIServerHealthy() bool {
+	for _, condition := range m.Machine.GetConditions() {
+		if condition.Type == "APIServerPodHealthy" && condition.Status == "True" {
+			return true
+		}
+	}
+	return false
 }
 
 // Role returns the machine role from the labels.
--- a/cloud/services/compute/instances/reconcile.go
+++ b/cloud/services/compute/instances/reconcile.go
@@ -89,8 +89,17 @@ func (s *Service) Reconcile(ctx context.Context) error {
 	s.scope.SetInstanceStatus(infrav1.InstanceStatus(instance.Status))
 
 	if s.scope.IsControlPlane() {
-		if err := s.registerControlPlaneInstance(ctx, instance); err != nil {
-			return err
+		// If the instance is part of the control plane, we need to ensure it's
+		// registered with the instance group. We only do this if the API server is healthy or if this is the first
+		// control plane machine. This prevents a hairpinning issue where a new control plane machine attempts to reach
+		// the API server via a load balancer that is not yet ready. The first control plane machine is handled specially
+		// by the kubeadm controller, so it can be added to the instance group immediately.
+		if s.scope.IsAPIServerHealthy() || s.scope.IsFirstMachine() {
+			if err := s.registerControlPlaneInstance(ctx, instance); err != nil {
+				return err
+			}
+		} else {
+			log.Info("Waiting for API server to be healthy before registering control plane instance in instance group")
 		}
 	}
 
--- a/controllers/gcpmachine_controller.go
+++ b/controllers/gcpmachine_controller.go
@@ -178,6 +178,13 @@ func (r *GCPMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, nil
 	}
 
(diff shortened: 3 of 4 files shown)


cloud/services/compute/instances/reconcile.go
```diff
@@ -89,8 +89,13 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
 	if s.scope.IsControlPlane() {
-		if err := s.registerControlPlaneInstance(ctx, instance); err != nil {
-			return err
+		if s.scope.IsMachineProvisioned() {
+			log.V(2).Info("[DEBUG] Registering control plane instance in the instancegroup", "name", instance.Name)
+			if err := s.registerControlPlaneInstance(ctx, instance); err != nil {
+				return err
+			}
+		} else {
+			log.V(2).Info("[DEBUG] Skipping registering control plane instance in the instancegroup because machine is not yet provisioned", "name", instance.Name)
```
justinsb: Some comments with more background on the problem would be helpful for our future selves I think!

controllers/gcpmachine_controller.go
```diff
@@ -63,6 +66,33 @@ func (r *GCPMachineReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 		Watches(
 			&clusterv1.Machine{},
 			handler.EnqueueRequestsFromMapFunc(util.MachineToInfrastructureMapFunc(infrav1.GroupVersion.WithKind("GCPMachine"))),
+			builder.WithPredicates(predicate.Funcs{
```
justinsb: I don't _think_ this is necessary?

controllers/gcpmachine_controller.go
```diff
@@ -224,7 +254,8 @@ func (r *GCPMachineReconciler) reconcile(ctx context.Context, machineScope *scop
 		return ctrl.Result{}, err
 	}
 
-	if err := instances.New(machineScope).Reconcile(ctx); err != nil {
+	instancesSvc := instances.New(machineScope)
+	if err := instancesSvc.Reconcile(ctx); err != nil {
```
justinsb: Similarly I don't think this is important to this PR.

justinsb:
Nice catch @bochengchu - I think having a test showing the problem would be awesome (I guess in another PR, and we can figure out the sequencing of merging it).  And some comments on the critical line when we don't add the new control plane node until it is ready would be helpful for others looking to understand the code / the issue.

/ok-to-test

cloud/scope/machine.go
```diff
@@ -140,6 +144,24 @@ func (m *MachineScope) IsControlPlane() bool {
...
+	return m.IsFirst
+}
+
+// IsAPIServerHealthy returns true if the machine's API server pod is healthy.
+func (m *MachineScope) IsAPIServerHealthy() bool {
+	if m.Machine.Status.V1Beta2 == nil {
+		return false
+	}
+	for _, condition := range m.Machine.Status.V1Beta2.Conditions {
+		if condition.Type == "APIServerPodHealthy" && condition.Status == "True" {
```
bochengchu, line 151: I think we should look for APIServerPodHealthy to be True here, but this is a v1beta2 condition and v1beta2 seems not yet fully migrated to in this repo, so I hardcoded here. Maybe we can have follow-up work to bring in v1beta2 and update this there?

salasberryfin (commented):
Thanks @bochengchu. Can you please add a release note in the PR description?

controllers/gcpmachine_controller.go
```diff
@@ -178,6 +178,13 @@ func (r *GCPMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
+	// List all machines in the cluster to check if this is the first one.
+	machineList := &clusterv1.MachineList{}
+	if err := r.List(ctx, machineList, client.InNamespace(cluster.Namespace), client.MatchingLabels{clusterv1.ClusterNameLabel: cluster.Name}); err != nil {
```
egsysoev, line 183: If you create the KCP and MachineDeployment at the same time, it won’t work as expected. You should add a filter to select only control plane machines.

barbacbd: +1 to this. The check will pull all machines.

controllers/gcpmachine_controller.go
```diff
@@ -194,6 +201,7 @@ func (r *GCPMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		Machine:       machine,
 		GCPMachine:    gcpMachine,
 		ClusterGetter: clusterScope,
+		IsFirst:       len(machineList.Items) == 1,
```
barbacbd, line 204: IsFirst is a point-in-time snapshot that can race (controllers/gcpmachine_controller.go)

  len(machineList.Items) == 1 is evaluated at reconcile time. If two control plane machines are created in rapid succession (e.g., KCP scaling from 0→3), both could see len == 1 and both
  skip the health check. Or more likely: the second machine's reconcile runs before the first's Machine object is fully created, also seeing len == 1. This is racy.

  A more robust approach: check if the control plane endpoint is already populated (cluster.Spec.ControlPlaneEndpoint.Host != ""), which is set after the first CP node bootstraps. Or check
  if the Machine has status.phase == Running and owns the init lock (KCP sets an annotation for this).

cloud/services/compute/instances/reconcile.go
```diff
@@ -89,8 +89,17 @@ func (s *Service) Reconcile(ctx context.Context) error {
...
+		// registered with the instance group. We only do this if the API server is healthy or if this is the first
+		// control plane machine. This prevents a hairpinning issue where a new control plane machine attempts to reach
+		// the API server via a load balancer that is not yet ready. The first control plane machine is handled specially
+		// by the kubeadm controller, so it can be added to the instance group immediately.
+		if s.scope.IsAPIServerHealthy() || s.scope.IsFirstMachine() {
+			if err := s.registerControlPlaneInstance(ctx, instance); err != nil {
+				return err
+			}
+		} else {
+			log.Info("Waiting for API server to be healthy before registering control plane instance in instance group")
```
barbacbd, line 102: When the API server isn't healthy yet, the reconciler logs and returns nil (success). This means the machine won't be re-reconciled until something else triggers a watch event. If no
  event arrives, the machine stays permanently unregistered. This should return ctrl.Result{RequeueAfter: 30 * time.Second} or similar, but the Reconcile method returns error not
  ctrl.Result — so the requeue needs to happen at the controller level (gcpmachine_controller.go), not in the service.

  As-is, the controller at gcpmachine_controller.go calls the instance service reconciler and doesn't know whether registration was deferred, so it can't requeue either. The service would
  need to signal this (e.g., return a sentinel error or a result type).

PR #1536, opened 2025-09-25
Add an e2e test for only creating an internal load balancer

bochengchu:
**What type of PR is this?**
/kind other

Currently we only have e2e tests for `loadBalancerType` `External` and `ExternalInternal`.

Local test log: [log.txt](https://github.com/user-attachments/files/23308613/e2e_1cp_log.txt)

**Release note**:
None

### DIFF
--- a/test/e2e/config/gcp-ci.yaml
+++ b/test/e2e/config/gcp-ci.yaml
@@ -74,6 +74,7 @@ providers:
       - sourcePath: "${PWD}/test/e2e/data/infrastructure-gcp/cluster-template-ci-gke.yaml"
       - sourcePath: "${PWD}/test/e2e/data/infrastructure-gcp/cluster-template-ci-gke-autopilot.yaml"
       - sourcePath: "${PWD}/test/e2e/data/infrastructure-gcp/cluster-template-ci-gke-custom-subnet.yaml"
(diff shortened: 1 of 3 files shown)


test/e2e/data/infrastructure-gcp/cluster-template-ci-with-internal-lb.yaml
```diff
@@ -33,8 +34,10 @@ spec:
         cidrBlock: "10.0.0.0/17"
         purpose: PRIVATE
         region: us-east4
+  controlPlaneEndpoint:
+    port: 6443
   loadBalancer:
-    loadBalancerType: InternalExternal
+    loadBalancerType: Internal
```
justinsb, line 40: Ah I see - our existing "internal" test was actually internal & external, and so we're adding a test for the existing test case.  SGTM

justinsb:
IIUC we do expect this to fail until we have #1533 , but yes will be great to see that failure!

/ok-to-test

justinsb:
(Not giving LGTM/approve because we do expect this to fail, at least AIUI :-) )

bochengchu:
This should pass because it only create 1 control node. Just realized I need to let the management cluster to be in the same private network...

bochengchu:
Skip the tests for now and remove the code for uploading images and creating GKE clusters while we are still figuring them out.

Also uploaded the log of local testing.

PR #1537, opened 2025-09-26
Add support for nested virtualization on GCE instances

jwmay2012:
**What type of PR is this?**
/kind feature

Adds support for enabling nested virtualization on GCE instances via the `enableNestedVirtualization` field in GCPMachineSpec. This maps directly to GCP's AdvancedMachineFeatures.EnableNestedVirtualization API field.

This enables running VMs inside GCE instances for container sandboxes, KVM, QEMU, etc.

This is a simple boolean field that maps directly to the GCP API. The field must be set at instance creation time and cannot be changed after creation (GCP limitation).

**TODOs**:
- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add support for nested virtualization on GCE instances. `GCPMachine.Spec.EnableNestedVirtualization` allows running VMs inside GCE instances.
```


api/v1beta1/gcpmachine_types.go
```diff
@@ -365,6 +365,13 @@ type GCPMachineSpec struct {
 	// attached to the instance.
 	// +optional
 	GuestAccelerators []Accelerator `json:"guestAccelerators,omitempty"`
+
+	// EnableNestedVirtualization specifies whether to enable nested virtualization or not
```
justinsb, line 369: I was wondering if we should nest this under `advancedMachineFeatures` to reflect the GCP API, but to be honest I'm not sure that "advanced" is really a semantic grouping, so I think I prefer the flattened structure.  It's always risky to change the structure because we can't predict the future evolution of the upstream API, but in this case I think it's fine!

docs/book/src/topics/nested-virtualization.md
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  template:
+    spec:
+      instanceType: n2-standard-8
+      enableNestedVirtualization: true
+```
+
+https://cloud.google.com/compute/docs/instances/nested-virtualization/overview
+
+NOTE: Nested virtualization must be enabled at instance creation time and cannot be changed after the instance is created.
```
justinsb: I think we generally treat machines as immutable anyway - i.e. we would launch a new machine, even if the field was mutable.  My understanding could be out of date though.

Not sure there's any harm in the NOTE though, but I don't think it's going to be a problem in practice

jwmay2012: Agree. Removed the note

api/v1beta1/gcpmachine_types.go
jwmay2012: Good thoughts. This was my instinct at the present time, but have no sense of how the API might change. If you think it's fine, I'm happy with it here.

PR #1549, opened 2025-10-23
Add verify-go-directive check from core CAPI

damdo:
**What type of PR is this?**
/kind other

Same reason as: https://github.com/kubernetes-sigs/cluster-api/issues/10258

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ export GO111MODULE=on
 
 # Go version
 GOLANG_VERSION := 1.24.6
+GOLANG_DIRECTIVE_VERSION ?= 1.24.0
 
 # Kubebuilder
 export KUBEBUILDER_ENVTEST_KUBERNETES_VERSION ?= 1.33.0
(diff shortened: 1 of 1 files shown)


damdo:
/assign @justinsb @salasberryfin 

This is used in CAPI upstream and checks the go.mod directive is in concert with the desired one specified in the Makefile.
Sounds like a solution to the common CAPI problem and go bumping strategy we discussed here: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1515

Makefile
```diff
@@ -611,3 +612,9 @@ verify-gen: generate
 		git diff HEAD; \
 		exit 1; \
 	fi
+
+.PHONY: verify-go-directive
+verify-go-directive:
+	# use the core Cluster API script directly to verify the go directive matches the desired one.
+	# ref: https://github.com/kubernetes-sigs/cluster-api/blob/v1.10.7/hack/verify-go-directive.sh
+	curl --retry $(CURL_RETRIES) -fsL https://raw.githubusercontent.com/kubernetes-sigs/cluster-api/refs/tags/v1.11.0/hack/verify-go-directive.sh | bash -s -- -g $(GOLANG_DIRECTIVE_VERSION)
```
chrischdi, line 620: (I'd copy over that script instead of relying on curl, maybe could hit rate limits or flake due to other things, but up to you)

chrischdi (commented):
Up to you

/lgtm
/hold

Makefile
damdo: I'd prefer not to maintain that also over here TBH
If we see issues with curl we might want to revisit :)

PR #1550, opened 2025-10-23
Add test case for 3 control plane nodes and internal load balancer

bochengchu:
**What type of PR is this?**
/kind other

Follow-up for #1536. This should fail now, but should be passing after #1533 is merged.
Local test log: [log.txt](https://github.com/user-attachments/files/23308638/e2e_3cp_fail.txt)
With #1533, local test passed: 
[log.txt](https://github.com/user-attachments/files/23308650/e2e_3cp_pass.txt)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -243,6 +243,36 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	// Skipping the test case for now. The internal load balancer test case requires the management cluster
+	// to have access to network that the cluster being created is in.
+	// An option to reach that is to use a GKE cluster as the management cluster.
+	Context("Creating a control-plane cluster with three control plane nodes and an internal load balancer", func() {
+		It("Should create a cluster with 3 control-plane and 1 worker node with an internal load balancer", func() {
+			Skip("This test requires a bootstrap cluster that has access to the network where the cluster is being created.")
+
+			clusterName := fmt.Sprintf("%s-internal-lb", clusterNamePrefix)
+			By("Creating a cluster with internal load balancer from GKE bootstrap cluster")
+			clusterctl.ApplyClusterTemplateAndWait(ctx, clusterctl.ApplyClusterTemplateAndWaitInput{
+				ClusterProxy: bootstrapClusterProxy,
+				ConfigCluster: clusterctl.ConfigClusterInput{
+					LogFolder:                clusterctlLogFolder,
+					ClusterctlConfigPath:     clusterctlConfigPath,
+					KubeconfigPath:           bootstrapClusterProxy.GetKubeconfigPath(),
+					InfrastructureProvider:   clusterctl.DefaultInfrastructureProvider,
+					Flavor:                   "ci-with-internal-lb",
+					Namespace:                namespace.Name,
+					ClusterName:              clusterName,
+					KubernetesVersion:        e2eConfig.MustGetVariable(KubernetesVersion),
+					ControlPlaneMachineCount: ptr.To[int64](3),
+					WorkerMachineCount:       ptr.To[int64](1),
+				},
+				WaitForClusterIntervals:      e2eConfig.GetIntervals(specName, "wait-cluster"),
+				WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
+				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
+			}, result)
+		})
+	})
+
 	Context("Creating a cluster using a cluster class", func() {
 		It("Should create a cluster class and then a cluster based on it", func() {
 			clusterName := fmt.Sprintf("%s-topology", clusterNamePrefix)


config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/k8s-staging-cluster-api-gcp/cluster-api-gcp-controller:e2e
+      - image: ${CONTROLLER_IMAGE}
```
justinsb: Do we need to to do this?  The tests are running I believe?

It's not a blocker for ok-to-test, but it will likely be a blocker for merging.  If we have to do this I suggest splitting it out into its own commit (or PR?), and a comment would help out future selves understand why

scripts/ci-e2e.sh
```diff
@@ -127,6 +133,14 @@ init_networks() {
     --nat-all-subnet-ip-ranges --auto-allocate-nat-external-ips
 }
 
+# create a GKE cluster to be used as a bootstrap cluster
+create_gke_bootstrap_cluster() {
+  gcloud container clusters create "${TEST_NAME}-gke-bootstrap" --project "$GCP_PROJECT" \
```
justinsb: OK this is nice, because we can we can develop Workload Identity support, and then hopefully add the same thing to kind.  (Workload Identity is Google's name for automatic authentication to GCP without needing a ServiceAccount Key in a secret, historically it used a proxy but now it can use identity federation via IAM workloadPools).

But ... probably could be a separate commit or PR.  This one is less of a blocker though IMO, because it's additive

scripts/ci-e2e.sh
```diff
@@ -275,6 +292,16 @@ EOF
     init_networks
   fi
 
+  # Initialize the GKE bootstrap cluster
+  if [[ -n "${SKIP_INIT_GKE_BOOTSTRAP:-}" ]]; then
+    echo "Skipping GKE bootstrap cluster initialization..."
+  else
+    create_gke_bootstrap_cluster
```
justinsb: I _think_ we don't have permission to create GKE clusters in prow (and we _probably_ don't want to take a hard dependency on GKE in OSS testing).    I suggest this should be a different script, or that the default behavior should not change (e.g. only create a GKE cluster if TEST_MANAGEMENT_CLUSTER=gke or something like that)

test/e2e/data/infrastructure-gcp/cluster-template-ci-with-external-and-internal-lb.yaml
```diff
@@ -0,0 +1,173 @@
+---
+apiVersion: cluster.x-k8s.io/v1beta1
```
justinsb, line 2: Not something to be addressed in this PR (i.e. this problem already exists), but we should consider using kustomize or patching a base config in code, so we can understand more easily what is different in each scenario.

test/e2e/e2e_test.go
```diff
@@ -35,19 +35,20 @@ import (
...
+		result                                      *clusterctl.ApplyClusterTemplateAndWaitResult
+		clusterNamePrefix                           string
+		clusterctlLogFolder, gkeClusterctlLogFolder string
 	)
 
 	BeforeEach(func() {
 		Expect(e2eConfig).ToNot(BeNil(), "Invalid argument. e2eConfig can't be nil when calling %s spec", specName)
 		Expect(clusterctlConfigPath).To(BeAnExistingFile(), "Invalid argument. clusterctlConfigPath must be an existing file when calling %s spec", specName)
 		Expect(bootstrapClusterProxy).ToNot(BeNil(), "Invalid argument. bootstrapClusterProxy can't be nil when calling %s spec", specName)
+		Expect(bootstrapGKEClusterProxy).ToNot(BeNil(), "Invalid argument. bootstrapGKEClusterProxy can't be nil when calling %s spec", specName)
```
justinsb: I'm not sure what this is (yet).  Is this required if we use kind for the bootstrap cluster (instead of GKE)?

test/e2e/e2e_test.go
```diff
@@ -213,6 +230,66 @@ var _ = Describe("Workload cluster creation", func() {
 		})
 	})
 
+	Context("Creating a control-plane cluster with an internal load balancer", func() {
+		It("Should create a cluster with 1 control-plane and 1 worker node with an internal load balancer", func() {
+			// This test requires a GKE bootstrap cluster.
```
justinsb: Does it _have_ to require a GKE cluster?  I don't think we allow GKE in prow (though we'll find out in a minute I guess!)

config/default/manager_image_patch.yaml
bochengchu: This is needed for the GKE cluster, since we have to push the image to GCR. Unlike the existing kind clusters where we can load the image into the cluster, for GKE we have to upload it.

scripts/ci-e2e.sh
bochengchu: Sounds good, if we end up using GKE cluster I can separate that.

test/e2e/e2e_test.go
bochengchu: It doesn't required to be GKE, but using a GKE cluster is the simplest way I can think of. Because of the ILB, we have to let the controllers stay within the same private network. GKE is one way to do that, and we can also have a kind cluster like the existing tests, but it'd be much more complicated to configure that. One possibility I found is to use Cloud VPN and create a tunnel or something, and we have to configure the network (and router?) stuff on where the kind cluster is hosted.

test/e2e/e2e_test.go
```diff
@@ -188,8 +188,38 @@ var _ = Describe("Workload cluster creation", func() {
...
+			}, result)
+		})
+	})
+
+	// Skipping the test case for now. The internal load balancer test case requires the management cluster
+	// to have access to network that the cluster being created is in.
+	// An option to reach that is to use a GKE cluster as the management cluster.
 	Context("Creating a control-plane cluster with an internal load balancer", func() {
 		It("Should create a cluster with 1 control-plane and 1 worker node with an internal load balancer", func() {
+			Skip("This test requires a bootstrap cluster that has access to the network where the cluster is being created.")
```
justinsb, line 221: Nit: we should have a way to test this locally (i.e. a conditional skip)

PR #1575, opened 2025-12-23
feat: implement proper context in scope methods

### LINKED ISSUE #1573: Implement proper contexts in scope methods
# Description

The project uses `context.TODO()` extensively because no proper context is implemented in some scope methods. This is an accepted solution for a simplified use of contexts in testing code and we can keep this unchanged, but it should be fixed in any non-test files as this context configuration is essentially a placeholder that will never time out.

A quick search shows that this should be addressed in the following files:
- `cloud/scope/machine.go`
- `cloud/scope/cluster.go`
- `cloud/scope/managedcluster.go`
- `cloud/scope/managedcontrolplane.go`
- `cloud/scope/managedmachinepool.go`

*I may have missed other occurrences of this syntax*

One possible solution is to pass contexts as a scope method parameter:
```golang
func (s *ClusterScope) PatchObject(ctx context.Context) error {
    ...
}
```

This solution is aligned with other CAPI projects, i.e. [CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/38a062d79b3694df4d4b1bed4eaf603e0d858104/azure/scope/machine.go#L659)

salasberryfin:
/kind feature

As discussed in https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1506#discussion_r2504245800 there's no proper context in place for scope methods and using `context.TODO()` is not a valid solution as it's only expected to be used temporarily, for testing or in non-production projects. Effectively this is simply a workaround for passing a nil value to functions that require a context object but it just won't ever timeout.

This PR tries to implement proper context to scope methods. I think I updated all of them, but feel free to call me out on any missing occurrences of `context.TODO()`.

Fixes #1573 

No changes are applied to testing files where `context.TODO()` is still used and should be considered a valid solution for simpler context management.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/interfaces.go
+++ b/cloud/interfaces.go
@@ -90,7 +90,7 @@ type MachineGetter interface {
 	ControlPlaneGroupName() string
 	GetInstanceID() *string
 	GetProviderID() string
-	GetBootstrapData() (string, error)
+	GetBootstrapData(ctx context.Context) (string, error)
 	GetInstanceStatus() *infrav1.InstanceStatus
 }
 
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -421,11 +421,11 @@ func (s *ClusterScope) TargetTCPProxySpec() *compute.TargetTcpProxy {
 // ANCHOR_END: ClusterControlPlaneSpec
 
 // PatchObject persists the cluster configuration and status.
-func (s *ClusterScope) PatchObject() error {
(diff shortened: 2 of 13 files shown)


salasberryfin (commented):
cc @damdo @cpanato

damdo (commented):
/lgtm

PR #1580, opened 2026-01-01
feat: added reservation affinity

### LINKED ISSUE #1579: Support Reservation Affinity in GCPMachine and GCPMachineTemplate
/kind feature

Currently, GCPMachine and GCPMachineTemplate do not appear to expose a field to configure Reservation Affinity.

Google Compute Engine allows instances to target specific reservations to ensure capacity guarantees or to utilize committed resources (Sole-tenant nodes or zonal reservations). Without this field, CAPG-provisioned nodes rely on the default behavior (often ANY_RESERVATION), which may not be sufficient for users who need to target specific reservations by name or explicitly opt-out of consuming reservations.

MenD32:
**What type of PR is this?**

Added reservation affinity support in `GCPMachineSpec`

Fixes #1579

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Added reservation affinity support in `GCPMachineSpec`, allowing users to specify how GCP machines consume reservations.
```

### DIFF
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -504,6 +504,18 @@ func (m *MachineScope) InstanceSpec(log logr.Logger) *compute.Instance {
 		instance.Scheduling.OnHostMaintenance = "TERMINATE"
 	}
 
+	if m.GCPMachine.Spec.ReservationAffinity != nil {
+		instance.ReservationAffinity = &compute.ReservationAffinity{
+			ConsumeReservationType: string(m.GCPMachine.Spec.ReservationAffinity.GetConsumeReservationType()),
+		}
+		if m.GCPMachine.Spec.ReservationAffinity.Key != nil {
+			instance.ReservationAffinity.Key = m.GCPMachine.Spec.ReservationAffinity.GetKey()
+		}
+		if len(m.GCPMachine.Spec.ReservationAffinity.Values) > 0 {
+			instance.ReservationAffinity.Values = m.GCPMachine.Spec.ReservationAffinity.GetValues()
+		}
+	}
+
 	return instance
 }
 
--- a/api/v1beta1/gcpmachine_types.go
+++ b/api/v1beta1/gcpmachine_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta1
 
 import (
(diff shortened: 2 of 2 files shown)


cloud/scope/machine.go
```diff
@@ -504,6 +504,18 @@ func (m *MachineScope) InstanceSpec(log logr.Logger) *compute.Instance {
 		instance.Scheduling.OnHostMaintenance = "TERMINATE"
 	}
 
+	if m.GCPMachine.Spec.ReservationAffinity != nil {
+		instance.ReservationAffinity = &compute.ReservationAffinity{
+			ConsumeReservationType: string(m.GCPMachine.Spec.ReservationAffinity.GetConsumeReservationType()),
```
salasberryfin, line 509:
```suggestion
			ConsumeReservationType: m.GCPMachine.Spec.ReservationAffinity.GetConsumeReservationType(),
```
`GetConsumeReservationType()` returns a string already, so there's no need to explicitly convert it.

cloud/scope/machine.go
```diff
@@ -504,6 +504,18 @@ func (m *MachineScope) InstanceSpec(log logr.Logger) *compute.Instance {
...
+		}
+		if m.GCPMachine.Spec.ReservationAffinity.Key != nil {
+			instance.ReservationAffinity.Key = m.GCPMachine.Spec.ReservationAffinity.GetKey()
+		}
+		if len(m.GCPMachine.Spec.ReservationAffinity.Values) > 0 {
+			instance.ReservationAffinity.Values = m.GCPMachine.Spec.ReservationAffinity.GetValues()
+		}
```
salasberryfin, line 516:
```suggestion
		instance.ReservationAffinity.Key = m.GCPMachine.Spec.ReservationAffinity.GetKey()
		instance.ReservationAffinity.Values = m.GCPMachine.Spec.ReservationAffinity.GetValues()

```
The `getter` methods used here already deal with nil values and we should be fine calling them without checking. What do you think?

salasberryfin (commented):
Thanks @MenD32!

The logic looks good to me. I suggested a couple of changes on how we deal with the `ReservationAffinity` struct.

Issue #1486, opened 2025-06-09
Support IPv6 configuration


barbacbd:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
Support clusters where the network infrastructure is IPv6 based rather than IPv4.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
The resources that most likely need to change:

- Load Balancers
  - Backend Service
  - Address
  - Forwarding Rules
- Health Checks
- Firewall rules
- Subnets
- Instances

**There may be more these are the recent findings**.

barbacbd:
Note: This may currently only be available as Dual Stack.

damdo:
@barbacbd is this issue still needed?

Issue #478, opened 2021-12-15
ipv6 support


sayantani11:
/kind feature

Recently, I was informed that CAPG, doesn't have ipv6 support. So it will be a great enhancement if we could add this feature. Although I haven't had the time to look into the issue, I will have to dig in more in order to understand about it. Any sort of leads and help will be helpful.

sayantani11:
After this we can proceed with #451

calderonIrene:
Hello! My name is Irene Calderón, I am a third year student of computer engineering eager to continue learning and start to be included in the working world.

I am enrolled in the GCoS and this project has caught my attention because it is not one of the topics that are most covered at the university and I really like it. 

I would be interested in helping and thus enhance my knowledge.

I hope you will consider my application for this project. 
Thank you very much!

@cpanato @dims

PR #1582, opened 2026-01-08
Adding dual-stack support for Cluster API GCP Provider

barbacbd:
api/v1beta/types.go:

Adding support for machines and simple network changes for IPV6 or dual stack work.
- InternalIpv6PrefixLength
- IPv6Address
- StackType

cloud/scope/machine.go:

Adding support for instances and networks to allow dual stack components.

- InstanceNetworkInterfaceSpec
  - InternalIpv6PrefixLength
  - Ipv6AccessConfigs
    - ExternalIPv6
    - ExternalIpv6PrefixLength
    - Type (when present always set to DIRECT_IPV6)
    - Name (always set to External IPv6)
  - IPv6AccessType
  - Ipv6Address
  - StackType

- InstanceSpec
  - PrivateIpv6GoogleAccess

- InstanceNetworkInterfaceAliasIPRangesSpec ** This did not change. The AliasIPs appear to only support IPv4 CIDR or single address format.

cloud/interfaces.go:

Expose a couple of new functions to get StackType and IPvAddress information from the cluster specs. These are only getter functions for the Machine.go to access.

 

 
**What type of PR is this?**

/kind feature
/kind api-change

This PR will serve as a focal point for adding dual stack support to CAPG.

Fixes #

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/1486
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/478

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Basic Support for Dual stack infrastructure is added to CAPG.

The resources include addresses, firewall rules, instances, vpc, and subnets.
```

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -503,6 +503,14 @@ type SubnetSpec struct {
 	// can be set only at resource creation time.
 	CidrBlock string `json:"cidrBlock,omitempty"`
 
+	// Ipv6CidrRange is the range of internal IPv6 addresses that are owned by
+	// this subnetwork. Provide this property when you create the subnetwork. The
+	// range must be a unique /64 or larger from the ULA (Unique Local Address) range.
+	// For example, fd00:1234:5678::/64. This field can be set only at resource
+	// creation time and is only applicable when StackType is set to DualStack.
+	// +optional
+	Ipv6CidrRange string `json:"ipv6CidrRange,omitempty"`
+
 	// Description is an optional description associated with the resource.
 	// +optional
 	Description *string `json:"description,omitempty"`
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -511,6 +511,16 @@ type SubnetSpec struct {
 	// +optional
 	Ipv6CidrRange string `json:"ipv6CidrRange,omitempty"`
 
+	// ExternalIpv6 specifies whether the subnet should have external IPv6 access.
+	// When true, the subnet will use EXTERNAL IPv6 access type and instances can
+	// have public IPv6 addresses. When false (default), the subnet will use INTERNAL
+	// IPv6 access type with ULA (Unique Local Address) ranges.
+	// This is only applicable when StackType is set to DualStack.
+	// Note: Subnets with EXTERNAL IPv6 cannot be used with internal load balancers.
+	// +kubebuilder:default=false
+	// +optional
+	ExternalIpv6 *bool `json:"externalIpv6,omitempty"`
+
 	// Description is an optional description associated with the resource.
 	// +optional
 	Description *string `json:"description,omitempty"`
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -309,6 +309,32 @@ const (
 	RulesManagementUnmanaged RulesManagementPolicy = "Unmanaged"
 )
 
+// StackType is a string enum type indicating the types of network addresses that are valid.
+// +kubebuilder:validation:Enum=IPv4Only;DualStack
+type StackType string
+
+const (
+	// IPv4OnlyStackType indicates a stack type where only IPv4 addresses are valid.
+	IPv4OnlyStackType StackType = "IPv4Only"
+
+	// DualStackType indicates a stack type where both IPv4 and IPv6 addresses are valid.
+	DualStackType StackType = "DualStack"
+)
+
+const (
+	// DualStackAdditionalResourceSuffix is an identifier appended to the resource name to indicate
+	// that it is not for the ipv4 [default] stack type.
+	DualStackAdditionalResourceSuffix string = "ipv6"
+
+	// DualStackNetworkAccess is an identifier to describe the network access settings
+	// for dual stack infrastructure. IPv4 resources are default to an internal settings, but
+	// IPv6 resources require the identifier for ULAs.
+	DualStackNetworkAccess = "INTERNAL"
+
+	// GCPDualStack is the identifier used by GCP to indicate a dual stack infrastructure.
+	GCPDualStack = "IPV4_IPV6"
+)
+
 // NetworkSpec encapsulates all things related to a GCP network.
 type NetworkSpec struct {
 	// Name is the name of the network to be used.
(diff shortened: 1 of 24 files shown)


cloud/scope/machine.go
```diff
@@ -326,12 +326,31 @@ func (m *MachineScope) InstanceNetworkInterfaceSpec() *compute.NetworkInterface
...
-			},
+		switch m.ClusterGetter.StackType() {
+		case infrav1.IPv4OnlyStackType:
+			networkInterface.AccessConfigs = []*compute.AccessConfig{
+				{
+					Type: "ONE_TO_ONE_NAT",
+					Name: "External NAT",
+				},
+			}
+		case infrav1.DualStackType:
```
salasberryfin: Shouldn't we have IPv4 access config here too? I guess we could just configure IPv4 no matter the stack type, then add IPv6 if dual stack is enabled. Does this make sense?

api/v1beta1/types.go
```diff
@@ -191,6 +203,28 @@ type NetworkSpec struct {
...
+	Ipv6Address string `json:"ipv6Address,omitempty"`
+
+	// StackType: The stack type for the subnet. If set to IPV4_ONLY, new VMs in
+	// the subnet are assigned IPv4 addresses only. If set to IPV4_IPV6, new VMs in
+	// the subnet can be assigned both IPv4 and IPv6 addresses. If not specified,
+	// IPV4_ONLY is used. This field can be both set at resource creation time and
+	// updated using patch.
+	// +kubebuilder:default=IPV4_ONLY
+	// +optional
+	StackType StackType `json:"stackType,omitempty"`
```
salasberryfin, line 445: I think `StackType` is a subnet option but it is defined here as a network configuration. Should we make it configurable at the subnet level?

salasberryfin (commented):
Thanks @barbacbd, I left a couple of comments.

api/v1beta1/types.go
barbacbd: @salasberryfin it is a stacktype in the network and the subnets. The reason that I set it up this way was for a case where I don't think we could get that information from the subnets. I believe that they are not accessible cluster.go. I wasn't sure what other design changes we could/should make to allow that information to be accessible.

cloud/scope/machine.go
barbacbd: Yup that makes sense to me. That was just a blunder on my part 😄

api/v1beta1/types.go
salasberryfin: We can initially add it at the network level and think about having a follow-up to investigate if it's worth working on supporting a per subnet configuration. Does this make sense?

salasberryfin (commented):
A couple of "nice to haves" that could be added here:
- E2E test case using a dual stack network configuration.
- A brief entry in the CAPG book that documents the new feature (this can be a follow-up PR).

api/v1beta1/types.go
barbacbd: @salasberryfin thats fine, but it is an API change so it may be best to work that out early. cc @damdo

api/v1beta1/types.go
```diff
@@ -137,6 +137,24 @@ const (
 	RulesManagementUnmanaged RulesManagementPolicy = "Unmanaged"
 )
 
+// StackType is a string enum type indicating the types of network addresses that are valid.
+// +kubebuilder:validation:Enum=IPV4_ONLY;IPV4_IPV6
```
JoelSpeed: Enum values should be PascalCase ideally, that would mean no _ here, what about?

```
IPv4Only, DualStack
```

api/v1beta1/types.go
```diff
@@ -191,6 +209,31 @@ type NetworkSpec struct {
...
+	// +kubebuilder:validation:Maximum=128
+	// +optional
+	InternalIpv6PrefixLength int `json:"internalIpv6PrefixLength,omitempty"`
+
+	// Ipv6Address: An IPv6 internal network address for this network interface.
+	// To use a static internal IP address, it must be unused and in the same
+	// region as the instance's zone. If not specified, Google Cloud will
+	// automatically assign an internal IPv6 address from the instance's subnetwork.
+	// +optional
+	Ipv6Address string `json:"ipv6Address,omitempty"`
```
JoelSpeed: If these are related, is it useful to have it as a combined CIDR string?

api/v1beta1/types.go
JoelSpeed: Ok so if we were to add this here, the later addition of a field at the subnet level would be the "I have an opinion for this specific subnet" style. When not set, I would image it falls back to the network level. I think that's probably ok

barbacbd: Discussing with my team they suggested that we only apply this field at the network level so its an all or nothing case. All subnets would be IPV4 Only or Dual Stack. This would be the implementation for now. We can expand this later if there are needs for it.

JoelSpeed: That's fine by me

salasberryfin: Sounds good to me too.

api/v1beta1/types.go
barbacbd: These are related but I believe that they are independent. Specifying one does not mean you must specify the other.

cloud/scope/machine.go
```diff
@@ -331,13 +331,37 @@ func InstanceNetworkInterfaceSpec(cluster cloud.ClusterGetter, publicIP *bool, s
...
+
 	if publicIP != nil && *publicIP {
 		networkInterface.AccessConfigs = []*compute.AccessConfig{
 			{
 				Type: "ONE_TO_ONE_NAT",
 				Name: "External NAT",
 			},
 		}
+
+		// For now we cannot assign the IPv6AccessConfigs. The bootstrap node is the only one on our side
```
barbacbd: @justinsb I would love to get your input on this area. The problem we are currently seeing is 
1. We need to set the ipv6 access to INTERNAL to mirror what is default for IPv4. This will also ensure we can have an internal Load Balancer.
2. We set the `network.EnableUlaInternalIpv6 = true` to reflect this also and allow internal addresses

The problem is on our bootstrap node. This usually is given a public IP address and is handled in the lines above. But when this instance is created it is no longer given a public/global IPv6 address because the network was set to INTERNAL. 

It looks like googles suggestion would be to add another subnet and set that to external then open the bootstrap node on that subnet. We would then create another AccessConfig to match. This seems like alot of work potentially adding more issues. 

Any input would be helpful. Thank you!

/cc @damdo @JoelSpeed @salasberryfin

damdo:
@barbacbd are you able to rebase this? Thanks!

damdo:
/assign @barbacbd 

For rebasing it

PR #1616, opened 2026-03-06
chore: harden git workflows

PrashantR30:
What type of PR is this?
/kind cleanup

What this PR does / why we need it:

Hardens the security of GitHub Actions workflows by:

 - Adding `persist-credentials: false` to all checkout steps that do not require git credentials after checkout
 - Adding `permissions: {}` at workflow level where missing
 - Adding `types: [opened, edited, synchronize, reopened]` to `pull_request` triggers to avoid running on every trivial PR event

Which issue(s) this PR fixes:

Fixes #

Special notes for your reviewer:

No image versions were changed in this PR. All changes are limited to GitHub Actions workflow security hardening. The `persist-credentials: false` changes are safe as verified step-by-step as no workflow requires git credentials after checkout except the dependabot workflow which intentionally retains `persist-credentials: true`.

TODOs:

 - [ ] squashed commits
 - [ ] includes documentation
 - [ ] adds unit tests

```release-note
NONE
```

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -3,6 +3,7 @@ name: Codespell
 
 on:
   pull_request:
+    types: [opened, synchronize, reopened]
 
 permissions: {}
 
@@ -11,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
       - uses: codespell-project/actions-codespell@8f01853be192eb0f849a5c7d721450e7a467c579 # v2.2
         with:
--- a/.github/workflows/pr-gh-workflow-approve.yaml
+++ b/.github/workflows/pr-gh-workflow-approve.yaml
@@ -7,6 +7,8 @@ on:
       - labeled
       - reopened
       - synchronize
+      
+permissions: {}
 
 jobs:
   approve:
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -8,6 +8,7 @@ on:
     branches:
       - main
   pull_request:
+    types: [opened, synchronize, reopened]
     paths: "docs/**"
 
 permissions: {}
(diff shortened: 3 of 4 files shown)


.github/workflows/codespell.yml
```diff
@@ -3,6 +3,7 @@ name: Codespell
 
 on:
   pull_request:
+    types: [opened, edited, synchronize, reopened]
```
salasberryfin:
```suggestion
    types: [opened, synchronize, reopened]
```
As far as I understand `edited` triggers when the PR body or title is changed. I think this is an unnecessary type for this (and other) workflow(s).

.github/workflows/pr-gh-workflow-approve.yaml
```diff
@@ -7,6 +7,8 @@ on:
       - labeled
       - reopened
       - synchronize
+      
+permissions: {}
```
salasberryfin, line 11: This specific workflow approves running all other checks and requires `write` permissions. In this case it is applied at the job level and that should be enough to run successfully but, can this interfere in some way?

.github/workflows/codespell.yml
PrashantR30: Thanks for this observation. You are right, we do not need to trigger the workflows (and others) when the title or body is changed. The PR was updated accordingly.

.github/workflows/pr-gh-workflow-approve.yaml
PrashantR30: When the explicitly set permissions at the job level, they are independent of the workflow level. The job-level block is a complete override, not additive.

That is:

```
permissions: {}  # workflow level - no permissions

jobs:
  approve:
    permissions:
      actions: write  # job level - completely overrides workflow level for this job
```

This means:

The job will have `actions: write` regardless of `permissions: {}` at the top. Other jobs in the same workflow would have no permissions. This is actually the recommended pattern by GitHub for least-privilege

So adding `permissions: {}` at the workflow level is safe and correct here and it does not cap or restrict what the job-level permissions can grant. 

Suggest that we let the test go through and we can confirm.

salasberryfin: Thanks @PrashantR30. Looks like it worked.

PR #1617, opened 2026-03-06
fix: Address the use of managing default firewall rules

barbacbd:
cloud/scope/firewall.go:
Fix the bug by adding the management policy to the firewall rules creation.

cloud/scope/cluster.go:
cloud/scope/managedcluster.go:
Provide the rules management policy to the firewall rule function.

 

 
**What type of PR is this?**

/kind bug

The default management policy was used in some regards but was not used when including firewall rules.

### Expected Behaviour table (edit: @damdo)

Default Rules Management | User Provided Rules | Shared VPC | Expected Result
-- | -- | -- | --
Managed | Yes | No | Create Default Rules, Create Custom Rules
Managed | No | No | Create Default
Managed | Yes | Yes | Skip Rules Creation
Managed | No | Yes | Skip Rules Creation
Unmanaged | Yes | No | Create Custom
Unmanaged | No | No | Skip Rules Creation
Unmanaged | Yes | Yes | Skip Rules Creation
Unmanaged | No | Yes | Skip Rules Creation

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -106,12 +106,11 @@ func (s *ClusterScope) NetworkProject() string {
 	return ptr.Deref(s.GCPCluster.Spec.Network.HostProject, s.Project())
 }
 
-// SkipFirewallRuleCreation returns whether the spec indicates that firewall rules
-// should be created or not. If the RulesManagement for the default firewall rules is
-// set to unmanaged or when the cluster will include a shared VPC, the default firewall
-// rule creation will be skipped.
-func (s *ClusterScope) SkipFirewallRuleCreation() bool {
-	return (s.GCPCluster.Spec.Network.Firewall.DefaultRulesManagement == infrav1.RulesManagementUnmanaged) || s.IsSharedVpc()
+// SkipFirewallRulesManagement returns whether the spec indicates that firewall rules
+// should be created or not. Firewall rules will not be created when the cluster
+// contains a shared VPC network.
+func (s *ClusterScope) SkipFirewallRulesManagement() bool {
+	return s.IsSharedVpc()
 }
 
 // IsSharedVpc returns true If sharedVPC used else , returns false.
@@ -293,7 +292,12 @@ func (s *ClusterScope) SubnetSpecs() []*compute.Subnetwork {
 
 // FirewallRulesSpec returns google compute firewall spec.
 func (s *ClusterScope) FirewallRulesSpec() []*compute.Firewall {
-	return createFirewallRules(s.Name(), s.NetworkLink(), s.GCPCluster.Spec.Network.Firewall.FirewallRules)
+	return createFirewallRules(
+		s.Name(),
+		s.NetworkLink(),
+		s.GCPCluster.Spec.Network.Firewall.DefaultRulesManagement,
+		s.GCPCluster.Spec.Network.Firewall.FirewallRules,
+	)
 }
 
 // ANCHOR_END: ClusterFirewallSpec
--- a/cloud/scope/firewall.go
+++ b/cloud/scope/firewall.go
@@ -11,46 +11,52 @@ import (
 )
 
 // createFirewallRules
-func createFirewallRules(clusterName, networkLink string, userSpecifiedRules []infrav1.FirewallRule) []*compute.Firewall {
-	firewallRules := []*compute.Firewall{
-		{
-			Name:    fmt.Sprintf("allow-%s-healthchecks", clusterName),
-			Network: networkLink,
-			Allowed: []*compute.FirewallAllowed{
-				{
-					IPProtocol: "TCP",
-					Ports: []string{
-						strconv.FormatInt(6443, 10),
+func createFirewallRules(clusterName, networkLink string, policy infrav1.RulesManagementPolicy, userSpecifiedRules []infrav1.FirewallRule) []*compute.Firewall {
+	firewallRules := []*compute.Firewall{}
+
+	// Only when the user explicitly states that it is unmanaged, the rules should be skipped.
+	// When the policy is Managed or missing/empty the rules should be included.
+	if policy != infrav1.RulesManagementUnmanaged {
+		firewallRules = append(firewallRules, []*compute.Firewall{
+			{
+				Name:    fmt.Sprintf("allow-%s-healthchecks", clusterName),
+				Network: networkLink,
+				Allowed: []*compute.FirewallAllowed{
+					{
+						IPProtocol: "TCP",
+						Ports: []string{
+							strconv.FormatInt(6443, 10),
+						},
 					},
 				},
-			},
-			Direction: "INGRESS",
-			SourceRanges: []string{
-				"35.191.0.0/16",
-				"130.211.0.0/22",
-			},
-			TargetTags: []string{
-				clusterName + "-control-plane",
-			},
-		},
-		{
-			Name:    fmt.Sprintf("allow-%s-cluster", clusterName),
-			Network: networkLink,
-			Allowed: []*compute.FirewallAllowed{
-				{
-					IPProtocol: "all",
+				Direction: "INGRESS",
+				SourceRanges: []string{
+					"35.191.0.0/16",
+					"130.211.0.0/22",
+				},
+				TargetTags: []string{
+					clusterName + "-control-plane",
 				},
 			},
-			Direction: "INGRESS",
-			SourceTags: []string{
-				clusterName + "-control-plane",
-				clusterName + "-node",
+			{
+				Name:    fmt.Sprintf("allow-%s-cluster", clusterName),
+				Network: networkLink,
+				Allowed: []*compute.FirewallAllowed{
+					{
+						IPProtocol: "all",
+					},
+				},
+				Direction: "INGRESS",
+				SourceTags: []string{
+					clusterName + "-control-plane",
+					clusterName + "-node",
+				},
+				TargetTags: []string{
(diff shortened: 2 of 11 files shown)


cloud/scope/cluster.go
```diff
@@ -111,7 +111,7 @@ func (s *ClusterScope) NetworkProject() string {
 // set to unmanaged or when the cluster will include a shared VPC, the default firewall
 // rule creation will be skipped.
 func (s *ClusterScope) SkipFirewallRuleCreation() bool {
-	return (s.GCPCluster.Spec.Network.Firewall.DefaultRulesManagement == infrav1.RulesManagementUnmanaged) || s.IsSharedVpc()
+	return s.IsSharedVpc()
 }
```
tthvo: I think the rule management mode check is still needed here right?

tthvo, line 115: Since it's now the same as `IsSharedVpc`, we can remove this func and use `IsSharedVpc` instead?

The reconciler delete handler is already doing it:


https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/58abcdcc039b8c7922057e95df5299ac779d040a/cloud/services/compute/firewalls/reconcile.go#L57-L60

cloud/scope/managedcluster.go
```diff
@@ -134,7 +134,7 @@ func (s *ManagedClusterScope) NetworkProject() string {
 // set to unmanaged or when the cluster will include a shared VPC, the default firewall
 // rule creation will be skipped.
 func (s *ManagedClusterScope) SkipFirewallRuleCreation() bool {
-	return (s.GCPManagedCluster.Spec.Network.Firewall.DefaultRulesManagement == infrav1.RulesManagementUnmanaged) || s.IsSharedVpc()
```
tthvo, line 137: Same idea as above to remove this func and use `IsSharedVpc` ☝️

cloud/scope/firewall.go
```diff
@@ -11,46 +11,52 @@ import (
...
-				{
-					IPProtocol: "TCP",
-					Ports: []string{
-						strconv.FormatInt(6443, 10),
+func createFirewallRules(clusterName, networkLink string, policy infrav1.RulesManagementPolicy, userSpecifiedRules []infrav1.FirewallRule) []*compute.Firewall {
+	firewallRules := []*compute.Firewall{}
+
+	// Only when the user explicitly states that it is unmanaged, the rules should be skipped.
+	// When the policy is Managed or missing/empty the rules should be included.
+	if policy != infrav1.RulesManagementUnmanaged {
```
tthvo, line 19: 🤔  I suppose the `.network.firewall.defaultRulesManagement` applies to the "base" or "default" rules here that CAPG creates (i.e. It does **not** mean skipping rule management all together...).

Maybe we should update the field descriptor to reflect this more clearly:

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/58abcdcc039b8c7922057e95df5299ac779d040a/api/v1beta1/types.go#L270-L272

cloud/scope/managedcluster.go
barbacbd: It may be the same but someone really liked the function name and it can be expanded upon in the future if needed. I would vote to keep this even if it is just the same function for now.

tthvo: Ahh okay, then I guess we should use this function for delete func of the reconciler https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1617#discussion_r2927083353?

cloud/scope/firewall.go
tthvo: Discussed with Brent offline that it's just a nit. The wording seems fine (just me being confused a bit :sweat_smile:).

cloud/scope/cluster.go
```diff
@@ -111,7 +111,7 @@ func (s *ClusterScope) NetworkProject() string {
 // set to unmanaged or when the cluster will include a shared VPC, the default firewall
 // rule creation will be skipped.
 func (s *ClusterScope) SkipFirewallRuleCreation() bool {
-	return (s.GCPCluster.Spec.Network.Firewall.DefaultRulesManagement == infrav1.RulesManagementUnmanaged) || s.IsSharedVpc()
```
damdo, line 114: What's the background of this change?

cloud/scope/cluster.go
```diff
@@ -111,7 +111,7 @@ func (s *ClusterScope) NetworkProject() string {
 // set to unmanaged or when the cluster will include a shared VPC, the default firewall
 // rule creation will be skipped.
 func (s *ClusterScope) SkipFirewallRuleCreation() bool {
```
damdo: Should we change this to:
```suggestion
func (s *ClusterScope) SkipFirewallRulesManagement() bool {
```
then?

cloud/services/compute/firewalls/reconcile.go
```diff
@@ -54,8 +54,8 @@ func (s *Service) Reconcile(ctx context.Context) error {
 // Delete delete cluster firewall compoenents.
 func (s *Service) Delete(ctx context.Context) error {
 	log := log.FromContext(ctx)
-	if s.scope.IsSharedVpc() {
-		log.V(2).Info("Shared VPC enabled. Ignore Deleting firewall resources")
+	if s.scope.SkipFirewallRuleCreation() {
+		log.V(2).Info("Firewall Rule creation skipped. Ignore Deleting firewall resources")
 		return nil
```
damdo: I may be missing some context, but this reads a bit strange to me:
`"Firewall Rule creation skipped. Ignore Deleting firewall resources"`

Does this mean that given the firewall rules were not created by the controller, then we are not deleting them because we weren't the ones managing them?

If so, how about?
```
Skipping firewall rules deletion. Firewall RulesManagement is set to Unmanaged.
```

barbacbd: fixed, thanks!

cloud/scope/cluster.go
barbacbd: If the default firewall rules management is set to unmanaged that means the firewall rules normally created by CAPG that are defaults found [here](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/cloud/scope/firewall.go#L15-L53), should not be created. The user can still provide firewall rules to be created.

cloud/scope/cluster.go
damdo: Ok let's change this then, thanks!

api/v1beta1/#types.go#
```diff
@@ -0,0 +1,619 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
```
tthvo: We don't want to include this file `api/v1beta1/#types.go#`, right?

damdo: Indeed!

cloud/services/compute/firewalls/reconcile.go
```diff
@@ -54,8 +54,8 @@ func (s *Service) Reconcile(ctx context.Context) error {
 // Delete delete cluster firewall compoenents.
 func (s *Service) Delete(ctx context.Context) error {
 	log := log.FromContext(ctx)
-	if s.scope.IsSharedVpc() {
-		log.V(2).Info("Shared VPC enabled. Ignore Deleting firewall resources")
+	if s.scope.SkipFirewallRulesManagement() {
+		log.V(2).Info("Skipping firewall rules deletion. Firewall RulesManagement is set to Unmanaged.")
```
tthvo:
```suggestion
		log.V(2).Info("Ignore Deleting firewall resources when using shared VPC")
```

I think we only skip the entire delete logic if in shared VPC, right? Otherwise, user-provided firewall rules should still be deleted regardless of `DefaultRulesManagement` mode (which only applies to "default" rules).

cloud/services/compute/firewalls/reconcile.go
```diff
@@ -28,7 +28,7 @@ import (
 // Reconcile reconcile cluster firewall compoenents.
 func (s *Service) Reconcile(ctx context.Context) error {
 	log := log.FromContext(ctx)
-	if s.scope.SkipFirewallRuleCreation() {
+	if s.scope.SkipFirewallRulesManagement() {
 		log.V(2).Info("Ignore Reconciling firewall resources")
```
tthvo, line 32:
```suggestion
		log.V(2).Info("Ignore Reconciling firewall resources when using shared VPC")
```

nit: same suggestion as above.

cloud/services/compute/firewalls/reconcile.go
tthvo: I guess simply "Ignore Deleting firewall resources" would work too. Then, we can ignore https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1617#discussion_r2948643026

tthvo: @damdo Does this sound OK to you :eyes:?

cloud/services/compute/firewalls/reconcile.go
damdo: Let's keep this more generic in case this is skipped for more reasons in the future

PR #1622, opened 2026-03-13
feat(lb): implement Regional External load balancer support for S3NS compatibility

DvdChe:
**What type of PR is this?**

This PR introduces support for Regional External and Regional Internal External load balancers for the Kubernetes API Server.

Previously, CAPG primarily relied on Global External Proxy Load Balancers. However, certain GCP-based environments like S3NS (the French sovereign cloud developed in collaboration with Thales) do not implement global load balancing. To enable Cluster API to bootstrap and manage Kubernetes clusters on S3NS, regional alternatives are required for the control plane endpoint.

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Changes
   - API Updates: Added RegionalExternal and RegionalInternalExternal to LoadBalancerType in api/v1beta1.
   - Reconciliation Logic:
       - Implemented createRegionalExternalLoadBalancer in the load balancer service.
       - Added support for EXTERNAL_MANAGED load balancing scheme (required for Regional External Proxy LBs).
       - Added logic to handle regional backends, health checks, and target TCP proxies.
   - Documentation: Created a new Load Balancers (docs/book/src/topics/load-balancers.md) topic in the CAPG book to document all supported types and the specific requirement for a proxy-only subnet when using regional external load balancers.
   - Testing: Added comprehensive unit tests in cloud/services/compute/loadbalancers/reconcile_test.go to verify the creation and deletion flow for the new load balancer types.
```


DvdChe:
Hi, @damdo  @salasberryfin,

Friendly ping — happy to address any feedback or split this into smaller PRs if it helps reviewing.

salasberryfin (commented):
Thanks, @DvdChe.

Can you please re-arrange your commits to maintain a clean history?

DvdChe:
> Thanks, @DvdChe.
> 
> Can you please re-arrange your commits to maintain a clean history?

Hello @salasberryfin, I did some cleanup and rearranged the commits to have 7 commits instead of 26.

api/v1beta1/types.go
```diff
@@ -395,9 +405,16 @@ type LoadBalancerSpec struct {
...
 	// If not set, a Global External Proxy Load Balancer will be created by default.
+	// +kubebuilder:validation:Enum=External;RegionalExternal;Internal;InternalExternal;RegionalInternalExternal
 	// +optional
 	LoadBalancerType *LoadBalancerType `json:"loadBalancerType,omitempty"`
 
+	// ExternalLoadBalancer is the configuration for an External Proxy Load Balancer.
+	// Applies to all load balancer types that include an external component:
+	// External, InternalExternal, RegionalExternal, and RegionalInternalExternal.
+	// +optional
+	ExternalLoadBalancer *LoadBalancer `json:"externalLoadBalancer,omitempty"`
```
barbacbd: The field is orthogonal to LoadBalancerType — it configures name/IP for whichever external LB gets created. Consider renaming to ExternalLoadBalancerConfig to clarify it's configuration metadata, not a resource declaration

barbacbd: Users may not understand that setting this with LoadBalancerType: External (global) vs LoadBalancerType: RegionalExternal produces different LB scopes.

barbacbd: Add API validation that returns a warning/error if ExternalLoadBalancer is set but LoadBalancerType is Internal (no external component exists)

api/v1beta1/types.go
barbacbd, line 370: The LoadBalancerType enum validation is added as a kubebuilder marker:
  ```
// +kubebuilder:validation:Enum=External;RegionalExternal;Internal;InternalExternal;RegionalInternalExternal
  LoadBalancerType *LoadBalancerType `json:"loadBalancerType,omitempty"`
```
This generates CRD schema validation but doesn't prevent Go code from constructing invalid values. If this enum is used in SDK/operator code, runtime validation would be helpful.

Consider adding a Validate() method on LoadBalancerSpec or using CEL validation rules (if CAPG supports them).

File: cloud/services/compute/loadbalancers/backward_compat_test.go [hunk 2]
```diff
@@ -0,0 +1,261 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
barbacbd:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

cloud/services/compute/loadbalancers/helper_test.go
```diff
@@ -0,0 +1,399 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
barbacbd:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -43,10 +43,69 @@ const (
...
+// maxConnections is applied as-is to each Backend; pass 0 to leave the field
+// unset (required for INTERNAL passthrough backend services — GCP rejects
+// maxConnections on those).
+func createBackends(instancegroups []*compute.InstanceGroup, mode loadBalancingMode, maxConnections int64) []*compute.Backend {
+	backends := make([]*compute.Backend, 0, len(instancegroups))
+	for _, group := range instancegroups {
+		backends = append(backends, &compute.Backend{
+			BalancingMode:  string(mode),
+			Group:          group.SelfLink,
+			MaxConnections: maxConnections,
```
barbacbd, line 132: We need to look into the maxConnections. I think that this field is not applicable to all types of backends. Its possible we need slightly different logic for regional vs global LBs.

cloud/services/compute/loadbalancers/reconcile.go
```diff
@@ -87,21 +148,24 @@ func (s *Service) Delete(ctx context.Context) error {
 	var allErrs []error
 	lbSpec := s.scope.LoadBalancer()
 	lbType := ptr.Deref(lbSpec.LoadBalancerType, infrav1.External)
-	if lbType == infrav1.External || lbType == infrav1.InternalExternal {
+	if shouldCreateExternalLoadBalancer(lbType) {
 		if err := s.deleteExternalLoadBalancer(ctx); err != nil {
```
barbacbd, line 181: The deletion logic aggregates errors but always returns the full error from kerrors.NewAggregate(). If deletion of one LB component fails (e.g., forwarding rule is stuck due to a GCP API issue), the entire delete operation fails, potentially blocking cluster deletion.

GCP resources should be deleted in dependency order (forwarding rule → proxy → backend service → health check), and transient failures should be logged with retries, not block cleanup of independent resources.

barbacbd (commented):
Regional External Proxy Load Balancers require a proxy-only subnet (documented in the proposal). However, there's no validation in the reconciliation code to verify this prerequisite exists before attempting creation.

```
// forwarding rule requires a proxy-only subnet.
  // Specifying the network will make the forwarding rule use the proxy-only subnet.
  network := s.scope.Network()
  if network.SelfLink == nil {
      return nil, errors.New("cannot create regional external forwarding rule: network selfLink not yet populated")
  }
```

  The check only validates the network exists, not that a proxy-only subnet is configured in that network. If the subnet is missing, the GCP API call
  will fail with an opaque error.


  - Add explicit validation that checks for the proxy-only subnet before creating the forwarding rule
  - Return a clear, actionable error message if missing, e.g., "Regional external load balancer requires a proxy-only subnet with purpose
  REGIONAL_MANAGED_PROXY in network <name>. See https://cloud.google.com/load-balancing/docs/proxy-only-subnets"
  - Consider adding a presubmit check or admission webhook

DvdChe:
Hi @barbacbd, I implemented the improvement you suggested last week. I forgot to notify you.

api/v1beta1/types.go
```diff
@@ -395,14 +406,57 @@ type LoadBalancerSpec struct {
...
+	Internal:                 {},
+	InternalExternal:         {},
+	RegionalInternalExternal: {},
+}
+
+// Validate returns admission warnings and a field.ErrorList for a LoadBalancerSpec.
+// The kubebuilder Enum marker on LoadBalancerType only enforces the CRD schema;
+// this runs the same check plus a semantic cross-field check that catches
+// configuration that is silently ignored by the reconciler.
+func (s *LoadBalancerSpec) Validate(fldPath *field.Path) (warnings []string, errs field.ErrorList) {
```
barbacbd, line 441: LoadBalancerSpec.Validate() is called in ValidateCreate but the diff shows ValidateUpdate was not modified to call it. A user could update an existing GCPCluster to set loadBalancerType
  to something random and pass webhook validation. Since the kubebuilder enum marker catches CRD-level creates, this mainly matters for programmatic updates via Go SDK/operators (which the PR's own
  commit message 3dafc6f6 explicitly calls out as the reason for runtime validation).

api/v1beta1/types.go
```diff
@@ -395,14 +406,57 @@ type LoadBalancerSpec struct {
...
+	// ExternalLoadBalancerConfig is the configuration (name, IP) applied to the
+	// external Load Balancer created by the controller. Its scope (global vs
+	// regional) is determined solely by LoadBalancerType; this field only carries
+	// naming and address settings and does not itself select a load balancer kind.
+	// Applies to all load balancer types that include an external component:
+	// External (global), InternalExternal (global external + regional internal),
+	// RegionalExternal (regional), and RegionalInternalExternal (regional external
+	// + regional internal). Ignored when LoadBalancerType is Internal.
+	// +optional
+	ExternalLoadBalancerConfig *LoadBalancer `json:"externalLoadBalancerConfig,omitempty"`
```
barbacbd: IPAddress is never consumed/used.

The internal LB path correctly reads InternalLoadBalancer.IPAddress at reconcile.go:526 to set a static IP. But neither createOrGetAddress() (global external) nor createOrGetRegionalAddress() reads ExternalLoadBalancerConfig.IPAddress. The field is declared in the API and documented but silently ignored, users who set it would expect their static IP to be used.

barbacbd: ExternalLoadBalancerConfig inherits inappropriate fields from LoadBalancer struct

  Reusing the LoadBalancer struct means the CRD schema for externalLoadBalancerConfig exposes internalAccess (with description "defines the access for the Internal Passthrough Load
  Balancer") and subnet — both are meaningless for external LBs. The CRD even shows internalAccess with a default of Regional under externalLoadBalancerConfig. This is confusing for users
  reading the CRD spec. Consider either:
  - A separate struct for external LB config (just Name + IPAddress)
  - CEL validation rules to reject those fields when set under externalLoadBalancerConfig

barbacbd: CRD description for externalLoadBalancerConfig is truncated

  In all four CRD files, the description ends mid-sentence:
  RegionalExternal (regional), and RegionalInternalExternal (regional external
  Missing the closing: + regional internal). Ignored when LoadBalancerType is Internal.

  This appears to be a controller-gen line-length truncation issue in the Go source godoc.

docs/proposals/cors-4448/CORS-4448-quick-reference.md
```diff
@@ -0,0 +1,220 @@
+# CORS-4448 Quick Reference
```
barbacbd: We should probably rename these. It was pointed out to me (even though these are mine) that they reference JIRA tickets that are or may not be public.

cloud/services/compute/loadbalancers/backward_compat_test.go
```diff
@@ -0,0 +1,261 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Verify default type resolution
+			gotType := ptr.Deref(tt.lbSpec.LoadBalancerType, infrav1.External)
+			if gotType != tt.wantType {
+				t.Errorf("Default type = %v, want %v", gotType, tt.wantType)
+			}
+
+			// Verify external LB creation logic
+			gotExternal := shouldCreateExternalLoadBalancer(gotType)
```
barbacbd: nit: Can we rename `shouldCreateExternalLoadBalancer` to `shouldCreateGlobalExternalLoadBalancer` or something similar. It explicitly excludes regional types.

barbacbd (commented):
The commit looks good from the fixes before, but looks like a few quick things to fix.

api/v1beta1/types.go
DvdChe: Good catch, thanks. Hooked LoadBalancerSpec.Validate() into ValidateUpdate in 9355d4de so an invalid LoadBalancerType (or the Internal + ExternalLoadBalancerConfig warning) is caught on updates too, not only on create. Warnings are now propagated from ValidateUpdate and I added two matching test cases (invalid LoadBalancerType is rejected on update, ExternalLoadBalancerConfig with Internal LB emits warning on update).

api/v1beta1/types.go
DvdChe: > IPAddress is never consumed/used.
> 
> The internal LB path correctly reads InternalLoadBalancer.IPAddress at reconcile.go:526 to set a static IP. But neither createOrGetAddress() (global external) nor createOrGetRegionalAddress() reads ExternalLoadBalancerConfig.IPAddress. The field is declared in the API and documented but silently ignored, users who set it would expect their static IP to be used.

Fixed in b5126120. Both createOrGetAddress (global) and createOrGetRegionalAddress now honor ExternalLoadBalancerConfig.IPAddress when set, mirroring what createOrGetInternalAddress already did for the internal LB. Added test cases in reconcile_test.go (static IP from ExternalLoadBalancerConfig is honored) for both helpers.

> ExternalLoadBalancerConfig inherits inappropriate fields from LoadBalancer struct
> 
> Reusing the LoadBalancer struct means the CRD schema for externalLoadBalancerConfig exposes internalAccess (with description "defines the access for the Internal Passthrough Load Balancer") and subnet — both are meaningless for external LBs. The CRD even shows internalAccess with a default of Regional under externalLoadBalancerConfig. This is confusing for users reading the CRD spec. Consider either:
> 
> * A separate struct for external LB config (just Name + IPAddress)
> * CEL validation rules to reject those fields when set under externalLoadBalancerConfig

Went with the dedicated-struct option in 1be8c161. ExternalLoadBalancerConfig is now typed as a new ExternalLoadBalancer struct that only exposes name and ipAddress, so subnet / internalAccess no longer show up under externalLoadBalancerConfig in any of the four CRDs. Since the field is new in this PR (introduced in 654021fa, not merged yet), there's no upgrade path to worry about.



> CRD description for externalLoadBalancerConfig is truncated
> 
> In all four CRD files, the description ends mid-sentence: RegionalExternal (regional), and RegionalInternalExternal (regional external Missing the closing: + regional internal). Ignored when LoadBalancerType is Internal.
> 
> This appears to be a controller-gen line-length truncation issue in the Go source godoc.

Nice find — the issue was a line starting with + in the godoc, which controller-gen interprets as a marker and stops rendering. Rewrapped the paragraph in 1be8c161 so + regional internal) no longer starts a line, and regenerated the CRDs. The description now ends correctly with Ignored when LoadBalancerType is Internal. in all four CRD files.

cloud/services/compute/loadbalancers/backward_compat_test.go
DvdChe: Renamed to `shouldCreateGlobalExternalLoadBalancer` in 2c411aad (and the corresponding test TestShouldCreateGlobalExternalLoadBalancer). Makes it clear at the call sites that the regional types are intentionally excluded.

docs/proposals/cors-4448/CORS-4448-quick-reference.md
DvdChe: Renamed in 644cf39b. The directory is now docs/proposals/regional-load-balancer-support/, and the three files were renamed to proposal.md, architecture.md, and quick-reference.md. Headings inside the files were updated and the Jira link at the bottom of the proposal was removed.

barbacbd (commented):
@DvdChe Can you refactor this PR? This is currently 19 commits and a lot of similar files changed. A lot of refactors too.

theobarberbany:
Looks like we've also updated .gitignore with claude related changes, might be worth taking that out too :)

DvdChe:
@barbacbd Done — reworked the history down to 4 commits:

  1. feat(lb): support RegionalExternal and RegionalInternalExternal load balancers
  2. feat(api): add ExternalLoadBalancerConfig with admission validation
  3. fix(lb): follow-up refinements to regional load balancer
  4. docs: add regional load balancer support proposal

 The cumulative diff is unchanged from the previous state of the branch. I also dropped an accidental `.claude/` ( cc @theobarberbany ) entry that had crept into .gitignore. Force-pushed.

PR #1637, opened 2026-04-22
Support `clusterSecondaryRangeName` and `servicesSecondaryRangeName` in `GCPManagedControlPlane`

### LINKED ISSUE #1636: Support `clusterSecondaryRangeName` and `servicesSecondaryRangeName` in `GCPManagedControlPlane`
/kind feature
/kind api-change

When provisioning VPC-native GKE clusters (`useIPAliases: true`), it is common to pre-allocate named secondary IP ranges on a subnet and share that subnet across multiple clusters. The GCP API exposes `clusterSecondaryRangeName` and `servicesSecondaryRangeName` on I`PAllocationPolicy` for exactly this purpose, but `GCPManagedControlPlane` currently only supports specifying pod and service ranges by CIDR block (`cidrBlock`), not by range name.

The proposed addition is two optional fields on the existing ClusterNetworkPod and ClusterNetworkService structs:

```yaml
spec:
  clusterNetwork:
    useIPAliases: true
    pod:
      secondaryRangeName: pods-range
    service:
      secondaryRangeName: services-range

```
Both fields would be immutable after cluster creation (matching GCP API behaviour) and require useIPAliases: true.

I've had a go at a PR to support this, but appreciate it's my first time contributing. I'll link it when I raise the PR!

jonathanrainer:
**What type of PR is this?**
/kind feature
/kind api-change

This PR adds two optional fields to the existing `ClusterNetworkPod` and `ClusterNetworkService` structs, namely `secondaryRangeName` in each case. This means we can pass these options through to the API when we make the call to create the cluster.

This allows us allocate secondary IP ranges on a subnet and then share that subnet across multiple clusters, or more precisely segment your IP address range.

Fixes #1636

This is my first PR so if I've got stuff wrong/not in line with the codebase, many apologies, more than happy to fix stuff up!

**TODOs**:
- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Adds optional `secondaryRangeName` fields to `ClusterNetwork.Pod` and `ClusterNetwork.Service` on `GCPManagedControlPlane`, mapping to the GKE API's `clusterSecondaryRangeName` and `servicesSecondaryRangeName`. This lets clusters consume pre-existing secondary IP ranges on their subnetwork, enabling subnet sharing across multiple clusters or finer-grained IP segmentation. Both fields require `useIPAliases: true` and are mutually exclusive with the corresponding `cidrBlock`.
```


cloud/services/container/clusters/reconcile.go
```diff
@@ -288,9 +288,15 @@ func (s *Service) createCluster(ctx context.Context, log *logr.Logger) error {
 			cluster.IpAllocationPolicy.UseIpAliases = cn.UseIPAliases
 			if cn.Pod != nil {
 				cluster.IpAllocationPolicy.ClusterIpv4CidrBlock = cn.Pod.CidrBlock
+				if cn.Pod.SecondaryRangeName != nil {
+					cluster.IpAllocationPolicy.ClusterSecondaryRangeName = *cn.Pod.SecondaryRangeName
```
barbacbd: Does specifying the Secondary Range cause potential issues in GCP when CidrBlock is also specified?

exp/webhooks/gcpmanagedcontrolplane_webhook_test.go
```diff
@@ -184,6 +185,46 @@ func TestGCPManagedControlPlaneValidatingWebhookCreate(t *testing.T) {
 				Version:             &vV1_32_5,
 			},
 		},
+		{
+			name:        "pod secondary range name without UseIPAliases should cause an error",
```
barbacbd, line 249: Can we add a couple of tests:
1. Setting a secondary range when it was previously nil/unset
2. Removing a secondary range when previously set (opposite of 1)

exp/api/v1beta1/gcpmanagedcontrolplane_types.go
```diff
@@ -75,6 +80,11 @@ type ClusterNetworkService struct {
 	// This setting is permanent.
 	// +optional
 	CidrBlock string `json:"cidrBlock,omitempty"`
+
+	// SecondaryRangeName is the name of the existing secondary range in the cluster's subnetwork to
```
barbacbd, line 85:
```suggestion
  // SecondaryRangeName is the name of the existing secondary range in the cluster's subnetwork to
  // use for pod IP addresses. The range must already exist on the subnet. Only applicable if                                                                                               
  // UseIPAliases is true. This setting is permanent.
```

jonathanrainer: Fixed this up here and made a similar change to the one above as the context is similar enough.

exp/webhooks/gcpmanagedcontrolplane_webhook_test.go
jonathanrainer: Done! Thanks for spotting the gap :)

cloud/services/container/clusters/reconcile.go
jonathanrainer: Ah yes it does, they're mutually exclusive in the API itself so I've added that validation to the webhooks so we don't end up submitting invalid API calls

barbacbd:
@damdo @salasberryfin What do you think about adding release notes ? Are we good or should these be updated?

exp/webhooks/gcpmanagedcontrolplane_webhook.go
```diff
@@ -126,6 +126,36 @@ func (*GCPManagedControlPlane) ValidateCreate(_ context.Context, obj runtime.Obj
...
+				))
+			}
+			if cn.Service != nil && cn.Service.SecondaryRangeName != nil {
+				allErrs = append(allErrs, field.Invalid(
+					field.NewPath("spec", "clusterNetwork", "service", "secondaryRangeName"),
+					cn.Service.SecondaryRangeName, "secondaryRangeName requires useIPAliases to be true",
+				))
+			}
+		}
+		if cn.Pod != nil && cn.Pod.CidrBlock != "" && cn.Pod.SecondaryRangeName != nil {
```
barbacbd: `*cn.Pod.SecondaryRangeName == ""` - check if the dereferenced pointer value is empty string

jonathanrainer: Thanks for spotting! Had another look at this and realised I could just use `ptr.Deref` and then a lot of this becomes unnecessary. What's there now is a lot better!

exp/webhooks/gcpmanagedcontrolplane_webhook.go
```diff
@@ -126,6 +127,42 @@ func (*GCPManagedControlPlane) ValidateCreate(_ context.Context, obj runtime.Obj
 		}
 	}
 
+	if cn := r.Spec.ClusterNetwork; cn != nil {
+		var podSecondaryRange, serviceSecondaryRange string
+		if cn.Pod != nil {
+			podSecondaryRange = ptr.Deref(cn.Pod.SecondaryRangeName, "")
+		}
+		if cn.Service != nil {
+			serviceSecondaryRange = ptr.Deref(cn.Service.SecondaryRangeName, "")
```
barbacbd: For things like this we probably need to Either reject explicit empty strings or document that "" and nil are equivalent. WDYT @salasberryfin @damdo ?

barbacbd: Invalid names currently accepted:
  - "123-invalid" (starts with number)
  - "Invalid" (uppercase) 
  - "my-range-" (ends with hyphen)

  Recommendation: Add format validation:
  const secondaryRangeNameRegex = `^[a-z]([a-z0-9-]{0,61}[a-z0-9])?$`

cloud/services/container/clusters/reconcile_test.go
```diff
@@ -540,6 +540,90 @@ func TestConvertToSdkBinaryAuthorizationEvaluationMode(t *testing.T) {
 	}
 }
 
+func TestCreateClusterSecondaryRangeNames(t *testing.T) {
```
barbacbd, line 653: TestCreateClusterSecondaryRangeNames manually duplicates the logic instead of calling createCluster(). Consider testing the actual function for better integration coverage.

barbacbd (commented):
nit: can we create documentation on how to use Secondary Range?

exp/webhooks/gcpmanagedcontrolplane_webhook.go
jonathanrainer: @barbacbd I refactored it a bit so this kind of becomes a moot point, I think explicitly rejecting the empty string is a better bet. I've also added a validation function and drawn from apimachinery so that the invalid names are rejected

cloud/services/container/clusters/reconcile_test.go
jonathanrainer: Thanks for spotting that! I've refactored it so that we're now calling createCluster instead alongside the functionality we want to test. Let me know if there's more you wanted there

jonathanrainer:
> nit: can we create documentation on how to use Secondary Range?

Where would you want to add the docs? I've added to the `book` folder but is there somewhere else it should go?

damdo:
/assign @jonathanrainer 

For rebase

PR #1638, opened 2026-04-25
Add WorkloadMetadataMode to GKE node pool security config

jonathanrainer:
**What type of PR is this?**

/kind feature
/kind api-change

Adds a `workloadMetadataMode` field to `NodeSecurityConfig` on `GCPManagedMachinePool` that controls how GCE/GKE metadata is exposed to workloads running on the node pool. This allows operators to enforce `GKE_METADATA` mode (which enables Workload Identity on node pools) or explicitly select `GCE_METADATA` mode.

Changes:
- New `WorkloadMetadataMode` type with `GKE_METADATA` and `GCE_METADATA` values
- `ConvertToSdkWorkloadMetadataMode` conversion helper
- `WorkloadMetadataMode` wired into the SDK node pool config in `ConvertToSdkNodePool`
- Reconciliation of `WorkloadMetadataConfig` during node pool updates
- The field is **mutable** (unlike most `nodeSecurity` fields) — the webhook previously marked all of `nodeSecurity` as immutable; this PR splits that check to keep the other fields immutable while allowing `workloadMetadataMode` to be changed after creation

The key behavioural change in the webhook is that `workloadMetadataMode` is now treated as mutable. The GKE API supports updating this field on existing node pools, so immutability was unnecessarily restrictive.

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Add `workloadMetadataMode` field to `GCPManagedMachinePool.spec.nodeSecurity` to control GKE/GCE metadata exposure on node pools. Valid values are `GKE_METADATA` and `GCE_METADATA`. The field is mutable after creation.
```


exp/api/v1beta1/types.go
```diff
@@ -127,6 +127,20 @@ func ConvertToSdkCgroupMode(cgroupMode ManagedNodePoolCgroupMode) containerpb.Li
...
+func ConvertToSdkWorkloadMetadataMode(mode *WorkloadMetadataMode) *containerpb.WorkloadMetadataConfig {
+	if mode == nil {
+		return nil
+	}
+	switch *mode {
+	case WorkloadMetadataModeGKEMetadata:
+		return &containerpb.WorkloadMetadataConfig{Mode: containerpb.WorkloadMetadataConfig_GKE_METADATA}
+	case WorkloadMetadataModeGCEMetadata:
+		return &containerpb.WorkloadMetadataConfig{Mode: containerpb.WorkloadMetadataConfig_GCE_METADATA}
+	}
```
barbacbd, line 143: Add a default warning or failure when the mode is not found. This can cause silent failures especially if the enum is expanded.

cloud/services/container/nodepools/reconcile.go
```diff
@@ -409,6 +409,12 @@ func (s *Service) checkDiffAndPrepareUpdateConfig(existingNodePool *containerpb.
 		needUpdate = true
 		updateNodePoolRequest.LinuxNodeConfig = desiredLinuxNodeConfig
 	}
+	// WorkloadMetadataConfig
+	desiredWorkloadMetadataConfig := infrav1exp.ConvertToSdkWorkloadMetadataMode(s.scope.GCPManagedMachinePool.Spec.NodeSecurity.WorkloadMetadataMode)
+	if !cmp.Equal(desiredWorkloadMetadataConfig, existingNodePool.GetConfig().GetWorkloadMetadataConfig(), cmpopts.IgnoreUnexported(containerpb.WorkloadMetadataConfig{})) {
```
barbacbd: the comparison should be nil protected. This is done in managedmachinepool.go. Below is the claude analysis:

```
The reconciliation logic will trigger an update on **every reconcile loop** when `workloadMetadataMode` is not set by the user, because it compares `nil` (from conversion) against whateve\
r the existing node pool has.

**Why This Happens:**
- When user doesn't set `workloadMetadataMode`, it's nil
- `ConvertToSdkWorkloadMetadataMode(nil)` returns `nil`
- Many GKE clusters have Workload Identity enabled by default, so existing node pool has `GKE_METADATA` set
- `cmp.Equal(nil, &WorkloadMetadataConfig{Mode: GKE_METADATA})` returns `false`
- Update is triggered on every reconcile

**Impact:**
- Continuous reconciliation loop
- Potentially sends nil WorkloadMetadataConfig to GCP API (could error or reset config)
- Wasted API calls and cluster churn
- Inconsistent behavior between creation (which has nil guard) and reconciliation (which doesn't)

```

exp/api/v1beta1/types.go
```diff
@@ -127,6 +127,20 @@ func ConvertToSdkCgroupMode(cgroupMode ManagedNodePoolCgroupMode) containerpb.Li
 	return containerpb.LinuxNodeConfig_CGROUP_MODE_UNSPECIFIED
 }
 
+// ConvertToSdkWorkloadMetadataMode converts a WorkloadMetadataMode to the GCP SDK WorkloadMetadataConfig.
+func ConvertToSdkWorkloadMetadataMode(mode *WorkloadMetadataMode) *containerpb.WorkloadMetadataConfig {
```
barbacbd, line 132: Can we unit test this function ?

cloud/services/container/nodepools/reconcile.go
barbacbd: This new functionality should be captured in unit tests and/or E2E tests.

exp/api/v1beta1/types.go
jonathanrainer: Done!

cloud/services/container/nodepools/reconcile.go
jonathanrainer: Oh yer, sorry I missed that one. Have fixed that in latest commit

exp/api/v1beta1/types.go
jonathanrainer: Yup have added that!

cloud/services/container/nodepools/reconcile.go
jonathanrainer: So I've added new unit tests, I've had a think but I'm not 100% sure on the efficacy of the E2E tests. We're basically just testing that we can set and read back the value from the GKE API, it feels like we'd be testing the GKE API rather than our own stuff. That said if you think they'd be useful to exercise some good surface area I'm more than happy to add some in.

barbacbd (commented):
LGTM @damdo WDYT?

salasberryfin (commented):
/ok-to-test

PR #1639, opened 2026-04-25
Support Workload Identity Federation for CAPG controller authentication

### LINKED ISSUE #311: Support Application Default Credentials or Workload Identity when running in GCP
/kind feature

Exporting service account credentials and storing in K8S secret store introduces an unnecessary secret and potential point of compromise if the cluster API provider is running inside GCP.

In this case, the provider should first look for Application Default Credentials or use workload identity before requiring service account credential keys.

jonathanrainer:
**What type of PR is this?**

/kind feature

Enables the CAPG controller to authenticate with GCP using [Workload Identity Federation (WIF)](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) when running on GKE, eliminating the need to mount JSON service account key files.

**Code changes:**

- `cloud/scope/credentials.go`: `getCredentialDataUsingADC()` now returns `nil, nil` (instead of an error) when `GOOGLE_APPLICATION_CREDENTIALS` is not set, signalling to callers that implicit ADC (including WIF via the GKE metadata server) should be used. `getCredentials()` short-circuits on nil credential data and returns `nil, nil` accordingly.
- `cloud/services/container/clusters/kubeconfig.go`: `generateToken()` falls back to `metadata.EmailWithContext()` to discover the bound GCP service account email from the GKE metadata server when no explicit credential is present, allowing kubeconfig generation to work under WIF.

**Config changes:**

- `config/wif/`: new self-contained kustomize overlay that omits the credentials Secret and volume mount, and annotates the manager `ServiceAccount` with `iam.gke.io/gcp-service-account: ${GCP_SA_EMAIL}`. Operators substitute `GCP_SA_EMAIL` via `envsubst` before applying.

**Documentation:**

- `docs/book/src/quick-start.md`: restructures the credentials section to cover both the existing JSON key approach and the new WIF approach, including the GCP IAM setup steps.
- `docs/book/src/managed/enabling.md`: clarifies that the `iam.serviceAccountTokenCreator` role requirement applies to the WIF-bound GCP service account.

Fixes #311

- The existing JSON key / `GOOGLE_APPLICATION_CREDENTIALS` path is **fully preserved** — this change only affects behaviour when neither `credentialsRef` nor the env var is set.
- `cloud.google.com/go/compute/metadata` was already an indirect dependency; it is promoted to direct.
- `metadata.EmailWithContext` is only called in `generateToken` (GKE managed cluster kubeconfig path), not in the general credential loading path.

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Add support for Workload Identity Federation (WIF) as an authentication method for the CAPG controller on GKE management clusters. A new `config/wif` kustomize overlay is provided; no JSON service account key is required when WIF is configured.
```

### DIFF
--- a/cloud/scope/credentials.go
+++ b/cloud/scope/credentials.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	infrav1 "sigs.k8s.io/cluster-api-provider-gcp/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
@@ -44,6 +45,7 @@ type Credential struct {
 }
 
 func getCredentials(ctx context.Context, credentialsRef *infrav1.ObjectReference, crClient client.Client) (*Credential, error) {
+	logger := log.FromContext(ctx)
 	var credentialData []byte
 	var err error
 
@@ -55,8 +57,19 @@ func getCredentials(ctx context.Context, credentialsRef *infrav1.ObjectReference
 	if err != nil {
 		return nil, fmt.Errorf("getting credential data: %w", err)
 	}
+	if credentialData == nil {
+		// No explicit credentials configured; the GCP client libraries will use
+		// implicit ADC (e.g. Workload Identity Federation via the GKE metadata server).
+		logger.Info("No explicit credentials found; using implicit ADC (e.g. Workload Identity Federation)")
+		return nil, nil
+	}
 
-	return parseCredential(credentialData)
+	cred, err := parseCredential(credentialData)
+	if err != nil {
+		return nil, err
+	}
+	logger.Info("Loaded explicit GCP credentials", "credentialType", cred.Type)
+	return cred, nil
 }
 
 func getCredentialDataFromRef(ctx context.Context, credentialsRef *infrav1.ObjectReference, crClient client.Client) ([]byte, error) {
@@ -81,7 +94,9 @@ func getCredentialDataFromRef(ctx context.Context, credentialsRef *infrav1.Objec
 func getCredentialDataUsingADC() ([]byte, error) {
 	credsPath := os.Getenv(ConfigFileEnvVar)
 	if credsPath == "" {
-		return nil, fmt.Errorf("no ADC environment variable found for credentials (expect %s)", ConfigFileEnvVar)
+		// No explicit credentials file configured; signal to callers to use
+		// implicit ADC (Workload Identity Federation, instance metadata, etc.).
+		return nil, nil
 	}
 
 	byteValue, err := os.ReadFile(credsPath) //nolint:gosec // We need to read a file here
--- a/cloud/services/container/clusters/kubeconfig.go
+++ b/cloud/services/container/clusters/kubeconfig.go
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"fmt"
 
+	"cloud.google.com/go/compute/metadata"
 	"cloud.google.com/go/container/apiv1/containerpb"
 	"cloud.google.com/go/iam/credentials/apiv1/credentialspb"
 	"github.com/go-logr/logr"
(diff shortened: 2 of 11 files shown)


cloud/services/container/clusters/kubeconfig.go
```diff
@@ -238,11 +239,22 @@ func (s *Service) createBaseKubeConfig(contextName string, cluster *containerpb.
 }
 
 func (s *Service) generateToken(ctx context.Context) (string, error) {
```
barbacbd, line 263: Can we write tests here for WIF mode and basic credentials?

cloud/scope/credentials.go
```diff
@@ -55,6 +55,11 @@ func getCredentials(ctx context.Context, credentialsRef *infrav1.ObjectReference
 	if err != nil {
 		return nil, fmt.Errorf("getting credential data: %w", err)
 	}
+	if credentialData == nil {
```
barbacbd, line 60: Can you add a logging statement that states what type of creds are being loaded? Nothing detailed about the creds but what type they are .

cloud/services/container/clusters/kubeconfig.go
jonathanrainer: Have done this, I had to refactor the methods a bit and so introduced a new interface that makes the tests a bit easier to write. Not the cleanest but it gets the job done

cloud/scope/credentials.go
jonathanrainer: This is all done, I think I've used the right logging idioms but let me know if I got this wrong, I'm still learning!

salasberryfin (commented):
/ok-to-test

abasu0713:
can we get this merged?

cloud/services/container/clusters/kubeconfig.go
```diff
@@ -237,12 +238,36 @@ func (s *Service) createBaseKubeConfig(contextName string, cluster *containerpb.
...
+type metadataEmailResolver struct{}
+
+func (r metadataEmailResolver) Email(ctx context.Context) (string, error) {
+	return metadata.EmailWithContext(ctx, "default")
+}
+
 func (s *Service) generateToken(ctx context.Context) (string, error) {
+	email, err := s.emailResolver.Email(ctx)
+	if err != nil {
+		return "", fmt.Errorf("fetching service account email from metadata server: %w", err)
```
barbacbd: nit: Error message in generateToken is slightly misleading

  return "", fmt.Errorf("fetching service account email from metadata server: %w", err)

  This wraps any emailResolver.Email() error with "from metadata server", but credentialEmailResolver.Email() never errors. Not a real bug (it's unreachable for that path), but the message
  would be confusing if credentialEmailResolver ever did error. Consider a more generic message like "resolving service account email: %w".

config/wif/kustomization.yaml
```diff
@@ -0,0 +1,55 @@
+namespace: capg-system
```
barbacbd: nit: The config/wif/ overlay duplicates a lot from config/default/

  The kustomize overlay replicates most of config/default/kustomization.yaml (bases, vars, configurations, webhook patches) minus the credentials patches. This means any future changes to
  config/default/ would need to be mirrored manually. A comment or note about this coupling in config/wif/kustomization.yaml would help maintainability. Alternatively, config/wif/ could
  potentially be based on config/default/ with a strategic merge patch that removes the credential volume/mount.

barbacbd:
@jonathanrainer My nit comments shouldn't hold up the PR merge process unless someone else finds them more of an issue.

jonathanrainer:
> @jonathanrainer My nit comments shouldn't hold up the PR merge process unless someone else finds them more of an issue.

No worries, I'm more than happy to get them fixed up :)

cloud/services/container/clusters/kubeconfig.go
jonathanrainer: Moved to the generic message, thanks for the suggestion!

config/wif/kustomization.yaml
jonathanrainer: Yer this was silly, I've moved to strategic patches on top of the default so it's a lot less code now!

abasu0713:
> can we get this merged?

@barbacbd @jonathanrainer .. First off I really appreciate you guys pushing this along.. I wanted to share some context - I am working on a hub spoke model for cluster management. This would really help without me having to add a non-CR way of minting service account credentials.. 
I am sorry I don't mean to rush you.. But if you could give me a timeline, when you can get this merged and what release this would be a part of - I would greatly appreciate it.  :)

salasberryfin (commented):
/lgtm
/approve

PR #1646, opened 2026-04-30
fix(gke): apply additionalLabels to GKE Cluster ResourceLabels

### LINKED ISSUE #1645: `additionalLabels` on `GCPManagedCluster` not applied to GKE cluster ResourceLabels
/kind bug

`GCPManagedClusterSpec.AdditionalLabels` is documented as setting labels on GCP resources managed by the provider, but for GKE clusters the field is silently ignored. Neither the cluster creation path nor any reconciliation loop reads this value.

Steps:
1. Set up a GCPManagedCluster CR and set AdditionalLabels to a map where `env: prod` is set
2. Let Cluster API create the Cluster

The label `env:prod` should be attached to the cluster in GCP.

This was partially addressed by #1275, which wired additionalLabels to disks and forwarding rules. However that fix was scoped to the non-managed (compute) path. The managed Kubernetes path (cloud/services/container/clusters/) was not updated.

**Environment:**

- Cluster-api version: Operator (0.26.0)
- Minikube/KIND version: N/A

jonathanrainer:
**What type of PR is this?**
/kind bug

`GCPManagedCluster.Spec.AdditionalLabels` was a no-op for GKE clusters. This was partially fixed for compute resources in #1275 but the managed Kubernetes path (`cloud/services/container/clusters/`) was missed.

This PR wires `additionalLabels` through in two places:

Creation — ResourceLabels is now set on the `containerpb.Cluster` struct passed to `CreateCluster`.
Reconciliation — a new `syncLabels` step calls `SetLabels` when the live cluster's `ResourceLabels` diverge from the spec. This runs only when no other UpdateCluster operation is pending, to avoid GKE's restriction on concurrent cluster mutations.

Fixes #1645 

Label management is opt-in: if additionalLabels is not set on the spec the controller takes no action, leaving any manually applied GCP labels untouched. This is a deliberate safety property for rollout — without it, existing clusters with GCP labels set outside of CAPI would have those labels silently wiped on the next reconcile.

When additionalLabels is set, the field acts as the complete desired label set for ResourceLabels. The GKE SetLabels API replaces all user-defined labels, so labels applied directly to GCP that are not present in the spec will be removed. This is consistent with CAPI's source-of-truth philosophy. This position is defensible today because the provider does not impose any mandatory labels on GKE clusters; if mandatory labels are introduced in future the opt-in decision should be revisited.

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests (marked ok as new tests are unnecessary and existing suites continue to pass)

**Release note**:
```release-note
Allows GCPManagedCluster.Spec.AdditionalLabels to set ResourceLabels on GKE Clusters
```

### DIFF
--- a/cloud/services/container/clusters/reconcile.go
+++ b/cloud/services/container/clusters/reconcile.go
@@ -31,6 +31,7 @@ import (
 	"github.com/googleapis/gax-go/v2/apierror"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc/codes"
+	infrav1 "sigs.k8s.io/cluster-api-provider-gcp/api/v1beta1"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-gcp/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-gcp/util/reconciler"
 	clusterv1beta1 "sigs.k8s.io/cluster-api/api/core/v1beta1"
@@ -154,6 +155,11 @@ func (s *Service) Reconcile(ctx context.Context) (ctrl.Result, error) {
 		s.scope.GCPManagedControlPlane.Status.Ready = true
 		return ctrl.Result{}, nil
 	}
+
+	if err = s.syncLabels(ctx, cluster, &log); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	v1beta1conditions.MarkFalse(s.scope.ConditionSetter(), infrav1exp.GKEControlPlaneUpdatingCondition, infrav1exp.GKEControlPlaneUpdatedReason, clusterv1beta1.ConditionSeverityInfo, "")
 
 	// Reconcile kubeconfig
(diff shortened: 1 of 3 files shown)


cloud/services/container/clusters/reconcile.go
```diff
@@ -408,6 +415,35 @@ func (s *Service) deleteCluster(ctx context.Context, log *logr.Logger) error {
 	return nil
 }
 
+// labelsNeedSync returns true if the desired labels differ from the existing cluster labels.
+// Returns false when desired is nil or empty, making label management opt-in: the controller
+// only manages ResourceLabels when additionalLabels is explicitly set on the spec.
+func labelsNeedSync(existing, desired infrav1.Labels) bool {
```
barbacbd, line 426: There is a limit to the number of labels that can be applied. Should we change this to an API change where the extra tags can be set/controlled by the user/caller?

jonathanrainer: Not quite sure I understand what you mean, my feeling here was that I didn't want to overwrite labels unless people did something to explicitly opt in. Are you saying that this API should just always replace the labels no matter what?

salasberryfin (commented):
/ok-to-test

salasberryfin (commented):
This looks good to me but I'll hold it for @barbacbd (and others) to take a second look at it.

/lgtm
/hold

PR #1648, opened 2026-04-30
fix(gke): treat `NotFound` as success in `deleteNodePool`

### LINKED ISSUE #1647: `GCPManagedMachinePool` finalizer gets stuck on cluster deletion, requiring manual removal
/kind bug

When deleting a managed GKE cluster, the `GCPManagedMachinePool` object frequently gets stuck in termination with its finalizer never removed, even after the GCP resources are confirmed gone. The only workaround is to manually kubectl edit the object and delete the finalizer.

The finalizer would be removed when the GCP NodePool disappears and the cascading deletion of the other dependent resources would continue with no manual intervention

When a GKE cluster is deleted, GCP automatically deletes all associated node pools. CAPG independently tries to call DeleteNodePool for each pool. If GCP has already cleaned up the node pool by the time CAPG issues its request, the API returns NotFound.

`deleteNodePool()` in cloud/services/container/nodepools/reconcile.go does not handle NotFound — it propagates the error, which sets GKEMachinePoolDeletingCondition to `ReconciliationFailed`. The finalizer removal gate in the controller checks specifically for DeletedReason, so the finalizer is never removed and the object is stuck until manually unblocked.

describeNodePool() in the same file already handles NotFound correctly (treating it as a nil return). deleteNodePool() should follow the same pattern.

Further it seems that the E2E tests, don't explicitly assert on deletion semantics so this is a small coverage gap in the E2Es

**Environment:**

- Cluster-api version: 0.26.0 (Cluster API Operator)
- Minikube/KIND version: N/A

jonathanrainer:
/kind bug

When a GKE cluster is deleted, GCP automatically removes all its node pools. If CAPG then calls DeleteNodePool for a pool GCP has already cleaned up, the API returns NotFound. Previously this error was propagated back to the controller, which set `GKEMachinePoolDeletingCondition` to `ReconciliationFailed` rather than `Deleted`. The finalizer removal gate in `gcpmanagedmachinepool_controller.go` checks specifically for DeletedReason, so the GCPManagedMachinePool finalizer was never removed — leaving the object stuck in termination indefinitely and requiring a manual kubectl edit to unblock.

The fix mirrors the pattern already used in describeNodePool: treat a NotFound response from DeleteNodePool as a no-op success, allowing the deletion condition to reach DeletedReason on the next reconcile loop when describeNodePool confirms the pool is gone.

Fixes #1647 

As well as the fix, have added a new E2E test that explicitly creates a cluster, deletes it, and then asserts that all managed resources (MachinePool, GCPManagedMachinePool, GCPManagedControlPlane, GCPManagedCluster) are fully removed from the API server within the wait-delete-cluster interval (currently 30m in the default CI config). Per-type assertions are used rather than relying solely on the top-level Cluster object disappearing, so a failure message identifies exactly which resource type is blocked.

Previously, deletion was only exercised in AfterEach cleanup blocks where a timeout would not produce a clear test failure.

**TODOs**:
- [x] squashed commits
- [x] includes documentation - not really necessary as this should be the expected behaviour already
- [x] adds unit tests

**Release note**:
```release-note
fix(gke): ensure finalizers are removed from `GCPManagedMachinePool` (it gets removed), even when the nodepool is automatically deleted
```

### DIFF
--- a/cloud/services/container/clusters/reconcile.go
+++ b/cloud/services/container/clusters/reconcile.go
@@ -28,9 +28,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-	"github.com/googleapis/gax-go/v2/apierror"
 	"github.com/pkg/errors"
-	"google.golang.org/grpc/codes"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-gcp/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-gcp/util/reconciler"
 	clusterv1beta1 "sigs.k8s.io/cluster-api/api/core/v1beta1"
@@ -231,11 +229,8 @@ func (s *Service) describeCluster(ctx context.Context, log *logr.Logger) (*conta
 	}
 	cluster, err := s.scope.ManagedControlPlaneClient().GetCluster(ctx, getClusterRequest)
 	if err != nil {
-		var e *apierror.APIError
-		if ok := errors.As(err, &e); ok {
-			if e.GRPCStatus().Code() == codes.NotFound {
-				return nil, nil
-			}
+		if shared.IsNotFound(err) {
+			return nil, nil
 		}
 		log.Error(err, "Error getting GKE cluster", "name", s.scope.ClusterName())
 		return nil, err
--- a/cloud/services/container/nodepools/reconcile.go
+++ b/cloud/services/container/nodepools/reconcile.go
@@ -24,14 +24,12 @@ import (
 	"sigs.k8s.io/cluster-api-provider-gcp/util/resourceurl"
 
 	"google.golang.org/api/iterator"
-	"google.golang.org/grpc/codes"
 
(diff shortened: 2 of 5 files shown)


barbacbd (commented):
I think this looks good @damdo can you take a look? my only nit would be maybe we can make or use a common function for Error Not Found.

jonathanrainer:
> I think this looks good @damdo can you take a look? my only nit would be maybe we can make or use a common function for Error Not Found.

Had a go at this, let me know if it was what you were expecting :)

barbacbd:
@jonathanrainer This does look good to me but the final commit should probably be a fixup or at least a squashed commit into the first.

jonathanrainer:
> @jonathanrainer This does look good to me but the final commit should probably be a fixup or at least a squashed commit into the first.

Thanks for having a look, I've sorted that out now :)

salasberryfin (commented):
/ok-to-test

damdo (commented):
/approve

Thanks!!

damdo:
/assign @salasberryfin 

for LGTM

salasberryfin (commented):
Thanks @jonathanrainer.

/lgtm
/approve

PR #1649, opened 2026-04-30
[release-1.10] chore(deps): bump dependencies from #1633

damdo:
Bump the subset of dependencies from #1633 that don't require Go 1.25/1.24

/kind other

**Release note**:

```release-note
NONE
```


cpanato (approved):
thanks

PR #1653, opened 2026-05-06
docs: fix release guide gcloud config and registry path

damdo:
- Add a note about activating the correct gcloud configuration before login when multiple configurations exist
- Fix the production registry path from `registry.k8s.io/cluster-api-provider-gcp` to `registry.k8s.io/cluster-api-gcp`


salasberryfin (commented):
/lgtm
/approve

PR #1657, opened 2026-05-07
Stop `GCPManagedNodePool` scaling being blocked by multiple factors

### LINKED ISSUE #1656: `GCPManagedMachinePool` scaling via `MachinePool.spec.replicas` is silently ignored
/kind bug

1. Created a GCPManagedMachinePool without an explicit spec.scaling block.
2. Waited for the node pool to reach RUNNING state and check the correct number of replicas was set in GCP.
3. Edit MachinePool.spec.replicas to a different value.
4. Observe: the controller reconciles repeatedly but the node pool size does not change.

You see a loop in the logs of 4 log lines
```
"Reconciling node pool resources"
"Node pool running"
"Node pool config update required" request="...resource_labels:{labels:{key:\"capg-cluster-<name>\" value:\"owned\"}}"
"Node pool config updating in progress"
```

The GCP nodepool would scale to the correct number of replicas

There seem to be three issues that compound on each other to cause this to happen:

1. There's a perpetual diff on `resource_labels`, so when the NodePool gets created the controller injects `capg-cluster-<name>` into `NodeConfig.ResourceLabels`. However, the semantics of this mean that that label is set on the individual VMs not on to the NodePool object itself. As such `checkDiffAndPrepareUpdateConfig` returns a diff each time and returns early, calling `updateNodePool` which is futile, because [NodePool](https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters.nodePools?hl=en#NodePool) doesn't have any labels, and so it loops endlessly.
 - If this was to be fixed so it worked properly it would need an additional API call to get the underlying `InstanceTemplate`, check the labels on that and then report back rather than relying on checking the NodePool object itself
2. In a related vein, `ConvertToSdkLinuxNodeConfig`, when called with a `nil` value, always produces a non-nil empty struct, which always differs from the result of the GKE API for pools that have no Linux Node config set. Which triggers the same bug.
3. The three update checks that are part of `checkDiffAndPrepareUpdateConfig` each return early if they find a difference. So the first point, starves the autoscaling and size checks of ever running
4. `ConvertSdkAutoscaling`, when called with a `nil` value, produces a struct with `Enabled: true`. As such, any size updates are skipped silently because the autoscaling guard returns early. As such if you want to do manual scaling, you're forced to explicitly set `spec.scaling.enableAutoscaling: false`. 
All of this compounds together to produce a constantly reconciling controller, and the ability to manually scale a nodepool effectively blocked.

**Environment:**

- Cluster-api version: v0.26.0 (Operator)
- Minikube/KIND version: N/A

jonathanrainer:
**What type of PR is this?**
/kind bug

This PR tackles the 4 bugs that are listed out in #1656 which prevent manual `MachinePool` scaling from working with a `GCPManagedNodePool`. Namely:

1. Add a `nil` guard to the Autoscaling check - This means that nodepools that don't have an Autoscaling config won't return early, so manual scaling can occur
2. Add a `nil` guard to `LinuxNodeConfig` - This means that `LinuxNodeConfig` is only compared when it's set, rather than comparing a non-nil empty struct generated by `ConvertToSdkLinuxNodeConfig` to the `nil` that GKE returns
3. Sync `AdditionalLabels` via the InstanceTemplate not the NodePool - The NodePool object doesn't have a labels attribute on it, so instead of that we create an InstanceTemplate client, look up the template that's relevant and check the labels we're setting are correctly set there, and rely on GKE to propagate those into the nodes themselves. 
4. Run all three checks (`checkDiffAndPrepareUpdateSize`, `checkDiffAndPrepareUpdateAutoscaling`, `checkDiffAndPrepareUpdateConfig`) before any update is applied - This means the current behaviour where updates are run sequentially and there's an early exit (and in turn causes size updates to be starved because of (3)), doesn't happen anymore. Updates are all checked and then run in priority order (size -> autoscaling -> general).

Fixes #1656

I've verified as much of this as I can on my own cluster, particularly the `additionalLabels` checks, but I also appreciate there's an alternative approach where you could fetch the Instances themselves and check, and I'm happy to pivot to that approach if we think that's demonstrably better.

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Ensure that GCPManagedNodePool scaling always works and improve how labels are synced to NodePools
```

### DIFF
--- a/cloud/scope/clients.go
+++ b/cloud/scope/clients.go
@@ -210,6 +210,24 @@ func newInstanceGroupManagerClient(ctx context.Context, credentialsRef *infrav1.
 	return instanceGroupManagersClient, nil
 }
 
+func newInstanceTemplatesRESTClient(ctx context.Context, credentialsRef *infrav1.ObjectReference, crClient client.Client, endpoints *infrav1.ServiceEndpoints) (*computerest.InstanceTemplatesClient, error) {
+	opts, err := defaultClientOptions(ctx, credentialsRef, crClient)
+	if err != nil {
+		return nil, fmt.Errorf("getting default gcp client options: %w", err)
+	}
+
+	if endpoints != nil && endpoints.ComputeServiceEndpoint != "" {
+		opts = append(opts, option.WithEndpoint(endpoints.ComputeServiceEndpoint))
+	}
+
+	instanceTemplatesClient, err := computerest.NewInstanceTemplatesRESTClient(ctx, opts...)
+	if err != nil {
+		return nil, errors.Errorf("failed to create gcp instance templates rest client: %v", err)
+	}
+
+	return instanceTemplatesClient, nil
+}
+
 func newTagBindingsClient(ctx context.Context, credentialsRef *infrav1.ObjectReference, crClient client.Client, location string, endpoints *infrav1.ServiceEndpoints) (*resourcemanager.TagBindingsClient, error) {
 	opts, err := defaultClientOptions(ctx, credentialsRef, crClient)
 
--- a/cloud/scope/managedmachinepool.go
+++ b/cloud/scope/managedmachinepool.go
@@ -43,6 +43,7 @@ import (
 type ManagedMachinePoolScopeParams struct {
 	ManagedClusterClient        *container.ClusterManagerClient
 	InstanceGroupManagersClient *compute.InstanceGroupManagersClient
+	InstanceTemplatesClient     *compute.InstanceTemplatesClient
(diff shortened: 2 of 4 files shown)


salasberryfin (commented):
/ok-to-test

salasberryfin (commented):
/retest

salasberryfin (approved):
Thanks @jonathanrainer.

/lgtm

PR #1660, opened 2026-05-07
:sparkles::bug: Add GCPManagedCluster firewall reconciliation, fix firewall rule validation, and add docs

barbacbd:
Add detailed documentation for the firewall API feature that allows users to configure GCP VPC firewall rules through the GCPCluster resource.

The documentation covers:
- Default firewall rule management (Managed/Unmanaged policies)
- Custom firewall rule configuration with complete field reference
- Support for ALLOW/DENY rules, Ingress/Egress directions
- Five practical examples including SSH, web traffic, and multi-protocol rules
- Best practices, limitations, and troubleshooting guidance

 
**What type of PR is this?**

/kind feature
/kind documentation
/kind bug

- Adds firewall rule reconciliation for GCPManagedCluster (GKE), bringing feature parity with GCPCluster
- Fixes several firewall rule validation and creation bugs
- Adds comprehensive firewall rules documentation

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Feature
- Firewall rule reconciliation is now supported for GCPManagedCluster (GKE). Custom firewall rules defined in spec.network.firewall.firewallRules are created during provisioning and deleted on teardown. Default CAPG rules (health-check, cluster-internal) are skipped for GKE since they target CAPI-specific instance tags that do not exist on GKE nodes.
- Deterministic reconciler ordering for GCPManagedCluster — The controller now enforces explicit ordering (networks → firewalls → subnets on create; subnets → firewalls → networks on delete) instead of random map iteration order.

Bug Fixes
- Egress firewall rules now correctly populate destinationRanges when sent to the GCP API.
- Firewall rule priority validation now permits the full GCP range of 0–65535. Priority is now a pointer (*int32) so that an explicit 0 is not silently overridden by the default of 1000.
- CEL validation now rejects direction-specific field mismatches (sourceRanges/sourceTags on Egress, destinationRanges on Ingress) at admission time.
- CEL validation now rejects rules that set both allowed and denied, and requires at least one.
- The name field now allows empty values (auto-generated) while validating RFC 1035 format when provided. Auto-generated names include the rule index to prevent collisions.

Documentation
- Added comprehensive firewall rules documentation covering configuration, lifecycle, field reference, examples, and troubleshooting.

API Change
- Name field is now required (API change) — The current note says "allows empty values (auto-generated)" but the diff shows the field changed from Optional to Required with MinLength=1. The auto-generation logic (<cluster>-<direction>) was removed. This is a breaking API change that needs to be called out.
```


damdo:
/assign @barbacbd 

Could you rebase please and drop the coauthor from the commit thanks!

damdo:
@barbacbd I feel like I know little about this, maybe someone like @tthvo or the team there has more GCP networking domain knowledge and can verify correctness? TY

tthvo:
Also cc @rochacbruno for reviews :point_up::pray:

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
+# Firewall Rules
+
+Cluster API Provider GCP allows you to configure GCP VPC firewall rules for your clusters through the `GCPCluster` resource. This feature provides fine-grained control over network access to your cluster infrastructure.
```
tthvo: What about `GCPManagedCluster` (GKE)? I think firewall spec is also supported there, right?

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+
+- **Default rule management**: Control whether the provider creates default firewall rules
+- **Custom firewall rules**: Define additional firewall rules to meet your specific security requirements
+
+## Default Firewall Rules
+
+By default, the provider creates two firewall rules to enable cluster functionality:
+
+1. **Health Check Rule** (`allow-<cluster-name>-healthchecks`):
+   - Allows TCP traffic on port 6443 from GCP health check IP ranges
```
tthvo:
```suggestion
   - Allows TCP ingress traffic on port 6443 from GCP health check IP ranges
```

nit: let's explicitly state that the rule is `INGRESS`.

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+   - Source ranges: `35.191.0.0/16`, `130.211.0.0/22`
+   - Target: control-plane nodes
+
+2. **Cluster Internal Rule** (`allow-<cluster-name>-cluster`):
+   - Allows all traffic between cluster nodes
```
tthvo:
```suggestion
   - Allows all ingress traffic between cluster nodes
```


nit: let's explicitly state that the rule is `INGRESS`.

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+  name: my-cluster
+spec:
+  network:
+    firewall:
+      defaultRulesManagement: "Managed"  # or "Unmanaged"
+```
+
+**Values:**
+- `Managed` (default): The controller creates and manages default firewall rules
+- `Unmanaged`: The controller does not create or modify default firewall rules
```
tthvo:
```suggestion
- `Managed` (default): CAPG creates and manages default firewall rules
- `Unmanaged`: CAPG does not create or manage default firewall rules
```

nit: let's use `CAPG` and similar verb between the 2 modes for consistency?

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+
+**Important Notes:**
+- Changing from `Managed` to `Unmanaged` after rules are created will not delete existing rules
```
tthvo:
```suggestion
- CAPG allows switching from `Managed` to `Unmanaged` and vice versa. However, changing from `Managed` to `Unmanaged` after rules are created will not delete existing rules. Switching from `Unmanaged` to `Managed` will cause CAPG to create the default rules if they don’t already exist
```

Let's state that "switching" is allowed before mentioning the caveats?

Additionally, what happens if changing from `Unmanaged` to `Managed`? CAPG will create the default rules if not exist, right?

docs/book/src/topics/firewall-rules.md
tthvo:
```suggestion
Cluster API Provider GCP (CAPG) allows you to configure GCP VPC firewall rules for your clusters through the `GCPCluster` resource. This feature provides fine-grained control over network access to your cluster infrastructure.
```

nit: Let's add this so we can use the word `CAPG` later on :D

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+- `defaultRulesManagement` has no effect when using a shared VPC (HostProject)
```
tthvo: 💡 The user may wonder what CAPG behaviours will be when using a shared VPC. How about calling out that:

>  When using a shared VPC (`HostProject`), the controller will not create, modify, or delete any firewall rules (default or custom).

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+### FirewallRule Fields
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `name` | string | Optional | Rule name (1-63 chars, must match `[a-z]([-a-z0-9]*[a-z0-9])?`). If not prefixed with cluster name, it will be prepended automatically. |
+| `description` | string | Optional | Description of the rule (max 2000 chars). Defaults to "Created by Cluster API GCP Provider". |
+| `direction` | string | Optional | Traffic direction: `Ingress` (default) or `Egress`. |
+| `priority` | integer | Optional | Rule priority (1-65535). Lower values = higher priority. Default: 1000. |
+| `allowed` | []FirewallDescriptor | Optional | List of ALLOW rules (max 1024). |
+| `denied` | []FirewallDescriptor | Optional | List of DENY rules (max 1024). DENY rules take precedence over ALLOW rules with equal priority. |
```
tthvo:
```suggestion
| `denied` | []FirewallDescriptor | Optional | List of DENY rules (max 1024). DENY rules take precedence over ALLOW rules if they have equal priority. |
```

nit: just a little clearer :D

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+| `sourceRanges` | []string | Optional | Source IP ranges in CIDR format (max 1024). Supports IPv4 and IPv6. |
+| `sourceTags` | []string | Optional | Source instance tags (max 30, 1-63 chars each). Only applies to traffic between instances in the same VPC. |
+| `destinationRanges` | []string | Optional | Destination IP ranges in CIDR format (max 1024). Only valid for Egress rules. |
```
tthvo, line 100: I notice that the `destinationRanges` is defined, but unused by CAPG.  Is this a bug (perhaps accidentially missed)?

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/23049b61863d1608729391f70f5a2dfb4257472b/cloud/scope/firewall.go#L96-L108

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `name` | string | Optional | Rule name (1-63 chars, must match `[a-z]([-a-z0-9]*[a-z0-9])?`). If not prefixed with cluster name, it will be prepended automatically. |
+| `description` | string | Optional | Description of the rule (max 2000 chars). Defaults to "Created by Cluster API GCP Provider". |
+| `direction` | string | Optional | Traffic direction: `Ingress` (default) or `Egress`. |
+| `priority` | integer | Optional | Rule priority (1-65535). Lower values = higher priority. Default: 1000. |
+| `allowed` | []FirewallDescriptor | Optional | List of ALLOW rules (max 1024). |
+| `denied` | []FirewallDescriptor | Optional | List of DENY rules (max 1024). DENY rules take precedence over ALLOW rules with equal priority. |
+| `sourceRanges` | []string | Optional | Source IP ranges in CIDR format (max 1024). Supports IPv4 and IPv6. |
+| `sourceTags` | []string | Optional | Source instance tags (max 30, 1-63 chars each). Only applies to traffic between instances in the same VPC. |
```
tthvo:
```suggestion
| `sourceTags` | []string | Optional | Source instance tags (max 30, 1-63 chars each). Only applies to traffic between instances in the same VPC. Not valid for Egress rules. |
```

This field is not valid for Egress, right?

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/23049b61863d1608729391f70f5a2dfb4257472b/api/v1beta1/types.go#L188-L189

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+          direction: "Egress"
+          priority: 1000
+          allowed:
+            - IPProtocol: "TCP"
+              ports:
+                - "443"
+          destinationRanges:
+            - "198.51.100.0/24"
+          sourceTags:
+            - "api-client"
```
tthvo: Egress rule cannot use `sourceTags` right?

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+          sourceTags:
+            - "app-tier"
+          targetTags:
+            - "database-tier"
+```
+
+## Best Practices
+
+1. **Use Descriptive Names**: Choose meaningful names that describe the rule's purpose
+2. **Set Appropriate Priorities**: Use lower values for more critical security rules
```
tthvo:
```suggestion
2. **Set Appropriate Priorities**: Use lower values (higher priority) for more critical security rules
```

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,331 @@
...
+4. **Use Tags Strategically**: Leverage instance tags for flexible rule targeting
+5. **Document Rules**: Always include a description explaining the rule's purpose
+6. **Test Before Production**: Verify firewall rules in a development environment first
+7. **Avoid Priority 65535**: GCP reserves this priority for implied rules
+8. **Consider DENY Rules**: Use DENY rules for explicit blocking with higher priority
+
+## Limitations
+
+- Maximum 50 custom firewall rules per cluster
+- Maximum 1024 allowed/denied descriptors per rule
```
tthvo: nit: we probably can remove this section since they are mostly mention in the table above :D

cloud/scope/firewall.go
```diff
@@ -94,17 +94,18 @@ func createFirewallRules(clusterName, networkLink string, policy infrav1.RulesMa
...
+			Name:              name,
+			Description:       description,
+			Network:           networkLink,
+			Allowed:           allowed,
+			Denied:            denied,
+			Direction:         direction,
+			Priority:          int64(rule.Priority),
+			Disabled:          false,
+			SourceRanges:      rule.SourceRanges,
+			DestinationRanges: rule.DestinationRanges,
```
tthvo, line 113: We probably should split bug fix changes into a separate commit with message `bug: fix...` instead in the same one with docs change  👀

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+### FirewallRule Fields
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `name` | string | Optional | Rule name (1-63 chars, must match `[a-z]([-a-z0-9]*[a-z0-9])?`). If not prefixed with cluster name, it will be prepended automatically. |
+| `description` | string | Optional | Description of the rule (max 2000 chars). Defaults to "Created by Cluster API GCP Provider". |
+| `direction` | string | Optional | Traffic direction: `Ingress` (default) or `Egress`. |
+| `priority` | integer | Optional | Rule priority (1-65535). Lower values = higher priority. Default: 1000. |
```
tthvo: This `1-65535` is the restriction that comes from `GCPCluster` CEL.

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/95b916103f0b609a8556b2f3345325fbaa04771c/api/v1beta1/types.go#L218-L221


However, GCP allows `Integer from 0 to 65535` according https://docs.cloud.google.com/firewall/docs/firewalls#priority_order_for_firewall_rules.

Is this a CEL bug 👀?

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,348 @@
...
+| `allowed` | []FirewallDescriptor | Optional | List of ALLOW rules (max 1024). |
+| `denied` | []FirewallDescriptor | Optional | List of DENY rules (max 1024). DENY rules take precedence over ALLOW rules if they have equal priority. |
+| `sourceRanges` | []string | Optional | Source IP ranges in CIDR format (max 1024). Supports IPv4 and IPv6. |
```
tthvo: `sourceRanges` is also _not_ applicable to `Egress` firewall rule, right?

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,348 @@
...
+| `sourceTags` | []string | Optional | Source instance tags (max 30, 1-63 chars each). Only applies to traffic between instances in the same VPC. Not valid for Egress rules. |
+| `destinationRanges` | []string | Optional | Destination IP ranges in CIDR format (max 1024). Only valid for Egress rules. |
```
tthvo: These 3 fields are only applicable to either Ingress or Egress direction (i.e. not both). AFAICT, the GCP API will return `400` error if used in unsupported direction.


How about adding some CEL or webhook validation to block such case early? This way, in `GCPCluster` object won't get reconciled and avoid making invalid API calls.

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,348 @@
...
+            - "172.16.0.0/12"
+          targetTags:
+            - "web-servers"
+```
+
+### FirewallRule Fields
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `name` | string | Optional | Rule name (1-63 chars, must match `[a-z]([-a-z0-9]*[a-z0-9])?`). If not prefixed with cluster name, it will be prepended automatically. |
```
tthvo, line 92: The doc defines name restrictions so but API doesn't enforce it. Should we?

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/95b916103f0b609a8556b2f3345325fbaa04771c/api/v1beta1/types.go#L207-L208

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,348 @@
...
+| `description` | string | Optional | Description of the rule (max 2000 chars). Defaults to "Created by Cluster API GCP Provider". |
+| `direction` | string | Optional | Traffic direction: `Ingress` (default) or `Egress`. |
+| `priority` | integer | Optional | Rule priority (1-65535). Lower values = higher priority. Default: 1000. |
+| `allowed` | []FirewallDescriptor | Optional | List of ALLOW rules (max 1024). |
+| `denied` | []FirewallDescriptor | Optional | List of DENY rules (max 1024). DENY rules take precedence over ALLOW rules if they have equal priority. |
```
tthvo: Field `allowed` and `denied` cannot be specified in a single rule, right? That is: a rule must either allow or deny (can't do both). We should mention it...

https://docs.cloud.google.com/firewall/docs/firewalls#action_of_the_rule

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,348 @@
+# Firewall Rules
+
+Cluster API Provider GCP (CAPG) allows you to configure GCP VPC firewall rules for your clusters through the `GCPCluster` and `GCPManagedCluster` (GKE) resources. This feature provides fine-grained control over network access to your cluster infrastructure.
+
+## Overview
+
+Firewall rules are configured through the `network.firewall` field in the `GCPCluster` or `GCPManagedCluster` spec. Both resource types share the same network specification, so firewall configuration works identically for self-managed and GKE clusters. The firewall configuration supports:
```
tthvo, line 7: 🤦 oh whoops, the firewall field is defined on `GCPManagedCluster` (GKE), but never really got reconciled/created?

I'm looking at these registered reconciler. IIUC, firewall specs have no effect.

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/95b916103f0b609a8556b2f3345325fbaa04771c/exp/controllers/gcpmanagedcluster_controller.go#L198-L201

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,348 @@
...
+
+- **Default rule management**: Control whether the provider creates default firewall rules
+- **Custom firewall rules**: Define additional firewall rules to meet your specific security requirements
```
tthvo, line 10: We should mention that:
- CAPG doesn't actively update the firewall rules to match their current (if updated) specs after creation
- CAPG doesn't remove firewall rules if they are not in the list anymore.

AFAIK, CAPG only creates rules and deletes the rules only when cluster is deleted, right?

exp/controllers/gcpmanagedcluster_controller.go
```diff
@@ -196,8 +197,9 @@ func (r *GCPManagedClusterReconciler) reconcile(ctx context.Context, clusterScop
...
 
 	reconcilers := map[string]cloud.Reconciler{
-		"networks": networks.New(clusterScope),
-		"subnets":  subnets.New(clusterScope),
+		"networks":  networks.New(clusterScope),
+		"subnets":   subnets.New(clusterScope),
+		"firewalls": firewalls.New(clusterScope),
 	}
 
 	for name, r := range reconcilers {
```
tthvo: Map ordering is non-deterministic, which would:
  - On create: firewalls may be created before the network exists.
  - On delete: the network may be deleted before firewalls.

These will result in API error, right? We should convert the map into a slice with a comment explaining ordering dependencies like `GCPCluster` controller.

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/95b916103f0b609a8556b2f3345325fbaa04771c/controllers/gcpcluster_controller.go#L201-L207

api/v1beta1/types.go
```diff
@@ -155,6 +155,10 @@ type FirewallDescriptor struct {
 }
 
 // FirewallRule describes a GCP firewall rule.
+// +kubebuilder:validation:XValidation:rule="self.direction != 'Egress' || !has(self.sourceRanges) || size(self.sourceRanges) == 0",message="sourceRanges cannot be set for Egress rules"
+// +kubebuilder:validation:XValidation:rule="self.direction != 'Egress' || !has(self.sourceTags) || size(self.sourceTags) == 0",message="sourceTags cannot be set for Egress rules"
+// +kubebuilder:validation:XValidation:rule="self.direction != 'Ingress' || !has(self.destinationRanges) || size(self.destinationRanges) == 0",message="destinationRanges cannot be set for Ingress rules"
+// +kubebuilder:validation:XValidation:rule="!(has(self.allowed) && size(self.allowed) > 0 && has(self.denied) && size(self.denied) > 0)",message="allowed and denied cannot both be set in the same rule"
```
tthvo, line 161: 💡 What about  `at least one of denied or allowed must be set`? A rule must have either, right?

api/v1beta1/types.go
```diff
@@ -205,6 +208,9 @@ type FirewallRule struct {
 	// If the firewall does not begin with the cluster name, then the cluster name
 	// will be prepended during the creation of the firewall rule.
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[a-z]([-a-z0-9]*[a-z0-9])?$`
 	Name string `json:"name,omitempty"`
```
tthvo, line 214: Sorry, on a second look, the `name` is optional (length 0 is allowed), but now has 1 min length constraint and regex check. This is conflicting, right?

```yaml
// +kubebuilder:validation:Optional
// +kubebuilder:validation:MaxLength=63
// +kubebuilder:validation:XValidation:rule="self == '' || self.matches('^[a-z]([-a-z0-9]*[a-z0-9])?$')",message="name must match RFC1035 format when specified
```

This allows empty (auto-generated name) and non-empty (check with regex).

tthvo: > This allows empty (auto-generated name) 

On that note, the rule name auto-generation logic may create conflicts if 2 rules have empty names and same direction. May we should pad the generated name with random chars.

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/95b916103f0b609a8556b2f3345325fbaa04771c/cloud/scope/firewall.go#L81

api/v1beta1/types.go
```diff
@@ -216,7 +222,7 @@ type FirewallRule struct {
 	// use a priority number less than `65535`.
 	// +kubebuilder:validation:Optional
 	// +default:value=1000
-	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=65535
 	Priority int `json:"priority,omitempty"`
```
tthvo: Sorry, I realized user-provided `priority: 0` won't take effect because it's treated as "zero" value. This means the API server will default it `1000` and override the user value, right?

I think it should be a 2-part fix:

- In `api/v1beta1/types.go`: make ``` Priority *int32 `json:"priority,omitempty"` ``` a pointer.
- In `cloud/scope/firewall.go`: add `ForceSendFields: []string{"Priority"}` because the sdk also treats `0` as "zero" value. I think 🤔.
  https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/95b916103f0b609a8556b2f3345325fbaa04771c/cloud/scope/firewall.go#L103

cloud/scope/firewall.go
```diff
@@ -78,7 +78,7 @@ func createFirewallRules(clusterName, networkLink string, policy infrav1.RulesMa
 		}
 
 		direction := strings.ToUpper(string(rule.Direction))
-		name := fmt.Sprintf("%s-%s", clusterName, strings.ToLower(direction))
+		name := fmt.Sprintf("%s-%s-%d", clusterName, strings.ToLower(direction), i)
```
tthvo, line 81: This would create duplicate firewall rules if the users re-order the firewall rule list (mutable I believe), right? How about random chars with cap at 63 max length?

**Note**: we need to make sure the padding is deterministically generated (e.g. from rule spec). Otherwise, CAPG will keep creating new rule in every reconciling cycle :sweat_smile: (basically same issue with index suffix). Or maybe some other better ways :eyes:

exp/controllers/gcpmanagedcluster_controller.go
```diff
@@ -196,13 +203,18 @@ func (r *GCPManagedClusterReconciler) reconcile(ctx context.Context, clusterScop
 	clusterScope.SetFailureDomains(failureDomains)
 
 	reconcilers := map[string]cloud.Reconciler{
-		"networks": networks.New(clusterScope),
-		"subnets":  subnets.New(clusterScope),
+		reconcilerNetworks:  networks.New(clusterScope),
+		reconcilerSubnets:   subnets.New(clusterScope),
+		reconcilerFirewalls: firewalls.New(clusterScope),
```
tthvo, line 208: Sorry for the back and forth... 

I think this is a potential breaking change. Specifically, the 2 "default" ingress rule will now get created:

1. Health Check Rule (`allow-<cluster-name>-healthchecks`)
2. Cluster Internal Rule (`allow-<cluster-name>-cluster`):

Previously, they are not. Is that OK? Or should the the default rule management mode be `"Unmanaged"` as default for GKE?

cloud/scope/firewall.go
tthvo: We discussed offline, but there are 2 other options:
- Make the firewall list spec "immutable" since CAPG doesn't fully manage the lifecycle of fire walls (only create and delete when cluster is created/deleted)
- Make the firewall name spec "required", which changes the API.

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,357 @@
+# Firewall Rules
+
+Cluster API Provider GCP (CAPG) allows you to configure GCP VPC firewall rules for your clusters through the `GCPCluster` and `GCPManagedCluster` (GKE) resources. This feature provides fine-grained control over network access to your cluster infrastructure.
+
+## Overview
+
+Firewall rules are configured through the `network.firewall` field in the `GCPCluster` or `GCPManagedCluster` spec. Both resource types share the same network specification, so firewall configuration works identically for self-managed and GKE clusters. The firewall configuration supports:
```
rochacbruno: I think this contradicts the change made on cloud/scope/managedcluster.go that is now hardcoding infrav1.RulesmanagementUnmanaged and ignores DefaultRuesManagement for GCPManagedCLuster.

A user following the docs will set defaultRulesManagement: "Managed" on a GKE cluster and silently get nothing. The docs should call out that defaultRulesManagement is ignored for GCPManagedCluster and remove/annotate it in Example 6

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,357 @@
...
+kind: GCPManagedCluster
+metadata:
+  name: my-gke-cluster
+spec:
+  project: my-gcp-project
+  region: us-central1
+  network:
+    name: gke-network
+    firewall:
+      defaultRulesManagement: "Managed"
```
rochacbruno: I think this will be silently ignored right? ` infrav1.RulesManagementUnmanaged` will be the hardcoded default.

api/v1beta1/types.go
```diff
@@ -205,6 +209,8 @@ type FirewallRule struct {
 	// If the firewall does not begin with the cluster name, then the cluster name
 	// will be prepended during the creation of the firewall rule.
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:XValidation:rule="self == '' || self.matches('^[a-z]([-a-z0-9]*[a-z0-9])?$')",message="name must match RFC1035 format when specified"
 	Name string `json:"name,omitempty"`
```
rochacbruno: Isn't this allowing the effective firewall rule name to exceed 63 chars?

the supplied name is validated for 63 chars, but the final name on cloud/scope/firewall.go will prepend `clusterName-` if the name is not prefixed making it larger than 63 chars.

docs/book/src/topics/firewall-rules.md
barbacbd: The managed cluster will be ignored but the GCP Cluster will not.

api/v1beta1/types.go
barbacbd: This is handled in the code not here. We don't know how long the cluster name will be.

salasberryfin (commented):
With the limited context I have on this topic I can say that this generally LGTM. I'll hold it to give time for others to review.

/lgtm
/hold

tthvo (commented):
LGTM 👍 

I have little experience in GKE, so leaving it to other reviewers in that part 👀

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,359 @@
...
+- **Custom firewall rules**: Define additional firewall rules to meet your specific security requirements
+
+### Lifecycle Behavior
+
+CAPG creates firewall rules during cluster provisioning and deletes them when the cluster is deleted. However, CAPG does **not**:
+
+- **Update** existing firewall rules if the spec is modified after initial creation
+- **Remove** individual firewall rules if they are removed from the spec
+
+To modify or remove a firewall rule after it has been created, you must make the change directly in GCP (e.g., via the Console, `gcloud`, or Terraform).
```
jonathanrainer, line 19: Is there a specific reason for picking this behaviour? If the change can be made directly in GCP then can we not do it via the API? Don't want to block on this (so could come as a future PR) but from what I've seen in the code the general approach is to not make things immutable if the API supports it. 

Further do we have immutability checks on these fields in that case?

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,359 @@
...
+      defaultRulesManagement: "Managed"  # or "Unmanaged"
+```
+
+**Values:**
+- `Managed` (default): CAPG creates and manages default firewall rules
+- `Unmanaged`: CAPG does not create or manage default firewall rules
+
+**Important Notes:**
+- CAPG allows switching from `Managed` to `Unmanaged` and vice versa. However, changing from `Managed` to `Unmanaged` after rules are created will not delete existing rules. Switching from `Unmanaged` to `Managed` will cause CAPG to create the default rules if they don't already exist
+- **GCPManagedCluster (GKE):** The `defaultRulesManagement` field is ignored. CAPG does not create default firewall rules for GKE clusters because the default rules target CAPI-specific instance tags (e.g. `<cluster>-control-plane`, `<cluster>-node`) that do not exist on GKE nodes. Only custom firewall rules defined in `firewallRules` are reconciled
```
jonathanrainer, line 55: Of that's the case then why include the field at all on `GCPManagedCluster` CRs?

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,359 @@
...
+## Best Practices
+
+1. **Use Descriptive Names**: Choose meaningful names that describe the rule's purpose
+2. **Set Appropriate Priorities**: Use lower values (higher priority) for more critical security rules
+3. **Minimize Source Ranges**: Restrict access to only necessary IP ranges
+4. **Use Tags Strategically**: Leverage instance tags for flexible rule targeting
+5. **Document Rules**: Always include a description explaining the rule's purpose
+6. **Test Before Production**: Verify firewall rules in a development environment first
+7. **Avoid Priority 65535**: GCP reserves this priority for implied rules
+8. **Consider DENY Rules**: Use DENY rules for explicit blocking with higher priority
```
jonathanrainer: Is it worth having these at the top before the examples so they don't get buried? It feels like a key piece of information here

docs/book/src/topics/firewall-rules.md
```diff
@@ -0,0 +1,359 @@
...
+## Troubleshooting
+
+### Rules Not Applied
+
+If firewall rules are not being applied:
+
+1. Check that `defaultRulesManagement` is set to `Managed` if you expect default rules
+2. Verify you're not using a shared VPC (HostProject), which disables custom firewall rules
+3. Ensure rule names are unique and follow GCP naming conventions
+4. Check GCP Cloud Console for any conflicting VPC firewall rules
```
jonathanrainer: Do we surface any of this in the controller logs? It feels like we should be able to give users an indication of what's failing there as well as telling them to look through these factors

exp/controllers/gcpmanagedcluster_controller.go
jonathanrainer: Is there a reason not to pull this out as a var at the top of the file and turn it into a slice of anonymous structs then you could have the reconciler and the name in one place? Then L215 just becomes an iteration over the slice which can be done in order and L273 can become a slices.Reverse call and it stops having to keep declaring the maps and indexing into them so you can keep the ordering and the name.

jonathanrainer: Something like 

```golang 
var reconcilers = []struct {
	reconciler cloud.Reconciler
	name string
}{
	{networks.New(clusterScope), "networks"},
	...
}
```

jonathanrainer (commented):
Generally looks ok but have some questions about the approach that are framed around the documentation pieces specifically and a small suggestion to clean up some of the code in the `gcpmanagedcluster_controller`

docs/book/src/topics/firewall-rules.md
barbacbd: It isn't declared on GCPManagedCluster — it comes in through the shared
infrav1.NetworkSpec, which embeds FirewallSpec. GCPManagedCluster.Spec.Nthat same type, so the field is inherited rather than deliberately exposed. (Same for GCPManagedClusterTemplate.)                                                   
The reason it's ignored rather than honored is deliberate, though, and there are
two parts to it:                                                                 
1. The default rules target <cluster>-control-plane and <cluster>-node instance
   tags. GKE nodes don't carry those — their tags are auto-generated            (gke-<cluster>-<hash>-node) and aren't known to CAPG when the spec is built. Creating the rules anyway would just add two dead rules to the VPC.
2. This PR wires the firewall reconciler into the GCPManagedCluster controller for the first time; it has never run on that path. Since defaultRulesManagement defaults to Managed, honoring it would silently create two new firewall rules in every existing GKE user's VPC on upgrade.

So the field can't be honored without breaking existing clusters, which is why it's explicitly documented as ignored and ManagedClusterScope hardcodes Unmanaged for the default rules while still reconciling custom firewallRules.

I've also added an Info log on the managed reconcile path so this isn't only
discoverable in the docs (per your other comment about surfacing things in the
controller logs).

If you'd rather the field be impossible to set on GKE rather than documented-as-ignored, the cheapest version is making DefaultRulesManagement a pointer and dropping the kubebuilder default, so the webhook can warn only when it's explicitly set. Splitting the type so it doesn't appear at all would need a separate network spec for the managed CRDs, since Firewall is nested inside the shared NetworkSpec. Happy to do either — just didn't want to reshape a shared API type without agreement.

docs/book/src/topics/firewall-rules.md
barbacbd: *Edit: the original version of this comment was mangled when I posted it — several spans of text were dropped mid-sentence, which made the middle section unreadable. I've restored it below from my notes; the substance is as originally intended, though the wording won't be character-for-character. One claim in it turned out to be wrong — see the follow-up comment rather than this edit for that correction.*

---

> Is there a specific reason for picking this behaviour?

No principled one — it's a gap rather than a decision, and you're right that it doesn't match the general approach. The GCP API supports it, and `firewallsInterface` already declares `Update`; it's just never called. There's precedent in-repo too, since the loadbalancers service updates backend services in place.

> Further do we have immutability checks on these fields in that case?

No. The CEL rules on `FirewallSpec`/`FirewallRule` are all field-shape validations (direction-specific fields, allowed/denied exclusivity, CIDR validity) — none use `oldSelf`. So the fields are fully mutable today, the API server accepts the edit, and the controller silently drops it. Agreed that's the worst combination, and I'd rather fix the behaviour than add immutability markers.

One related gap worth flagging: `Delete()` iterates the current spec, so a rule removed from `firewallRules` isn't cleaned up on update or at cluster deletion — it's orphaned in the VPC permanently.

**Proposed follow-up, split by risk:**

Update in place is straighter than `Patch`, since `createFirewallRules` builds the complete desired `compute.Firewall` on every reconcile — so a full `Update` correctly clears fields the user removed, where a merge patch would leave stale values behind. `name` and `network` can't be changed on an existing rule, but neither participates in the drift comparison, so no unsatisfiable updates get generated.

Pruning is the part I'd like to defer. Ownership detection is genuinely awkward here: firewall rules have no `labels` field, so the usual tagging approach isn't available. `Description` exists but is documented as being provided at create time, so I didn't think it could be backfilled onto existing rules. That leaves name prefixes, which don't cleanly cover the two rule families — custom rules are `<cluster>-<name>` while the defaults are `allow-<cluster>-healthchecks` / `allow-<cluster>-cluster` — and a user's own rule named `<cluster>-something` in the same VPC would get swept up.

I'd suggest:

1. **Delete path now** — at teardown, list rules filtered to the cluster's network, match the cluster-owned naming shapes, delete. The cluster is going away, the names embed the cluster name, and shared VPC is already short-circuited so a host project is never touched. Worst case is a marginal leak.
2. **Reconcile-time pruning separately** — deleting a live cluster's rule on a false positive is much worse than leaving an orphan, so I don't want to ship it on prefix matching alone. The option I can see is a machine-readable marker in `Description` on rules CAPG creates, pruning only rules that carry it (rules from older versions lack it and are never auto-pruned, which fails safe). That's a bit ugly since it's visible in the console — if you know a better ownership signal I'm happy to take it.

Renames would still orphan until teardown under (1), since the name is the resource key. I'd document that explicitly.

Does that split seem reasonable, and do you have a preference on the ownership marker?

jonathanrainer: I think the split makes sense, but on the ownership marker do we expect the cluster CR to be the source of truth? If we do then yes there's a risk when folks turn the feature on but otherwise we shouldn't be encouraging people to run rules that aren't on the cluster object anyway so the marker point ceases to be an issue if we insist the CR is the source of truth.

docs/book/src/topics/firewall-rules.md
jonathanrainer: Ah I didn't see the type was shared, as long as we're surfacing in the logs when there's an issue that's fine with me :)

docs/book/src/topics/firewall-rules.md
barbacbd: That framing is appealing, and I think you're right about the core of it — the CR should be authoritative for the rules CAPG owns, and we shouldn't be encouraging people to keep cluster rules off the cluster object. The part I'm still chewing on is whether that extends all the way to the VPC, and there are a couple of cases I keep getting stuck on. Very possibly I'm overweighting them, so tell me if so.

**First, a correction to my previous comment.** I said `Description` was INSERT-only and therefore couldn't be backfilled. That turns out to be wrong — `firewalls.patch` accepts it, and `gcloud compute firewall-rules update --description=...` documents "Set to an empty string to clear existing." So a marker can be applied to rules that already exist, which makes that option more viable than I made it sound. (`labels` really isn't available — `compute.Firewall` has no such field — so that part stands.)

**The two cases I keep running into:**

- **Pre-existing networks.** `createOrGetNetwork` only creates the VPC if it's absent, so pointing a `GCPCluster` at an existing network is supported and fairly ordinary. Shared VPC is short-circuited by `SkipFirewallRulesManagement`, but a pre-existing VPC *in the same project* isn't — so "prune everything not in the CR" would reach unrelated rules in it. As far as I can tell nothing in the CR distinguishes "CAPG created this network" from "this is the user's VPC", though I'd be glad to be wrong about that.
- **`DefaultRulesManagement: Unmanaged`.** Today it only suppresses the default rules; custom rules still reconcile. Under CR-as-source-of-truth, the field that means "I manage firewall rules myself" would end up deleting the rules the user hand-wrote. That's the risk you flagged, and it seems to land hardest on exactly the people who opted out.

**The good news is I don't think we have to choose between them**, because the repo already has the ownership convention this needs. `networks/reconcile.go` declines to delete a network whose `Description != infrav1.ClusterTagKey(clusterName)`, and does the same for the NAT router — so a description-based "did I create this?" check is an established pattern here rather than something new. It lines up with CAPA too, where the `sigs.k8s.io/cluster-api-provider-aws/cluster/<name>: owned` tag drives it and untagged resources are explicitly hands-off.

There may be an even nicer signal available: `status.network.firewallRules` already exists in the API (`api/v1beta1/types.go`, "a map from the name of the rule to its full reference") and is in all four CRDs — but nothing currently writes it. Populating it on insert would give a per-cluster record of what CAPG created, with no false positives on a user's own `<cluster>-*` rule and nothing user-visible in the console. The gap is rules created before we start populating it, which have no record — the `Description` marker covers those as a fallback, and per the correction above it can be backfilled. Either way it fails safe: unknown provenance means we don't prune.

So concretely, for the follow-up:

| | pre-existing network |
|---|---|
| create missing rules | unchanged — additive, and both defaults are target-tag scoped to `<cluster>-control-plane` / `<cluster>-node`, so they can't widen access to anything that isn't CAPG's |
| update drifted rules | unchanged — only touches rules matched by name from the desired set |
| prune rules not in spec | only rules CAPG can show it created |

One related question I ran into while thinking this through, in case it's worth settling at the same time: whether a pre-existing network ought to push users toward `Unmanaged` automatically. I lean against it — the default rules aren't really optional, since without `allow-<cluster>-healthchecks` the API server LB health check fails, and a custom-mode VPC starts with no rules and deny-all ingress. That would make BYO-network mean "bring your own network and reproduce two rules from CAPG's source." CAPA seems to take the same line: BYO VPC still gets provider-created security groups, with `securityGroupOverrides` as the opt-out. `DefaultRulesManagement: Unmanaged` is already our equivalent, so the choice stays available without being inferred. Keying it off network provenance would also mean the same CR behaves differently depending on GCP state at reconcile time, which feels like something to avoid — but I haven't thought about it as long as you've thought about this area, so I'm open on it.

None of this needs to block the PR from my side — happy to settle the pruning design in the follow-up if you'd rather keep this one to the reconciliation and docs changes.

jonathanrainer: Hmmm yes, I think there's a lot of nuance here and actually having these discussions with real code in front of us is a better idea. In which case, I'm happy to resolve this thread and approve but it would be good to have a fast-follow PR which implements the kind of reconciling behaviour described here, then we can continue the conversation there.

barbacbd: WDYT @damdo @salasberryfin

PR #1667, opened 2026-05-16
e2e: template cloud-controller-manager version as a variable

damdo:

The CCM image tag was hardcoded in the DaemonSet manifest (`test/e2e/data/ccm/gce-cloud-controller-manager.yaml`). This extracts it into a `CCM_VERSION` variable in `gcp-ci.yaml`, making version bumps a single-line config change instead of editing the manifest directly.

The manifest uses `${CCM_VERSION:-v34.2.0}` so it falls back to the current pinned version if the variable is unset.

**What type of PR is this?**
/kind cleanup

**Release note**:
```release-note
NONE
```


salasberryfin (commented):
/lgtm
/approve

PR #1668, opened 2026-05-16
ci: dynamically compute Ginkgo parallel nodes based on available CPUs

damdo:
**What type of PR is this?**
/kind cleanup

Replace the hardcoded `GINKGO_NODES=3` in `ci-e2e.sh` with Ginkgo's `--parallel` flag, which dynamically sets the number of parallel test nodes based on available CPUs. Remove the now-unused `GINKGO_NODES` variable from the Makefile.

Local `make test-e2e` runs remain single-node (Ginkgo's default), CI runs auto-detect from CPU count.

**Note:** this PR pairs with [kubernetes/test-infra#37046](https://github.com/kubernetes/test-infra/pull/37046), which bumps CPU requests for CAPG e2e jobs from 2 to 4 cores. That PR should merge first to avoid a temporary regression from 3→2 nodes.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,6 @@ test: $(SETUP_ENVTEST) ## Run unit and integration tests
 # Allow overriding the e2e configurations
 GINKGO_FOCUS ?= Workload cluster creation
 GINKGO_SKIP ?= API Version Upgrade
-GINKGO_NODES ?= 1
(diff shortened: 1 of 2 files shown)


salasberryfin (commented):
/lgtm
/approve

PR #1676, opened 2026-05-24
Make GCPCluster network name and subnets immutable

### LINKED ISSUE #1576: GCPCluster network and subnet fields are mutable but updates result in broken/inconsistent state
/kind bug

I performed an experiment to test the behavior of updating network infrastructure on a running cluster.
1. Created a valid GCPCluster with a defined network and subnets.
2. Waited for the infrastructure to be provisioned (VPC and Subnets created).
3. Updated the GCPCluster manifest, changing the spec.network.name and spec.network.subnets to new values (simulating a move to a new VPC).
4. Applied the change.

Observed Behavior: The API server accepted the update (the fields are mutable). The CAPG controller reconciliation loop then proceeded to create the new subnets defined in the updated spec. However, it did not migrate existing Control Plane or Worker machines to the new network, nor did it clean up the old network resources.

This resulted in a "split" state:
- The GCPCluster object references a new VPC.
- Existing VMs (Control Plane/Workers) remain stranded in the old VPC.
- New MachineDeployments (or rollouts) attempt to provision in the new VPC, potentially causing connectivity failures (e.g., if the new VPC is isolated/not peered).

I expected the API to reject the update to these fields, Since these are destructive changed for any running cluster.

Infrastructure fields that cannot be reconciled without breaking cluster continuity (like spec.network.name, spec.project, and spec.region) should be marked as Immutable to prevent users from accidentally entering this undefined state.

I'd like to hear the community's opinion regarding this behavior. I find it very concerning to work with clusters in such a way, especially in production environments...

**Environment:**
- CAPI version: v0.24.1
- CAPG version: v0.17
- Kubernetes Version: v1.33.5-gke.1308000 (GKE)

bobu-putheeckal:
Fixes #1576

This updates the GCPCluster validating webhook to reject updates to `spec.network.name` and `spec.network.subnets`. Changing those fields after infrastructure has been provisioned can leave existing machines on the old VPC/subnets while future reconciliation uses the new network configuration.

The PR also adds webhook unit coverage for both immutable fields.

Tests:
- `go test ./webhooks`
- `KUBEBUILDER_ASSETS="/Users/agentzero/Library/Application Support/io.kubebuilder.envtest/k8s/1.34.0-darwin-arm64" go test ./api/... ./webhooks/...`

```release-note
GCPCluster `spec.network.name` and `spec.network.subnets` are now immutable.
```

### DIFF
--- a/webhooks/gcpcluster_webhook.go
+++ b/webhooks/gcpcluster_webhook.go
@@ -103,6 +103,20 @@ func (*GCPCluster) ValidateUpdate(_ context.Context, oldObj, newObj runtime.Obje
 		)
 	}
 
+	if !reflect.DeepEqual(c.Spec.Network.Name, old.Spec.Network.Name) {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "network", "name"),
+				c.Spec.Network.Name, "field is immutable"),
+		)
+	}
+
+	if !reflect.DeepEqual(c.Spec.Network.Subnets, old.Spec.Network.Subnets) {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "network", "subnets"),
+				c.Spec.Network.Subnets, "field is immutable"),
+		)
+	}
+
 	if c.Spec.Network.Mtu < int64(1300) {
 		allErrs = append(allErrs,
 			field.Invalid(field.NewPath("spec", "Network", "Mtu"),
--- a/webhooks/gcpcluster_webhook_test.go
+++ b/webhooks/gcpcluster_webhook_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+	"k8s.io/utils/ptr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-gcp/api/v1beta1"
 )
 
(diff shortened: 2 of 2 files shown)


webhooks/gcpcluster_webhook.go
```diff
@@ -103,6 +103,20 @@ func (*GCPCluster) ValidateUpdate(_ context.Context, oldObj, newObj runtime.Obje
...
+		)
+	}
+
+	if !reflect.DeepEqual(c.Spec.Network.Subnets, old.Spec.Network.Subnets) {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "network", "subnets"),
+				c.Spec.Network.Subnets, "field is immutable"),
+		)
+	}
+
```
jonathanrainer, line 119: Do we want to just test for equality here? Are there cases where say adding extra subnets might be legitimate but wholesale changing them isn't? I appreciate the original ask here is immutable but I wonder if this could be relaxed slightly?

jonathanrainer (commented):
Generally looks good to me, just one question

jonathanrainer:
@bobu-putheeckal One other thing that's come up while I've been investigating something else and I thought I'd throw it in here. This is going to be equally as applicable to `GCPManagedCluster` as well, so we might as well extend the idea and do the check in there too. Appreciate the original PR only makes reference to `GCPCluster` but it seems silly to not extend the fix as we're already doing the work anyway

PR #1694, opened 2026-06-22
fix: allow GCPMachineTemplate network tag updates

### LINKED ISSUE #1693: GCPMachineTemplate webhook rejects changes to additionalNetworkTags
/kind bug

**Action**: Modified ClusterClass to add additionalNetworkTags and seeing admission webhook errors during reconciliation.

**Observed error from the CAPG controller logs:**
`TopologyReconciled: error reconciling the Cluster topology: failed to create patch helper for GCPMachineTemplate ns/cluster-name: server side apply dry-run failed for modified object: admission webhook "validation.gcpmachinetemplate.infrastructure.cluster.x-k8s.io" denied the request: GCPMachineTemplate.infrastructure.cluster.x-k8s.io "ns/cluster-name" is invalid: spec: Forbidden: cannot be modified
`

**Investigation:** I reviewed gcpmachinetemplate_webhook.go and found it deletes mutable fields from the top-level spec map (e.g. newGCPMachineTemplate["spec"]) rather than the nested spec.template.spec where additionalNetworkTags actually lives. Because of this the webhook does not strip/ignore the mutable field and the validation rejects server-side apply.

**Code reference:** the problematic logic is here: https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/0fb8ab30/webhooks/gcpmachinetemplate_webhook.go#L70-L83

The CAPG controller should be able to apply the updated generated GCPMachineTemplate containing changed additionalNetworkTags (or the webhook should ignore that mutable field), allowing topology reconciliation to succeed.

**Root cause:**
The code extracts spec from the unstructured map, but GCPMachineTemplate nests the actual spec two levels deep:

Level 1: spec.template (a GCPMachineTemplateResource)
Level 2: spec.template.spec (the actual GCPMachineSpec)
The current code deletes from the wrong level:
```
newGCPMachineTemplateSpec := newGCPMachineTemplate["spec"].(map[string]interface{})
// This only contains: {template: {...}}
delete(newGCPMachineTemplateSpec, "providerID")  // Does nothing - providerID isn't here
**Resolution:**
Navigate to the correct nested map (spec.template.spec) before deleting the mutable fields, so they are properly excluded from the immutability comparison:
```
**Resolution:**
The original code extracts only spec, which contains template. The mutable fields (providerID, additionalLabels, additionalNetworkTags) live inside spec.template.spec.
By adding ["template"]["spec"] to the extraction path, the code reaches the correct nested level.

Also, make image and imageFamily mutable fields.

**Environment:**
- Cluster-api version: v1.12.8
-  CAPG version: v1.12.0

nanookclaw:
/kind bug

`GCPMachineTemplate` update validation intended to allow changes to `providerID`, `additionalLabels`, and `additionalNetworkTags`, but it removed those keys from the top-level `spec` map before comparing objects.

Those fields live under `spec.template.spec` for machine templates, so changes to `additionalNetworkTags` still made the webhook reject the update as an immutable `spec` change. This removes the allowed mutable fields at the nested template-spec path before comparison and keeps unrelated template changes immutable.

Fixes #1693

Added focused unit coverage for allowing `additionalNetworkTags` changes while still rejecting an `instanceType` change.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
fix(webhooks): allow updates to GCPMachineTemplate additionalNetworkTags
```

### DIFF
--- a/webhooks/gcpmachinetemplate_webhook.go
+++ b/webhooks/gcpmachinetemplate_webhook.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	infrav1 "sigs.k8s.io/cluster-api-provider-gcp/api/v1beta1"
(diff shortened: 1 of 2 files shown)


nanookclaw:
Addressed the review comment by adding the requested comment above the mutable-fields loop. New head: `c0d42af1`. Verification: `git diff --check` (comment-only change).

salasberryfin (commented):
Thanks @nanookclaw.

/ok-to-test

damdo (commented):
/approve
/lgtm

nickinkaty:
@damdo Is it possible to review which fields are mutable? imageFamily and the image spec should be mutable fields? I think these should be added to the list.

damdo:
> @damdo Is it possible to review which fields are mutable? imageFamily and the image spec should be mutable fields? I think these should be added to the list.

@nickinkaty could you please open a GitHub issue for this? Thanks!

PR #1713, opened 2026-07-02
fix: fetch GCPManagedMachinePools in order

anmazzotti:
**What type of PR is this?**

/kind bug

This PR fixes a bug where GCPManagedMachinePools and MachinePools are sorted in the wrong order.
This creates wrong comparisons that result in `GCPManagedControlPlane` reconciliation errors like:
```
status:
  conditions:
  - lastTransitionTime: 2026-06-30T01:47:56Z
    message: "creating cluster: preflight checks on machine pools before cluster create: expect machinepool infraref (cluster-gke-9tzw7a-system-pool-dlcjd) to match managed machine pool name (cluster-gke-9tzw7a-worker-pool-94j8l)"
    reason: GKEControlPlaneReconciliationFailed
    severity: Error
    status: "False"
    type: Ready
```

The code is now ensuring the GCPManagedMachinePool list is in the same order as the MachinePool list.

Also note that there was a check to ensure that the number of GCPManagedMachinePools and MachinePools are the same. I couldn't find a reason for this check to exist, given that GCPManagedMachinePools are owned by MachinePools, so the number should always be implicitly the same. Ownership should also guarantee GCPManagedMachinePool deletion once the owner MachinePool is deleted.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
fix: GCPManagedControlPlane reconciliation error when multiple GCPManagedMachinePools are in use
```

### DIFF
--- a/cloud/scope/managedcontrolplane.go
+++ b/cloud/scope/managedcontrolplane.go
@@ -28,6 +28,7 @@ import (
 	credentials "cloud.google.com/go/iam/credentials/apiv1"
 	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
 	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-gcp/exp/api/v1beta1"
 	clusterv1beta1 "sigs.k8s.io/cluster-api/api/core/v1beta1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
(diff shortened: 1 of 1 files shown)


salasberryfin (commented):
/ok-to-test

salasberryfin (commented):
Thanks @anmazzotti.

I'm thinking that eventual consistency may be the reason this length check exists. Could there be a (transient) scenario where the number of `GCPManagedMachinePools` has not yet caught up with the number of `MachinePools`? This could cause a problem when ordering, couldn't it?

_nit_: can you add a release note to your PR description?

salasberryfin (approved):
/lgtm
/approve

salasberryfin (approved):
/lgtm

PR #1715, opened 2026-07-07
Add MonitoringConfig to GCPManagedControlPlane to disable managed Prometheus

jonathanrainer:
**What type of PR is this?**
/kind feature

GKE enables Google Cloud Managed Service for Prometheus by default on new clusters, and there was no way to disable it through CAPG. This adds `spec.monitoringConfig.managedPrometheusConfig.enabled` on `GCPManagedControlPlane`, mapped through to the GKE cluster on create and diffed/updated on existing clusters without disturbing other GKE-managed monitoring settings (e.g. `ComponentConfig`).

- The API shape mirrors GKE's own nested `MonitoringConfig` → `ManagedPrometheusConfig` → `Enabled` structure rather than a flattened bool, to leave room for other monitoring knobs (e.g. `ComponentConfig`) later without a breaking change.
- The update-diff logic only replaces `ManagedPrometheusConfig` on `ClusterUpdate.DesiredMonitoringConfig`, explicitly preserving whatever `ComponentConfig`/`AdvancedDatapathObservabilityConfig` GKE already has, since those are GKE-managed defaults this feature isn't meant to touch.
- Added an E2E assertion (`VerifyManagedPrometheusEnabled` in `test/e2e/gke.go`) that queries the live GKE cluster via the `container` API after creation to confirm the setting actually took effect on GCP, not just that the CR round-tripped — this is a new pattern for this test suite (existing GKE E2E specs only assert on CAPI/Kubernetes object status, never call the GCP API directly), reusing the existing `ci-gke` flavor/cluster rather than provisioning a new one.

**TODOs**:
- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
NONE
```


salasberryfin (commented):
/ok-to-test

PR #1732, opened 2026-07-29
Make load balancer health checks configurable via API

### LINKED ISSUE #1513: loadBalancerBackendPort - health check
/kind bug

Hello, we have case where we want to use different loadBalancerBackendPort, but if I define different port, e.g.:
```
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: GCPCluster
metadata:
  name: capi-gcp
  namespace: capi-gcp
spec:
  loadBalancer:
    loadBalancerType: InternalExternal
  network:
    name: private-vpc-k8s-capi
    autoCreateSubnetworks: false
    loadBalancerBackendPort: 443
```

CAPG creates health check with default port 6443 and it's not possible to change it in cluster definition (I could change it through console).

It should be great if I define loadBalancerBackendPort it will be used for all port definitions which I need.

Try to change default API port.

Will be used right loadBalancerBackendPort from manifest.

**Environment:**

- Cluster-api version: v1.9.4
- Kubernetes version: (use `kubectl version`): 1.32
- OS (e.g. from `/etc/os-release`):  Ubuntu 24.04.2 LTS

patrickdillon:
**What type of PR is this?**
/kind feature
/kind bug

The load balancer health checks may be too relaxed--allowing nodes to be offline for up to a minute before being removed from the pool. This allows users to customize the health checks to ensure offline nodes are removed more quickly.

Fixes #1513

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Added health check API for load balancers.
```


api/v1beta1/types.go
```diff
@@ -401,6 +401,42 @@ type LoadBalancerSpec struct {
 	// InternalLoadBalancer is the configuration for an Internal Passthrough Network Load Balancer.
 	// +optional
 	InternalLoadBalancer *LoadBalancer `json:"internalLoadBalancer,omitempty"`
+
+	// HealthCheck defines health check configuration for the load balancer.
+	// This configuration applies to health checks for both the external and
+	// internal load balancers. The health check protocol (HTTPS), port (6443),
```
tthvo, line 407: The API server port is configurable so it's not always `6443`, right? The health check port should follow `loadBalancerBackendPort` field or default to `6443`.

https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/ce90cddde868776ac04f722904ca2ef941f3f5b1/api/v1beta1/types.go#L334-L336

api/v1beta1/types.go
```diff
@@ -401,6 +401,42 @@ type LoadBalancerSpec struct {
...
+	// and request path (/readyz) are not configurable as they are specific to
+	// the Kubernetes API server.
+	// +optional
+	HealthCheck *LoadBalancerHealthCheck `json:"healthCheck,omitempty"`
+}
+
+// LoadBalancerHealthCheck defines the configuration for the health check
+// used by the API server load balancer(s).
+// +kubebuilder:validation:XValidation:rule="!has(self.timeoutSec) || !has(self.checkIntervalSec) || self.timeoutSec <= self.checkIntervalSec",message="timeoutSec must be less than or equal to checkIntervalSec"
```
tthvo, line 416:
```suggestion
// +kubebuilder:validation:XValidation:rule="self.timeoutSec <= self.checkIntervalSec",message="timeoutSec must be less than or equal to checkIntervalSec"
```

I think all the health check fields are _required_ so we can skip the `has` validation?

Issue #1168, opened 2024-03-07
Move managed cluster support (GKE) to general availability


salasberryfin:
/kind feature

**Describe the solution you'd like**

As of today, support for provisioning GKE clusters is still an [experimental](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/tree/main/exp) feature and has been for a number of months.

In recent community meetings we discussed the viability of moving GKE support out of the experimental folder and into general availability. This would allow CAPG to follow other providers, such as CAPA and CAPZ which have full support for their corresponding managed clusters: [CAPA](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awsmanagedcluster_controller.go), [CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/controllers/azuremanagedcluster_controller.go).

The promotion of GKE to stable could be a major step towards a more feature rich and stable CAPG, and we could push to add support for existing GKE functionalities that are still not covered by the provider (there are a few related PRs).

Since this is an important change, we thought it would be great to create this issue where we can discuss pros/cons and plan what needs to be done if we decide to go ahead.

Initially, this seems to be a non-breaking change for the API, since it means adding types but not modifying existing ones, but it would be great to have feedback from those that are more knowledgeable of the codebase. 

**Anything else you would like to add:**

As a follow up to this, `ClusterClass` support for managed clusters would be one of the relevant features we could be adding, as non-managed GCP clusters can already be provisioned from a class (refer to [this](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/402)) and this is expected to become a powerful/distinctive functionality of CAPI.

Please, feel free to share your thoughts on this.

### Related issues

- [x] https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/1387
- [x] https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/1309

damdo:
@salasberryfin is looking for some feedback on this one from core maintainers
/assign @richardcase @cpanato

salasberryfin:
Looking at unblocking this and moving the feature out of experimental.

Let's collect a list of PRs related to managed clusters that we should address:
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1024
- [ ] https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/901
- [ ] 

[...]

PR #1733, opened 2026-07-30
Adds the capability to enable/disable the Gateway API from `GCPManagedControlPlane`

jonathanrainer:
**What type of PR is this?**
/kind feature

At the minute there's no way to enable the Gateway API from the CRD, this adds the ability to do that which means we can stop having to work around this CronJobs looking at the cluster objects.

Partially addresses some of the concerns in #1168, I know that @abasu0713 was interested in this too

- There's nothing really that new here it just plumbs through the value from the CRD into the API calls, we have had to do a bit of surgery on NetworkConfig so values get saved rather than clobbered but that's it really
- There's also a special case bit to reject the field on Autopilot clusters because they handle this themselves/it's a nonsense value to set

**TODOs**:
- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Adds the capability to enable/disable the Gateway API from the GCPManagedControlPlane CustomResource
```


salasberryfin (commented):
/ok-to-test

salasberryfin (commented):
Thanks @jonathanrainer.

/lgtm
/approve

jonathanrainer:
@salasberryfin Seems like I managed to force push just after you added `lgtm` many apologies

salasberryfin (commented):
No worries @jonathanrainer 

/lgtm

PR #1735, opened 2026-08-01
Allow GKE Clusters To Be Registered To A GKE Fleet

jonathanrainer:
**What type of PR is this?**
/kind feature

At the minute there's no way to register a cluster from the controller itself, you're required to use extra infrastructure (such as doing it manually via the `gcloud` CLI or invoking a CronJob etc.) Since the goal of Cluster API is to be able to administer lots of clusters easily it seems like a very natural fit that you might want to register those clusters to a GKE Fleet. This PR makes that possible, so on a cluster by cluster basis you can opt-in to registering to either a fleet in the same project as the cluster or in another project if that's the required configuration.

It's achieved by adding a new field to the ManagedControlPlane CR which can be further configured to add a project for the fleet if required and changes to that field can be made and reconciled after the cluster's creation.

Relates to #1168

- Most of this is very standard stuff but it adds a new GCP API client and thus a new controller
- Because of the above I've added a new E2E test to prove that wiring up the new API client etc. works end-to-end
- I've also added a new section into the documentation

**TODOs**:
- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
Adds the field `fleet` to `GCPManagedControlPlane` allowing you to register the cluster to a GKE Fleet from the CustomResource
```


salasberryfin (commented):
/ok-to-test

jonathanrainer:
This is currently blocked on this PR https://github.com/kubernetes/k8s.io/pull/9983, because we need to enable the API in the E2E projects for this to be testable

PR #1754, opened 2026-08-30
fix(gke): make ClusterNetwork IP allocation fields immutable

jonathanrainer:
**What type of PR is this?**
/kind bug

`GCPManagedControlPlane.Spec.ClusterNetwork.UseIPAliases`, `.Pod`, and `.Service` (the pod/service secondary IP ranges) can only be set at GKE cluster creation time — GKE has no `UpdateCluster` field for changing a cluster's IP allocation mode or ranges afterward (confirmed against `containerpb.ClusterUpdate`), and `reconcile.go`'s `createCluster` is the only place these are read; the update path never touches them.

Previously, changing these fields on an existing `GCPManagedControlPlane` was accepted by the API server and then silently had no effect on the live cluster — the same failure class as #1576 (mutable network fields on the generic `GCPCluster` type). This adds the missing immutability checks to `ValidateUpdate`, following the same pattern already used for `GatewayAPIChannel`.

None filed — found while investigating a related, still-open bug (#1576) on the generic `GCPCluster` type. Happy to file one if preferred; didn't think it necessary for a fix this small.

No new types or reconciler changes — these fields were already create-only in practice, so the fix is purely rejecting an update that was already a no-op, not changing any runtime behavior for anyone using the fields correctly.

**TODOs**:
- [x] squashed commits
- [x] adds unit tests

**Release note**:
```release-note
Fixes a bug whereby changes to `ClusterNetwork.UseIPAliases`, `ClusterNetwork.Pod` and `ClusterNetwork.Service` would silently no-op rather than being rejected as there is no GKE API to update these after the creation of the cluster.
```

### DIFF
--- a/exp/webhooks/gcpmanagedcontrolplane_webhook.go
+++ b/exp/webhooks/gcpmanagedcontrolplane_webhook.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/ptr"
 
 	"github.com/pkg/errors"
 	expinfrav1 "sigs.k8s.io/cluster-api-provider-gcp/exp/api/v1beta1"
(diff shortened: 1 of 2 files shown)


damdo:
@jonathanrainer In principle I agree that we could make these immutable.
But given at the moment they silently get ignored it means that if we turn this on it'll prevent existing users that have such an object in the cluster to update it at all unless they drop those fields. And this might confuse people.
IMO webhooks are part of the API contract so we'd need to merge this only when we create a new CAPG v1beta2 version.

Let's add this PR to https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/1287
And set it as draft for now :) As soon as we decide to go for v1beta2 we'll be able to merge this in.

Thanks!

damdo:
/hold

API breaking changes

jonathanrainer:
> @jonathanrainer In principle I agree that we could make these immutable. But given at the moment they silently get ignored it means that if we turn this on it'll prevent existing users that have such an object in the cluster to update it at all unless they drop those fields. And this might confuse people. IMO webhooks are part of the API contract so we'd need to merge this only when we create a new CAPG v1beta2 version.
> 
> Let's add this PR to #1287 And set it as draft for now :) As soon as we decide to go for v1beta2 we'll be able to merge this in.
> 
> Thanks!

Yer that's completely fair, my thought was if we're going to move GKE support to GA (and by extension move it out of `exp` and into a proper versioned API) this is a change to make at that point, it's really just tidying up more than anything else

damdo:
SGTM, let's draft it until then, TY.

PR #1755, opened 2026-08-30
fix(gke): Various small fixes around DiskSizeGb, Dead Code & Docs

jonathanrainer:
**What type of PR is this?**

/kind bug
/kind cleanup
/kind documentation

**What this PR does / why we need it**: While I was doing some work in another area for another PR I discovered that `GCPManagedMachinePoolClassSpec` has two fields that both configure a GKE node pool's disk size: `DiskSizeGb` (`int32`) and `DiskSizeGB` (`int64`, added earlier). `ConvertToSdkNodePool` let `DiskSizeGB` silently win when both were set, via a lossy `int64`->`int32` cast, and the update webhooks only enforced immutability on `DiskSizeGb` - `DiskSizeGB` meanwhile could be freely changed on a running node pool with no rejection, bypassing the intended immutability.

This PR:
- Makes `DiskSizeGb` canonical (it matches GKE's own `int32` field with no lossy cast) and ensures it takes precedence when both fields are set. `DiskSizeGB` is marked deprecated but kept for backward compatibility - it's a released CRD field. If we wanted to remove it completely we could, but that would be a breaking change and I don't know how best to handle that here.
- Adds the missing `ValidateImmutable` check for `diskSizeGB` in both the `GCPManagedMachinePool` and `GCPManagedMachinePoolTemplate` webhooks, closing the bypass.
- Removes two harmless-but-confusing dead code blocks in `ConvertToSdkNodePool` that re-checked `Spec.ImageType`/`Spec.DiskType` against themselves a second time.
- Fixes the `GCPMachineTemplate` example manifests in `docs/book/src/topics/preemptible-vms.md`, which used fields that don't exist on this provider's API (`osDisk`, `managedDisk.storageAccountType`, `vmSize`, top-level `region`).

The three commits are intentionally separated: dead code removal, doc fixes, and the actual duplication/immutability fix (with its tests and CRD regen).

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
fix(gke): close a webhook gap that allowed `diskSizeGB` to bypass disk-size immutability on GCPManagedMachinePool/GCPManagedMachinePoolTemplate; DiskSizeGb is now the canonical field and DiskSizeGB is deprecated.
```

### DIFF
--- a/exp/api/v1beta1/types_class.go
+++ b/exp/api/v1beta1/types_class.go
@@ -98,6 +98,7 @@ type GCPManagedMachinePoolClassSpec struct {
 	MachineType *string `json:"machineType,omitempty"`
 	// DiskSizeGb is the size of the disk attached to each node, specified in GB.
 	// The smallest allowed disk size is 10GB. If unspecified, the default disk size is 100GB.
+	// If DiskSizeGB is also set, DiskSizeGb takes precedence.
 	// +optional
 	DiskSizeGb *int32 `json:"diskSizeGb,omitempty"`
 	// LocalSsdCount is the number of local SSD disks to be attached to the node.
@@ -121,6 +122,12 @@ type GCPManagedMachinePoolClassSpec struct {
 	DiskType *DiskType `json:"diskType,omitempty"`
 	// DiskSizeGB is size of the disk attached to each node,
 	// specified in GB.
+	//
+	// Deprecated: DiskSizeGB duplicates DiskSizeGb (both configure the same
+	// underlying GKE node pool disk size) and is retained only for backward
+	// compatibility with existing resources that already set it. This field
+	// remains immutable and will not gain new functionality; use DiskSizeGb
+	// for new manifests. If both fields are set, DiskSizeGb takes precedence.
 	// +kubebuilder:validation:Minimum:=10
 	// +optional
 	DiskSizeGB *int64 `json:"diskSizeGB,omitempty"`
--- a/exp/webhooks/gcpmanagedmachinepool_webhook.go
+++ b/exp/webhooks/gcpmanagedmachinepool_webhook.go
@@ -108,6 +108,11 @@ func (*GCPManagedMachinePool) ValidateCreate(_ context.Context, r *expinfrav1.GC
 	gcpmanagedmachinepoollog.Info("Validating GCPManagedMachinePool create", "name", r.Name)
 
 	var allErrs field.ErrorList
+	var allWarns admission.Warnings
+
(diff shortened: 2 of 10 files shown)


exp/api/v1beta1/types_class.go
```diff
@@ -121,6 +122,12 @@ type GCPManagedMachinePoolClassSpec struct {
 	DiskType *DiskType `json:"diskType,omitempty"`
 	// DiskSizeGB is size of the disk attached to each node,
 	// specified in GB.
+	//
+	// Deprecated: DiskSizeGB duplicates DiskSizeGb (both configure the same
+	// underlying GKE node pool disk size) and is retained only for backward
+	// compatibility with existing resources that already set it. This field
+	// remains immutable and will not gain new functionality; use DiskSizeGb
+	// for new manifests. If both fields are set, DiskSizeGb takes precedence.
```
barbacbd, line 130: Do you think we could/should add a deprecated warning tag here? WDYT @damdo @salasberryfin

exp/webhooks/gcpmanagedmachinepool_webhook.go
```diff
@@ -202,6 +202,13 @@ func (*GCPManagedMachinePool) ValidateUpdate(_ context.Context, old, r *expinfra
 		allErrs = append(allErrs, err)
 	}
 
+	if err := webhookutils.ValidateImmutable(
+		field.NewPath("spec", "diskSizeGB"),
```
barbacbd, line 211: nit: You are cleaning up this one already, should we add test cases for the other immutable fields? this is just a thought.

barbacbd (commented):
/lgtm

barbacbd:
@jonathanrainer It looks like golint is failing. Can you take a look?

jonathanrainer:
> @jonathanrainer It looks like golint is failing. Can you take a look?

No worries! I'll get it fixed up once I'm back at my desk

exp/api/v1beta1/types_class.go
jonathanrainer: I've added an admission warning in the same style as `ControlPlaneVersion` so it follows the patterns we already have

exp/webhooks/gcpmanagedmachinepool_webhook.go
jonathanrainer: Yer you make a fair point, it wasn't a whole lot of extra work to add the other test cases so I filled them out to give us more coverage

damdo (commented):
/approve

Also re-adding

/lgtm

as it was added by @barbacbd but then dropped due to force-push

I also updated the release notes to make sure we warn users of the field ordering change.

PR #1756, opened 2026-08-31
GCP scale from zero

daniel-rejniak:
**What type of PR is this?**
  
  /kind feature

  Implements scale-from-zero autoscaling support for GCP by adding `Status.Capacity` and `Status.NodeInfo` to `GCPMachineTemplate`. This enables cluster-autoscaler
  to bootstrap NodePools from 0 replicas by providing instance type resource information (CPU, memory, architecture, OS) without requiring existing nodes.

  **Changes:**
  - Add `GCPMachineTemplateStatus` struct with `Capacity` (CPU, memory) and `NodeInfo` (architecture, OS) fields
  - Add `+kubebuilder:subresource:status` marker and regenerate CRD
  - Implement `GCPMachineTemplateController` to reconcile templates and populate status
  - Query GCP Compute API for machine type specs (capacity + architecture)
  - Query GCP Images API to detect OS from image metadata (Linux vs Windows)
  - Use region-level fallback (`region-a`) when failure domains not yet discovered (mirrors CAPA approach)
  - Add required RBAC permissions for status subresource
  - Add e2e test verifying Status.Capacity and Status.NodeInfo populated after cluster creation

  Fixes #
  
  - Follows CAPI scale-from-zero proposal: https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210310-opt-in-autoscaling-from-zero.md
  - Matches CAPA/CAPZ pattern: both implement Capacity + NodeInfo
  - Fallback zone logic is safe: GCP machine type specs are identical across all zones in a region
  - Controller only reconciles templates with an owner Cluster to avoid orphaned API calls
  - Unit tests validate capacity conversion from GCP API format to Kubernetes resource quantities
  - Unit tests validate architecture detection (ARM64 vs X86_64) and OS detection (Linux vs Windows)
  - E2E test integrated into existing cluster creation test to avoid extra cluster overhead

  **Release note**:
```release-note
feat: GCPMachineTemplate now populates Status to enable cluster-autoscaler scale-from-zero functionality
```


damdo:
cc. @elmiko @theobarberbany

controllers/gcpmachinetemplate_controller.go
```diff
@@ -0,0 +1,221 @@
...
+
+func machineTypeCapacity(machineType *compute.MachineType) (corev1.ResourceList, error) {
+	if machineType.GuestCpus <= 0 || machineType.MemoryMb <= 0 {
+		return nil, fmt.Errorf("cpu=%d memoryMb=%d", machineType.GuestCpus, machineType.MemoryMb)
+	}
+
+	return corev1.ResourceList{
+		corev1.ResourceCPU:    *resource.NewQuantity(machineType.GuestCpus, resource.DecimalSI),
+		corev1.ResourceMemory: resource.MustParse(fmt.Sprintf("%dMi", machineType.MemoryMb)),
+	}, nil
```
elmiko, line 234: should we be detecting/adding GPUs in here as well?

elmiko (commented):
looking good to me, i just have a question about gpus.

daniel-rejniak:
Apologies for adding changes mid-review. Added Status.NodeInfo field alongside the existing Status.Capacity implementation.

  Why: CAPI scale-from-zero proposal recommends both fields. CAPA and CAPZ both implement NodeInfo + Capacity. This completes the pattern.

  What changed:
  - Added NodeInfo struct (architecture: amd64/arm64, operatingSystem: linux/windows)
  - Controller detects arch from MachineType API, OS from Images API
  - Updated reconcile condition to check both capacity AND nodeInfo
  - Added unit tests

  Commit: 5040f63d

controllers/gcpmachinetemplate_controller.go
daniel-rejniak: CAPA (AWS) and CAPZ (Azure) both populate only CPU/Memory in Status.Capacity — neither includes GPUs. I don't think it's needed for current use cases. I can add in follow-up PR if GPU nodepool support is required unless you think otherwise ?

damdo:
Personally I'd try and see if we can add an e2e test for this if possible.
Something not too complex, that tests the basic functionality is working.
@elmiko do you know of any prior art for this in CAPI land?

controllers/gcpmachinetemplate_controller.go
elmiko: i was just curious. ultimately we probably want this info, but if you are just wanting to reach parity with aws and azure, then i'm fine to leave it out for now.

elmiko:
@damdo there are some core autoscaler tests, https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/autoscaler.go , and i do see some scale from zero testing in there. perhaps a start?

daniel-rejniak:
Il look into this, will try to add some basic e2e tests as per request to ensure the core functionality is tested.

damdo:
Thanks @elmiko 

@daniel-rejniak cool thanks, keep us posted!

daniel-rejniak:
@damdo I extended the already existing test under `test/e2e/e2e_test.go` (mainly for simplicity and low overhead since we have resources there that we can assert against) this proves that my implementation is populating `Status.Capacity` and `Status.NodeInfo` in GCPMachineTemplate. 

This approach matches other CAPI providers (CAPA, CAPZ) - they test Status fields are populated, not full autoscaler integration. I believe the full autoscale behaviour is tested in the external repo provided by @elmiko 

Please let me know if this is sufficient ?

elmiko:
i think this is making sense to me.

/lgtm

controllers/gcpmachinetemplate_controller.go
```diff
@@ -0,0 +1,370 @@
...
+		logger.Info("GCPMachineTemplate does not yet have an owner Cluster")
+		return ctrl.Result{}, nil
+	}
+
+	if annotations.IsPaused(cluster, template) {
+		logger.Info("GCPMachineTemplate or owner Cluster is paused")
+		return ctrl.Result{}, nil
+	}
+
+	if cluster.Spec.InfrastructureRef.Kind != "GCPCluster" || cluster.Spec.InfrastructureRef.Name == "" {
```
damdo, line 141: The reconciler returns without requeueing when the owner Cluster's `spec.infrastructureRef` is not populated, but this controller does not watch Cluster resources. When Cluster infrastructure is initialized later, no event enqueues the GCPMachineTemplate, so `status.capacity` and `status.nodeInfo` can remain unset. Could we add a Cluster watch and mapper for owned GCPMachineTemplates, or otherwise requeue until the infrastructure reference is available? This is especially relevant during ClusterClass initialization.

controllers/gcpmachinetemplate_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	return machineType, nil
+}
+
+func machineTypeCapacity(machineType *compute.MachineType) (corev1.ResourceList, error) {
+	if machineType.GuestCpus <= 0 || machineType.MemoryMb <= 0 {
+		return nil, fmt.Errorf("cpu=%d memoryMb=%d", machineType.GuestCpus, machineType.MemoryMb)
+	}
+
+	return corev1.ResourceList{
```
damdo, line 231: `status.capacity` currently includes only CPU and memory. Since GCPMachineTemplate supports `GuestAccelerators`, GPU-backed machine groups should also report their GPU capacity. Could we add the appropriate GPU resource to the capacity calculation and add a test for a template with `GuestAccelerators`?

controllers/gcpmachinetemplate_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	nodeInfo, err := machineTypeNodeInfo(ctx, clusterScope.Compute, clusterScope.Project(), machineType, template)
+	if err != nil {
+		logger.Info("Failed to determine nodeInfo from image, using defaults", "error", err)
+		// Use defaults: set only architecture from machine type, omit OS
+		nodeInfo = &infrav1.NodeInfo{
+			Architecture: getArchitectureFromMachineType(machineType),
+		}
+	}
+
+	original := template.DeepCopy()
```
damdo, line 207: When an image is configured but the image lookup fails, the reconciler stores a partial `NodeInfo` containing only the architecture. On the next reconcile, the existing capacity and NodeInfo make the template look complete, so the image lookup is not retried. Could we avoid persisting this fallback for image lookup errors and requeue or otherwise mark the status incomplete so transient failures can recover?

test/e2e/e2e_test.go
```diff
@@ -104,6 +106,25 @@ var _ = Describe("Workload cluster creation", func() {
...
+
+			Expect(template.Status.Capacity).NotTo(BeNil(), "Status.Capacity should be populated")
+			Expect(template.Status.Capacity.Cpu().IsZero()).To(BeFalse(), "CPU capacity should be set")
+			Expect(template.Status.Capacity.Memory().IsZero()).To(BeFalse(), "Memory capacity should be set")
+
+			Expect(template.Status.NodeInfo).NotTo(BeNil(), "Status.NodeInfo should be populated")
+			Expect(template.Status.NodeInfo.Architecture).To(BeElementOf(infrav1.ArchitectureAmd64, infrav1.ArchitectureArm64), "Architecture should be amd64 or arm64")
+			Expect(template.Status.NodeInfo.OperatingSystem).To(Equal(infrav1.OperatingSystemLinux), "OperatingSystem should be linux")
+
 			By("Scaling worker node to 3")
```
damdo, line 128: The current e2e test checks that capacity and node information are populated, then scales the MachineDeployment from one replica to three. Could we also test the actual Cluster Autoscaler scale-from-zero flow? For example, create the MachineDeployment with zero replicas, schedule a workload that requires that node group, and verify that the autoscaler increases the replica count and the workload becomes schedulable.

controllers/gcpmachinetemplate_controller.go
daniel-rejniak: @damdo I addressed the PR comments the only one left is this one related to GPU capacity. The original requirement I had was to add `capacity` and `nodeInfo` for autoscaler.

Questions from my side : 

1. PR is getting bigger and harder to review, can the GPU capacity be added later as a feature request ? It would be easier to build on top of what I already have in this PR or is it absolutely required for the autoscaler to function properly ? 
2. No prior art : I cross referenced implementations from CAPZ and CAPA and none of them have this implemented "yet".  The proposal : https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210310-opt-in-autoscaling-from-zero.md is very generic.

do you have any specific examples / more detailed requirements around this ?

daniel-rejniak: @damdo I fixed all the PR comments you had the only one remaining is this one related to GPU's.

Can we defer this to next PR / feature request ? The reasons from my side : 

- Current PR is already fairy bulky and complex adding the GPU capacity will increase the scope even more
- Implementing the GPU capacity requires more thought my main requirements covered the memory and CPU. Neither CAPZ nor CAPA cover those so there is not much prior art it requires proper design especially that from my research Im unable to grab the GPU types through API but rather need to parse them out. 
- Current PR already gives solid foundation for the GPU capacity implementation

what do you think about this ?

damdo: I guess this something we can do without in this PR and add with a follow-up. I'm fine with that but I think it'd be good to see this happening across all 3 providers you mentioned.

Once this PRs merges would you mind opening an issue to track this? Thanks!

daniel-rejniak: sure Il open up a feature request to track this

test/e2e/data/cluster-autoscaler/deployment.yaml.tmpl
```diff
@@ -0,0 +1,71 @@
...
+      app: cluster-autoscaler
+  template:
+    metadata:
+      labels:
+        app: cluster-autoscaler
+    spec:
+      serviceAccountName: cluster-autoscaler
+      containers:
+        - name: cluster-autoscaler
+          image: registry.k8s.io/autoscaling/cluster-autoscaler:v1.33.3
```
damdo, line 22: It makes sense to make the Cluster Autoscaler version configurable because the image is pinned to `v1.33.3`, while the E2E config currently uses Kubernetes `v1.35.5`. Let's set the E2E configs to a released Cluster Autoscaler version with a major.minor version compatible with `KUBERNETES_VERSION`.

Add the variable constant in `test/e2e/suite_test.go`:

```go
const ClusterAutoscalerVersion = "CLUSTER_AUTOSCALER_VERSION"
```

Add the version to the E2E config files, for example `test/e2e/config/gcp-ci.yaml`:

```yaml
variables:
  # Keep this major.minor version in sync with KUBERNETES_VERSION when bumping Kubernetes minor releases.
  # For example, if KUBERNETES_VERSION=v1.35.5, then CLUSTER_AUTOSCALER_VERSION=v1.35.<latest available>
  # For the latest available patch version of Cluster Autoscaler, see:
  # https://github.com/kubernetes/autoscaler/releases?q=cluster-autoscaler&expanded=true
  CLUSTER_AUTOSCALER_VERSION: "v1.35.2"
```

In `test/e2e/autoscaling_from_zero_test.go`, pass `e2eConfig.MustGetVariable(ClusterAutoscalerVersion)` to the deployment template. Then, in this template:

```yaml
image: registry.k8s.io/autoscaling/cluster-autoscaler:{{ .ClusterAutoscalerVersion }}
```

The registry and image name can remain hard-coded.

damdo:
@elmiko does the e2e testing look as you would expect? TY

elmiko:
e2e looks good to me, thanks for adding that scale from zero test.

damdo:
@daniel-rejniak  Looks reasonable, I'll give it a final look (I've also pinged the other maintainers to see if they are all ok with that), meanwhile, could you squash the commits into fewer ones to clean up the fixups & rebases/merges? Thanks!

daniel-rejniak:
Thanks for looking into this @damdo , I squashed the commits into two logical ones.

PR #1758, opened 2026-09-02
fix(gke): don't force a Gateway API channel when unset

jonathanrainer:
**What type of PR is this?**
/kind bug

`checkDiffAndPrepareUpdate` always compares GKE's actual Gateway API channel against `Spec.ClusterNetwork.GatewayAPIChannel`, even when that field is left unset on the `GCPManagedControlPlane`. An unset field computes a desired channel of `CHANNEL_UNSPECIFIED`, but GKE never actually reports that back - it always has some real channel (e.g. Autopilot clusters mandate `STANDARD` and reject any other value). The result is a permanent mismatch: every reconcile issues an update GKE immediately rejects, in a tight retry loop that never resolves.

This change only compares/applies the Gateway API channel when the user has explicitly configured `GatewayAPIChannel`, matching how GKE itself treats it: GKE decides its own default when the field is omitted, so an unset spec should never fight whatever GKE actually has.

This is one of three related PRs from the same diagnosis. It's independent of the other two and can be reviewed/merged on its own. For proof this composes with the other fixes to get a fully green GKE e2e run, see this Prow run: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-gcp/1750/pull-cluster-api-provider-gcp-e2e-test/2095209473210585088

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
fix(gke): GCPManagedControlPlane was getting stuck in a permanent update loop when Gateway API channel is left unset (e.g. on GKE Autopilot clusters)
```

### DIFF
--- a/cloud/services/container/clusters/reconcile.go
+++ b/cloud/services/container/clusters/reconcile.go
@@ -562,20 +562,22 @@ func (s *Service) checkDiffAndPrepareUpdate(existingCluster *containerpb.Cluster
 		log.V(4).Info("Master authorized networks config update check", "desired", desiredMasterAuthorizedNetworksConfig)
 	}
 
-	// Gateway API channel
-	var desiredGatewayAPIChannel *infrav1exp.GatewayAPIChannel
-	if cn := s.scope.GCPManagedControlPlane.Spec.ClusterNetwork; cn != nil {
-		desiredGatewayAPIChannel = cn.GatewayAPIChannel
-	}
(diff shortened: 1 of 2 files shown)


damdo (commented):
Makes sense to me

/approve
/lgtm

PR #1761, opened 2026-09-05
fix(gke): avoid nil pointer panic when network name is unset

### LINKED ISSUE #1187: Nil pointer dereference, on an invalid spec

jonathanrainer:
/kind bug

`GCPManagedCluster.spec.network.name` is `+optional`, but `createCluster` dereferenced it unconditionally when building the GKE `CreateClusterRequest`, so creating a `GCPManagedCluster` without `network.name` set crashed the controller with a nil pointer dereference.

The GKE Container API already documents that an unset `network` field means "connect the cluster to the project's `default` network" (https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster), so the fix simply stops forcing the dereference and lets the configured value (or nothing) pass through, deferring to GKE's own default-network handling rather than CAPG inventing a stricter requirement of its own.

Fixes #1187

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
fix(gke): avoid nil pointer panic when GCPManagedCluster.spec.network.name is unset
```

### DIFF
--- a/cloud/services/container/clusters/reconcile.go
+++ b/cloud/services/container/clusters/reconcile.go
@@ -29,6 +29,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/pkg/errors"
+	"k8s.io/utils/ptr"
 	infrav1exp "sigs.k8s.io/cluster-api-provider-gcp/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-gcp/util/reconciler"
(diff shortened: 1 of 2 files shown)


damdo (commented):
/approve
/lgtm

PR #1026, opened 2023-09-22
Support more configurations for GCPManagedMachinePool

gzcharleszhang:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Currently, GCPManagedMachinePool only supports a limited set of configurations. This PR adds support for a wider set of configurations made available by the [GCP SDK](https://pkg.go.dev/cloud.google.com/go/container/apiv1/containerpb#NodePool). Specifically, it adds support for the following fields:
```
- Config
  + MachineType
  + DiskSizeGb
  + ServiceAccount
  + ImageType
  + Labels
  + LocalSsdCount
  + Tags
  + DiskType
  + Taints
  + ResourceLabels
- Autoscaling
  + Enabled
  + LocationPolicy
  + TotalMinNodeCount
  + TotalMaxNodeCount
- Management
  + AutoUpgrade
  + AutoRepair
- MaxPodsConstraint
  + MaxPodsPerNode
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1025

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
GCPManagedMachinePool now supports more configurations made available by https://pkg.go.dev/cloud.google.com/go/container/apiv1/containerpb#NodePool
- Config
  + MachineType
  + DiskSizeGb
  + ServiceAccount
  + ImageType
  + Labels
  + LocalSsdCount
  + Tags
  + DiskType
  + Taints
  + ResourceLabels
- Autoscaling
  + Enabled
  + LocationPolicy
  + TotalMinNodeCount
  + TotalMaxNodeCount
- Management
  + AutoUpgrade
  + AutoRepair
- MaxPodsConstraint
  + MaxPodsPerNode
```

PR #944, opened 2023-06-06
Add additional fields to GCPManagedMachinePool

akshay196-rafay:
**What type of PR is this?**

/kind api-change

**What this PR does / why we need it**:

Support additional fields to GCPManagedMachinePool kind.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Part of #906

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squashed commits
- [X] includes documentation

**Release note**:

```release-note
Support additional fields to GCPManagedMachinePool type.
```

PR #1763, opened 2026-09-11
fix(gke): deprecate duplicate `machineType` in favor of `instanceType`

jonathanrainer:
**What type of PR is this?**

/kind bug
/kind deprecation

**This was edited from a previous version that I wrote while I was clearly half asleep, apologies, the version below is now correct**

`GCPManagedMachinePoolClassSpec` has two fields that both configure a GKE node pool's machine type: `MachineType` and `InstanceType`. `ConvertToSdkNodePool` let `InstanceType` silently win whenever both were set, purely because it happened to be applied later in the function - there was no documented or intentional precedence.

This PR:
- Makes `InstanceType` canonical - this matches CAPA and also is in-keeping with the original version of what was added
- Emits a create-time admission warning when `MachineType` is set, on both `GCPManagedMachinePool` and `GCPManagedMachinePoolTemplate`, steering users towards `InstanceType`.

#1755 independently touches nearby lines in `ConvertToSdkNodePool` (removing a duplicate `ImageType`/`DiskType` check) as part of an unrelated `DiskSizeGb`/`DiskSizeGB` dedup fix. This PR deliberately leaves those lines alone to avoid two PRs racing on the same hunk - whichever merges first, the other should apply cleanly on rebase.

Both `MachineType` and `InstanceType` were already immutable on update, so no new immutability check was needed here - this is purely about resolving which one wins on create when both are set.

**TODOs**:

- [x] squashed commits
- [x] adds unit tests

**Release note**:
```release-note
fix(gke): deprecate machineType field in favour of instanceType
```


damdo:
@jonathanrainer Is there a reason why there were 2 such similar fields? Is this limited to GCPManagedMachinePoolClassSpec or is this present in other types? Is this also present in other CAPI providers?

TY

jonathanrainer:
> @jonathanrainer Is there a reason why there were 2 such similar fields? Is this limited to GCPManagedMachinePoolClassSpec or is this present in other types? Is this also present in other CAPI providers?
> 
> TY

@damdo So I did some digging around in the history of the repo and it seems the following happened:

- `InstanceType` gets added in https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/944, which is June 2023
- `MachineType` gets added in https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1026, which was in December 2023
- 1026 is the only PR to ever mention MachineType and everything else including `GCPMachine` and `GCPMachinePool` uses `InstanceType`, so it feels like a duplicate that got missed in a bigger PR rather than anything else

In terms of it being present in other types, something similar is happening with `DiskSizeGb` and `DiskSizeGB` (which are both different types entertainingly enough) but that's the only other example I can find.

In terms of other providers I dug through CAPA and CAPZ and the former has `InstanceType` and CAPZ has a `SKU` field which serves the same purpose. 

So aligning on `InstanceType` means we can both align with CAPA and also use the same naming as the underlying GKE API

damdo:
> > @jonathanrainer Is there a reason why there were 2 such similar fields? Is this limited to GCPManagedMachinePoolClassSpec or is this present in other types? Is this also present in other CAPI providers?
> > TY
> 
> @damdo So I did some digging around in the history of the repo and it seems the following happened:
> 
> * `InstanceType` gets added in [Add additional fields to GCPManagedMachinePool #944](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/944), which is June 2023
> * `MachineType` gets added in [Support more configurations for GCPManagedMachinePool #1026](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1026), which was in December 2023
> * 1026 is the only PR to ever mention MachineType and everything else including `GCPMachine` and `GCPMachinePool` uses `InstanceType`, so it feels like a duplicate that got missed in a bigger PR rather than anything else
> 
> In terms of it being present in other types, something similar is happening with `DiskSizeGb` and `DiskSizeGB` (which are both different types entertainingly enough) but that's the only other example I can find.
> 
> In terms of other providers I dug through CAPA and CAPZ and the former has `InstanceType` and CAPZ has a `SKU` field which serves the same purpose.
> 
> So aligning on `InstanceType` means we can both align with CAPA and also use the same naming as the underlying GKE API

Thanks @jonathanrainer !
Are you proposing we align on instanceType or machineType? 😄 

I'd say we go with instanceType as that's the one we also have for the main GCPMachine type https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/12f7a81a2044f63ea67ba400c4208aeed272fd5e/api/v1beta1/gcpmachine_types.go#L250-L251

and it is also what CAPA has.

Let's do that (code and PR title/desc need changing) :)

jonathanrainer:
> > > @jonathanrainer Is there a reason why there were 2 such similar fields? Is this limited to GCPManagedMachinePoolClassSpec or is this present in other types? Is this also present in other CAPI providers?
> > > TY
> > 
> > 
> > @damdo So I did some digging around in the history of the repo and it seems the following happened:
> > 
> > * `InstanceType` gets added in [Add additional fields to GCPManagedMachinePool #944](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/944), which is June 2023
> > * `MachineType` gets added in [Support more configurations for GCPManagedMachinePool #1026](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/1026), which was in December 2023
> > * 1026 is the only PR to ever mention MachineType and everything else including `GCPMachine` and `GCPMachinePool` uses `InstanceType`, so it feels like a duplicate that got missed in a bigger PR rather than anything else
> > 
> > In terms of it being present in other types, something similar is happening with `DiskSizeGb` and `DiskSizeGB` (which are both different types entertainingly enough) but that's the only other example I can find.
> > In terms of other providers I dug through CAPA and CAPZ and the former has `InstanceType` and CAPZ has a `SKU` field which serves the same purpose.
> > So aligning on `InstanceType` means we can both align with CAPA and also use the same naming as the underlying GKE API
> 
> Thanks @jonathanrainer ! Are you proposing we align on instanceType or machineType? 😄
> 
> I'd say we go with instanceType as that's the one we also have for the main GCPMachine type
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/12f7a81a2044f63ea67ba400c4208aeed272fd5e/api/v1beta1/gcpmachine_types.go#L250-L251
> 
> and it is also what CAPA has.
> 
> Let's do that (code and PR title/desc need changing) :)

I clearly must have been half asleep when I wrote this *facepalm*

jonathanrainer:
@damdo Have re-written it now so it fits what I actually wanted, not what my half asleep brain thought was sensible. Thanks for your patience!

damdo (commented):
/approve

/assign @salasberryfin @theobarberbany @barbacbd

damdo:
@jonathanrainer needs rebasing when you have a moment

